PR #25, opened 2017-03-22
Try a conditional in after_success

bensteinberg:


emmalemma:
Looks like that did the trick!

PR #58, opened 2017-05-02
WIP Casebooks remodel and UI

emmalemma:


app/controllers/application_controller.rb
```diff
@@ -20,55 +18,10 @@ def layout_switch
     return false if request.xhr?
   end
 
-  def load_single_resource
```
cgruppioni: Nice

app/controllers/casebooks/books_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def book_params
+    params.require(:casebook_book).permit(:title, :subtitle, :ordinals, :headnote)
+  end
+
+  def find_book
+    @book = Casebook::Book.find params[:id]
+  end
+
+  def canonical_redirect
```
cgruppioni: When would this be used?

app/models/casebook/concerns/contents.rb
```diff
@@ -0,0 +1,53 @@
+module Casebook::Concerns::Contents
+  extend ActiveSupport::Concern
+
+  def children
+    contents.where(['array_length(casebooks.ordinals, 1) = ?', ordinals.length + 1])
```
cgruppioni: I don't fully understand what an ordinal is..

app/models/casebook/contents.rb
```diff
@@ -0,0 +1,102 @@
...
+#  copy_of_id    :integer
+#  is_alias      :boolean
+#  material_type :string
+#  material_id   :integer
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+
+class Casebook::Contents < Casebook::Generic
+  default_scope {where.not(root_id: nil)}
```
cgruppioni: Would be a better question in the book Casebook::Book model... but what is a root_id?

app/models/casebook/concerns/contents.rb
```diff
@@ -0,0 +1,53 @@
...
+      else
+        element.update_column :ordinals, prefix + [ord]
+        ord += 1
+      end
+    end
+  end
+
+  # Transform a flat sorted array of casebooks into an array tree
+  # [1, 1.1, 1.1.1, 1.2, 2, 2.1, 2.1.1] => [1, [1.1, [1.1.1], 1.2], 2, [2.1, [2.1.1]]]
+  def tree
```
cgruppioni: Why is this here if it's also in the generic model?

app/models/casebook/resource.rb
```diff
@@ -0,0 +1,32 @@
...
+#  material_type :string
+#  material_id   :integer
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+
+class Casebook::Resource < Casebook::Contents
+  default_scope {where.not(material_id: nil)}
+
+  belongs_to :material, polymorphic: true, inverse_of: :casebooks, required: true
```
cgruppioni: Can you explain the relationship between contents & resources and what the polymorphic material path is for?

app/controllers/casebooks/sections_controller.rb
```diff
@@ -0,0 +1,121 @@
+require 'net/http'
+require 'uri'
+
+class Casebooks::SectionsController < ApplicationController
+  layout 'casebooks'
+  before_action :find_book
+  before_action :find_section, except: [:new, :create, :index]
+  before_action :find_parent, only: [:new, :create]
+  before_action :canonical_redirect, only: [:show]
+  before_action :disable_turbolinks_cache, only: [:new]
```
cgruppioni: why disable_turbolinks_cache?

app/controllers/casebooks/sections_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+  def find_parent
+    @parent = if params[:parent]
+      @section = @book.contents.find_by_ordinals parse_ordinals(params[:parent])
+    else
+      @book
+    end
+  end
+
+  def parse_ordinals ordinals
+    ordinals = ordinals.split(/\.|,/) if ordinals.is_a? String
```
cgruppioni: What does the ordinal object look like here?

app/javascript/ui/casebooks/dragging.coffee
```diff
@@ -0,0 +1,97 @@
+ui = require 'ui/turbolinks'
```
cgruppioni: You're doing coffeescript?

app/controllers/casebooks/books_controller.rb
emmalemma: The idea here is if we have a descriptive id in the url, like `/casebooks/100-third-restatement-torts` for a casebook titled "The Third Restatement of Torts", we want the slug portion to be optional— so `/casebooks/100` works, or `/casebooks/100-torts-draft`, since the title may change and we don't want to break links. But we want to canonicalize the current slug, so that new URLs use it instead of the one the user followed there.

So `canonical_redirect` checks if the user followed the *currently* canonical URL to get here, and redirects them if not. This is obviously only useful to do for GET routes, nothing that a user wouldn't be linking to.

app/models/casebook/concerns/contents.rb
emmalemma: Yeah this is confusing! Sorry I didn't write anything up about it, It's one of the things I was meaning to do when I got sick D:

"Ordinal" is just a fancy word for ordering numbers like "first", "second", "third", etc. Another name for this column could be "path"; it just an array of integers which stores the full position of this section in the main table of contents. 

So the third resource in the second part of the fourth chapter has a path of [4, 2, 3].

This column is indexed, so all sections of a casebook can be looked up by the combination of their Book's (root) ID and their ordinals. The upshot of this is you can easily locate a section from a descriptive URL like `/casebooks/100-third-restatement-torts/sections/4.3.2-negligence`

It does add some weight to tasks like reordering of sections, so it's something I'd love to get your intuition on, but I feel like it's a simpler model overall, since each piece of content in a casebook can be looked at in isolation.

app/models/casebook/contents.rb
emmalemma: `root_id` should probably  be `book_id`. Every piece of content that isn't a book stores the id of the book it belongs to. So `root_id` + `ordinals` uniquely identifies every piece of content in the database.

Each piece of content could also store its immediate parent, but that's not necessary to represent the tree.

app/models/casebook/concerns/contents.rb
emmalemma: Oh, I'd meant to move it from there to here. This is supposed to only be relevant to Books and Chapters (anything that has contents).

It might be more simple to just keep that stuff in Generic, though.

app/models/casebook/resource.rb
emmalemma: Sure. Casebook::Contents is an abstract class for everything that can be content— so anything that's not a Book, ie Sections and Resources.

A resource is a _reference_ to a piece of real material, like a Case or a Text. So it has to be polymorphic to represent the different things it could point to.

Think of the Casebook::Resource as the table of contents entry, where the `material` is the actual content of that section. (This column could be renamed `resource`.)

app/controllers/casebooks/sections_controller.rb
emmalemma: I think this isn't used presently because clicking 'new section' hits `create` and goes straight to the edit state for the section. The issue here was that if you go to the new section route, and that view renders the full table of contents, you'll see a flash of content from the previous hit to new section. I.e. you'll see the last section you added briefly become editable before it actually adds a new section and makes that one editable.

Solved by disabling cache, but this would also be fixed by adding a little more advanced JS so that new routes happen mostly client-side. I might prefer to do that.

app/controllers/casebooks/sections_controller.rb
emmalemma: This is just a helper method to break up a string like '1.2.3', which is the format of the section id, into [1,2,3] which is the value in the database. It should probably live somewhere else?

app/models/casebook/resource.rb
emmalemma: The benefit of breaking things up like this (rather than pointing directly to a Case) is that you can clone a casebook without needing to copy all of its resources, just copying the references. Then the actual resource can be copied when the user actually edits it (copy-on-write). This will make cloning faster for large casebooks,  as well as less data-heavy for those where few changes are made to the clone.

app/javascript/ui/casebooks/dragging.coffee
emmalemma: I defaulted to coffeescript since it's the default in new Rails projects. (At least I think that's still true?) 

I like CS for some of its utility features that replace a big library like Underscore, but I'm not wedded to it if you would be more comfortable working in pure JS or ES6 with Babel.

app/models/casebook/concerns/contents.rb
emmalemma: Just to be more explicit: The major benefit of doing it this way is that we can pull out all the contents for a given book, in sorted order by table-of-contents number, in a single SQL query: `where(root_id: book.id).order(:ordinals)`. Pulling out children and subtrees is also easy by looking for ordinals having a given prefix. We don't need any complex SQL or multiple database hits to follow a parent-child tree. That list can then be more or less rendered straight to HTML (albeit needing the `tree` transformation for the `.chapter-contents` tags. That could be eliminated with slightly more complex HTML generation, though.) 

It adds complexity to inserting rows, since siblings need to be pushed over and all children in a subtree need to be updated. But those steps can also be done in single SQL queries (see the code in Casebook::Contents), and in trade we remove a bunch of complexity from the `SELECT` side, which I think is a good deal.

app/models/casebook/resource.rb
emmalemma: (This is what the `is_alias` boolean column is theoretically for. If `is_alias` is true, this Resource record points to a Case (for example) that is shared by another casebook, and it should be copied before making edits.)

app/controllers/casebooks/books_controller.rb
cgruppioni: Oh wow!

app/models/casebook/contents.rb
```diff
@@ -0,0 +1,102 @@
...
+#  updated_at    :datetime         not null
+#
+
+class Casebook::Contents < Casebook::Generic
+  default_scope {where.not(book_id: nil)}
+
+  validates_length_of :ordinals, minimum: 1
+  validates_uniqueness_of :ordinals, scope: :book_id, if: :book_id?, message: 'are not unique'
+
+  before_validation :move_siblings, if: :ordinals_changed?
```
cgruppioni: I don't see `orderinals_changed` defined anywhere --- is that ActiveRecord?

app/models/casebook/concerns/contents.rb
cgruppioni: 👍

app/controllers/casebooks/books_controller.rb
cgruppioni: 👍

app/models/casebook/contents.rb
cgruppioni: It feels weird that the root_id of a book will be null --- are you checking if root_id = null a lot?

app/models/casebook/resource.rb
cgruppioni: I don't understand how resource is a reference to a case but the schema for it is casebooks? 

In my mind
`material_type` = `'case'`
`material_id` = `case_id`

app/controllers/casebooks/sections_controller.rb
cgruppioni: I think it makes sense to live there

app/javascript/ui/casebooks/dragging.coffee
cgruppioni: I would like to do ES6 --

Is babel necessary or can we just use sprockets-es6 for the ES6 syntax? 
webpacker is in rails 5 but that's probably more than we need?

emmalemma: Cool, let's go with that. 

sprockets-es6 is just babel for sprockets, yeah? Seems like it doesn't provide module support, though... Are you thinking we'd just vendor dependencies and use sprockets requires? 

Alternative is we could keep using `yarn` with something simpler like `sprockets-commoner` to replace webpack.

app/models/casebook/contents.rb
emmalemma: Yep, AR. `attribute_changed?` is true if `attribute` has a different value from the persisted value (up to when the `save` callbacks fire).

app/models/casebook/resource.rb
emmalemma: Yeah, there's a naming issue. "Casebook" is not totally right for this concept.

Think of the abstract class `Casebook::Generic` as any node in a tree representation of a Table of Contents. It could be the root (a Casebook), a branch (a Section), or a leaf (a Resource).

All three share a conceptual model in that they have a position in a Table of Contents, they have a `title`, `headnote`, `collaborators`, they can be public or not, they can be cloned, and so on.

They differ in that only Casebooks don't have a root (or have themselves as root), and only Resources have a reference to a piece of real content (a Case or TextBlock or Link).

So yes: a Resource is a reference to a case which shares behavior with an abstract class of objects that includes Sections and Casebooks. In the data model, that makes it a row in `casebooks`; but that's not a good name, because a Resource is conceptually _not_ the same thing as a Casebook.

We could call the abstract module/table something more generic like ContentNode?

emmalemma: Yeah, that feels much nicer. I'll probably go ahead and do that.

app/models/casebook/contents.rb
emmalemma: The condition for a Book could be that `book_id = id`; it would make the same sense conceptually. There are a few reasons `null` is a more comfortable value for the root:

- Even though the primary key of a record shouldn't really change, so it's nbd, it's still not technically normalized to have the primary key duplicated in `book_id`.
- We want to be able to pull out a list of all Books; `book_id = null` is easier than `book_id = id`.
- We often want to be able to get a list of all contents of a book, not including the book; as null, we don't need the condition that `id != book_id`.

But on the flip side, finding the parent of a content node requires extra logic to check if the parent is a book. I don't feel super strongly about this, I guess I can see what it looks like that way.

app/javascript/ui/casebooks/dragging.coffee
cgruppioni: I had https://github.com/rails/webpacker w/ sprockets-es6  recommended to me. 

but if you want to do yarn that's fine with me

app/models/casebook/resource.rb
cgruppioni: Nice! I agree

PR #60, opened 2017-05-03
Cap Api Case Import integration 

cgruppioni:
User can search for a case by name or citation.
A list of results from cap api are rendered and the user can select one to download. 
After a successful or unsuccessful download the user is taken back to the search results page (with a flash saying 'success' or 'fail'), giving them the opportunity to search and download another case. 
The future view implementation will ideally be in-line with the normal h20 search results. 

CaseCitation is an important part in user's searching for cases and it's important to note that it's stored in a separate table than Case. This is because cases can be cited in multiple volumes (ie: state, regional, federal) and therefore have multiple citations. though this isn't that common in our data so far. 

There were 3 service objects created: 
* CapApiSearchResults
* CaseDownloader
* DuplicateCaseChecker ( called inside CaseDownloader )

A new Notifier Mailer was created to notify the h20 admin of case import failures. The admin is sent the user that attempted the download, the case metadata and the cap api response if the error was an API failure.


app/services/duplicate_case_checker.rb
```diff
@@ -0,0 +1,22 @@
...
+  def self.perform(cases)
+    new(cases).perform
+  end
+
+  def initialize(cases)
+    @cases = cases
+  end
+
+  def perform
+    cases.reject { |kase| case_already_exists?(kase) }
```
emmalemma: This is just a personal thing, but I dislike the `c -> k` convention... I'd prefer adding an extra label, either something descriptive like `test_case` or something redundant like `case_record`.

app/views/cap_api_imports/_cases.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <th>Short Name></th>
+        <th>Full Name</th>
+        <th>Citation</th>
+        <th>Descision Date (Original)</th>
+        <th>Court Name</th>
+      </tr>
+    </thead>
+    
+    <tbody>
+      <% @cases.each do |kase| %>
```
emmalemma: `case_result`?

test/services/cap_api_search_results_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'application_system_test_case'
+
+class CapApiSearchResultsTest < ApplicationSystemTestCase
```
emmalemma: Seeing as this is a service test, does it need all the system test machinery? Could we make a `ServiceTestCase < TestCase` class that just pulls in the CapApi/ActionMailer stubs and whatever else these actually depend on to run?

config/application.rb
```diff
@@ -34,5 +34,7 @@ class Application < Rails::Application
 
     config.action_mailer.default_url_options = { :host => 'localhost' }
     config.user_verification_recipients = ['cgruppioni@law.harvard.edu']
+    config.admin_email = 'cgruppioni@law.harvard.edu'
+    config.cap_api_key = '2c62c54b47e507b2eee20a70f29f1b4ae0ccd1a3'
```
emmalemma: Can this be in `development.rb`/`test.rb`, so it doesn't cover up a potential bug on prod/staging?

app/services/duplicate_case_checker.rb
cgruppioni: awesome -- i totally agree

PR #71, opened 2017-05-27
Translate .html.haml templates into .html.erb 

cgruppioni:
All files in commit `migrate .haml files into .erb - round 1` were done with https://haml2erb.org/

The other commits could not be done automatically. They were done by hand and then verified that they looked the same as the .haml verion.


cgruppioni:
Merged in in an emma commit

PR #78, opened 2017-06-28
allow dev users to self-verify, give passwords to seed users

emmalemma:


emmalemma:
This adds three seed users: admin@example.edu, professor@example.edu, and student@example.edu. All passwords are "password". Also adds a dev-flagged verify button to the edit user page.

PR #94, opened 2017-08-07
Finish search filtering 

cgruppioni:
* Fix filter selected text not showing up on initial page load
* Add filters for cases & users, filters stay selected while navigating between tabs
* Filters only show options from search results ( add solr configs ) 
* Render 'no result' heading when 0 search results
* Render citations in case search results
* Fix sort by date

### DIFF
--- a/app/models/content/casebook.rb
+++ b/app/models/content/casebook.rb
@@ -61,9 +61,7 @@ class Content::Casebook < Content::Node
       owners.map &:id
     end
 
-    string :owner_affiliation, stored: true, multiple: true do
-      owners.map &:affiliation
-    end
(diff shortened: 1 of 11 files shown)


app/models/content/casebook.rb
```diff
@@ -61,9 +61,7 @@ class Content::Casebook < Content::Node
       owners.map &:id
     end
 
-    string :owner_affiliation, stored: true, multiple: true do
-      owners.map &:affiliation
-    end
+    string(:affiliation, stored: true) { owners.first.affiliation }
```
emmalemma: `owners.first.try(:affiliation)`

emmalemma:
Looks ok other than that bug 👍

PR #109, opened 2017-08-22
Update installation instructions

jcushman:
I was curious to try out the new H2O code! Got it running with a couple of tweaks to the installation instructions -- thanks to @emmalemma for help debugging the solr issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ solr/data/
 solr/pids/
 tmp/
 guard-sunspot/
+.idea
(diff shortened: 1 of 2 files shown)


bensteinberg:
I use `rake` instead of `rails` to start and stop solr, but it looks like it works either way.

jcushman:
It sounds like in Rails 5 it's a good practice to just use `rails` for
everything for simplicity?

http://blog.bigbinary.com/2016/01/14/rails-5-supports-rake-commands-using-rails.html
http://www.lessanvaezi.com/use-rails-instead-of-rake-in-rails-5/


On Aug 22, 2017 5:07 PM, "Ben Steinberg" <notifications@github.com> wrote:

I use rake instead of rails to start and stop solr, but it looks like it
works either way.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/harvard-lil/h2o/pull/109#issuecomment-324152035>, or mute
the thread

cgruppioni:
Thanks @jcushman! & thanks @bensteinberg for saying that .. i also have still been using `rake`

PR #110, opened 2017-08-23
word export table of contents, bugfixes for annotation

emmalemma:
- word exports includes table of contents (click on the arrow, then Update Table > Update page numbers only. this can be shown automatically when the document is first opened.)
- fixes a couple of breaking cross-browser annotation bugs
- fixes an issue with table of contents nesting


emmalemma:
screen shot 2017-08-23 at 5 00 41 am

PR #135, opened 2017-08-31
Re-enable tests and remove unused legacy code

emmalemma:
-  fixes skipped tests, including annotation and export tests
- deletes lots of un-tested, un-used code


cgruppioni:
Merged in with: https://github.com/harvard-lil/h2o/pull/202

PR #137, opened 2017-09-02
Migration bugfix prior to re-import of 100 casebooks dump

emmalemma:
- fixes a bug causing migrated annotations to appear in the wrong paragraph


emmalemma:
This re-import has:

- no debug IDs
- associated Users, where present in the dump
- fixed annotations

PR #194, opened 2017-11-15
New migrate script instructions

cgruppioni:
* ancestry is not carried over from playlists to casebooks (since it is linking to playlists that were not imported over). 
* instead we are saving the root playlist's user id into the new content_node field root_user_id
    * all users were already being imported over with the migration so this works without other scripts needed to grab/manipulate data
* if a user clones a casebook that was originally created in h2o v1, the ancestry data will only be filled in with v2 data. This doesn't really matter though because when the root_user_id field is filled in, that is the information that will always be used. 

Instructions for how to migrate now are in `lib/playlist_id_migration_directions.md` including some notes about how the data was set. One playlist item in particular breaks the script and needs to be deleted before migration.


lib/playlist_id_migration_directions.md
```diff
@@ -0,0 +1,9 @@
+## Translate playlists from [this file](https://docs.google.com/spreadsheets/d/1gHHsqbiZNxeYeWApM7GhU4L8wiZuIJP3EvCd_hXk8xk/edit?usp=sharing) these ids were pulled to create the playlist id array. Some of the numbers in the google sheet aren't the most recent version of that playlist, so the spreadsheet and array don't exactly match. This file was last updated 11/14/17 - any playlists added to this file after that date are not included in this array.
```
edwardloveall: This is a little silly, but you might replace this with a title instead of a whole description in a header tag:

```
## Translate playlists notes

These are instructions on how to translate playlists into casebooks from (google spreadsheet). Note that these ids were pulled...

1. First step...
2. Second step...
3. etc.
```

lib/playlist_id_migration_directions.md
```diff
@@ -0,0 +1,9 @@
+## Translate playlists from [this file](https://docs.google.com/spreadsheets/d/1gHHsqbiZNxeYeWApM7GhU4L8wiZuIJP3EvCd_hXk8xk/edit?usp=sharing) these ids were pulled to create the playlist id array. Some of the numbers in the google sheet aren't the most recent version of that playlist, so the spreadsheet and array don't exactly match. This file was last updated 11/14/17 - any playlists added to this file after that date are not included in this array. 
+
+1. `rake db:schema:load` clear the database and rebuild the sceme
+2. `pg_restore --verbose --clean --no-acl --no-owner -U h2oadmin -h h2o-alpha-94.coc4xoq8nz1r.us-east-1.rds.amazonaws.com -d h2o ~/h2o_prod-data.dump` port the data dump into the database
```
edwardloveall: Also, you may want to have them make sure they have an `h2o_prod_data.dump`. I'd also probably use the current directory, not the home directory.

lib/playlist_id_migration_directions.md
```diff
@@ -0,0 +1,9 @@
...
+2a. (for running locally: `pg_restore --verbose --clean --no-acl --no-owner -h localhost -d h2o_dev ~/h2o_prod-data.dump `)
```
edwardloveall: Same here

bensteinberg: Maybe don't include the address of the RDS instance...

PR #198, opened 2017-11-16
Deleting section reflow ordinals

cgruppioni:
casebook.contents has data from before the resource/section is destroyed - so passing in the child itself allows you to check if it's been deleted.

### DIFF
--- a/app/models/content/concerns/has_children.rb
+++ b/app/models/content/concerns/has_children.rb
@@ -20,7 +20,8 @@ module Content::Concerns::HasChildren
     true
   end
 
-  def reflow_contents
+  def reflow_contents(child)
+    @child = child
     reflow_tree(tree, ordinals)
   end
 
@@ -31,7 +32,7 @@ module Content::Concerns::HasChildren
         element_prefix = ord == 1 ? prefix : prefix + [ord-1]
         reflow_tree(element, element_prefix)
       else
-        unless element.destroyed?
+        unless element == @child && @child.destroyed?
           element.update_column :ordinals, prefix + [ord]
           ord += 1
         end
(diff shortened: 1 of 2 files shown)


app/models/content/concerns/has_children.rb
```diff
@@ -20,17 +20,19 @@ def present?
     true
   end
 
-  def reflow_contents
-    reflow_tree tree, ordinals
+  def reflow_contents(child)
+    @child = child
```
cgruppioni: since reflow_tree is a recursive method (see line#33) it felt weird to be re-passing in child into reflow_tree

app/models/content/concerns/has_children.rb
```diff
@@ -20,17 +20,19 @@ def present?
...
+    reflow_tree(tree, ordinals)
   end
 
-  def reflow_tree tree, prefix = []
+  def reflow_tree(tree, prefix = [])
     ord = 1
     tree.each do |element|
       if element.is_a? Enumerable
-        reflow_tree element, ord == 1 ? prefix : prefix + [ord-1]
+        element_prefix = ord == 1 ? prefix : prefix + [ord-1]
```
edwardloveall: As long as you’re in here, let’s get rid of the ternary and replace it with an `if`🙂. It took me a while to parse it.

edwardloveall: Also, are you sure this will do what you want? Maybe I’m missing something, but if you always set ord to 1 at the start of the method, this conditional will always return true.

cgruppioni: No it's looping through the tree after setting ord to 1...  so it's not always 1

PR #202, opened 2017-11-20
Merge in 3 month old PR 

cgruppioni:
* https://github.com/harvard-lil/h2o/pull/135/files

* added `config/initializers/RailsPatch22584.rb file` which doesn't seem like it should be necessary in Rails 5 but it does fix the problem ` ActiveRecord::Fixture::FixtureError: table "case_citations" has no column named "case".` when running tests 
* `test/services/casebooks_test` runs & passes


test/system/casebooks_test.rb
```diff
@@ -26,53 +26,48 @@ class CasebookSystemTest < ApplicationSystemTestCase
...
 
-      fill_in 'Title', with: 'Test casebook title'
-      fill_in 'Subtitle', with: 'Test casebook subtitle'
-      fill_in 'Headnote', with: 'Test casebook headnote'
+      fill_in 'content_casebook_title', with: 'Test casebook title'
+      fill_in 'content_casebook_subtitle', with: 'Test casebook subtitle'
+      fill_in 'content_casebook_headnote', with: 'Test casebook headnote'
 
-      click_button 'Save'
+      click_link 'Save'
```
edwardloveall: Pro tip for the future: You can use `click_on` to click a link or a button

cgruppioni: Nice

PR #207, opened 2017-12-01
Correct formated_decision_date

bensteinberg:
This fixes a typo, and adds a conditional. I'm not sure if this is the right move, or if it would be better to ensure that all cases have decision dates.

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -150,7 +150,7 @@ class Case < ApplicationRecord
     "cases/court_case"
   end
 
-  def formated_decision_date
+  def formatted_decision_date
     self.decision_date.strftime("%B %d, %Y")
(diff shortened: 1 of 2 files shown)


app/models/case.rb
```diff
@@ -150,8 +150,10 @@ def to_partial_path
     "cases/court_case"
   end
 
-  def formated_decision_date
-    self.decision_date.strftime("%B %d, %Y")
+  def formatted_decision_date
+    if !self.decision_date.blank?
```
cgruppioni: this is great! You can also do `self.decision_date.present?`

bensteinberg: That seems nicer, thanks! I'll change it.

PR #208, opened 2017-12-05
Static pages content and styling

abziegler:
This PR adds content and styling for static pages. 

User guide content is very bare-bones for now. Code in views/layouts/main.html.erb for changing page <title> works but is funky.


cgruppioni:
This looks great! Some of the spacing indentation is weird which can be tricky so let's look at your sublime settings after the h2o meeting

PR #210, opened 2017-12-10
Fix search filters to include all authors & schools 

cgruppioni:
The pagination of search results occurs in the sunspot search - so it only returns the top 15 results and was only returning authors/schools for that data. Contrary to the way the filters were originally built, sunspot has a build in feature called facets that does not add to any latency time. The search results then were modified so that the user can start typing in the name of the filter they want with selectize.js. Style is consistent across all filters. 

- Add selectize2 gem to add selectize class to dropdowns (autocomplete)
- For authors sunspot search - search for attribution (name) instead of id /// this might be a problem with old data who use title instead of attribution, but i looked at the production data and it didn’t look like it would be an issue - users who don’t have an attribution also didn’t have a title and therefore users#display_name would be characterized as anonymous. Something to keep an eye on. 
- Renamed filters from @authors and @schools to @attributions and @affiliations to match data models
- User Sunspot facet feature to build filter data for attributions and affiliations
- Removed main.css and main.js files which were overriding application.css & application.js 
- Removed labels for filters
- Edited no results found message to no include type or search term which was overkill


app/controllers/searches_controller.rb
```diff
@@ -8,82 +8,59 @@ def show
...
-        end
-      end
-    else
-      authors.push User.find(params[:author]) if params[:author].present?
-      schools.push params[:school] if params[:school].present?
-    end 
-
-    { authors: authors.compact, schools: schools.compact }
-  end
+  def extract_filter_data_from_sunspot_object(rows)
```
edwardloveall: Nice method name!

PR #236, opened 2018-01-18
Professor Verification

cgruppioni:
Deploy: 
* Need to reindex

After Deploy
* Need to manually verify known professors so that they do show up in search results 

PR Description: 
* Rename user#verified to user#verified_email
* Add columns user#verified_professor and user#professor_verification_requested
* User requests professor verification on edit profile page. An email is sent to the email that is set in the ENV variable professor_verification_email ( Brett in production, Casey in all other environments). The field then is disabled and said verification requested
* The email has a link to the admin edit user page and you check the top box that says "Verified professor" and save form. 
* Then the professor's edit profile page would show Verified Professor
* Green checkmark only shows up for verified professors
* Green checkmarks show up on casebook search results page ( user page has a field to show their profile picture so I did not add the checkmark there) 

 **See images below**

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -33,6 +33,15 @@ section.sign-up, section.sign-in, section.reset-password {
   .help {
     @include make-md-column(5.5);
   }
+  .verified {
+    display: inline-block;
+    width: 12px;
+    height: 12px;
+    line-height: 5px;
+
+    background-image: asset-url('ui/verified.png');
+    color: transparent;
+  }
 }
 
 section.profile {
@@ -46,5 +55,27 @@ section.reset-password {
     h3 {
       margin-left: 15px;
     }
+  }
+}
+.verified-professor {
+  font-weight: bold;
+  margin: 20px 0;
+}
+
+.user_professor_verification_requested, .user_professor_verification_requested {
+  .help-block {
+    display: none;
+  }
+
+  label {
+    font-weight: bold;
+  }
+
+  .necessity {
+    display: none;
+  }
+
+  .disabled {
+    color: grey;
   }
 }
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -66,7 +66,7 @@ class UsersController < ApplicationController
     end
     user_params.delete :current_password
 
-    if !@user.professor_verification_sent && !@user.professor_verification_requested
+    if user_params[:professor_verification_requested] == "1" && !@user.professor_verification_requested
       @user.send_professor_verification_request_to_admin
     end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -14,7 +14,7 @@ class UsersController < ApplicationController
 
   def verify
     if current_user.present? && current_user == User.where(id: params[:id]).first && params[:token] == current_user.perishable_token
-      current_user.update_column(:verified, true)
+      current_user.update_column(:verified_email, true)
       flash[:notice] = 'Thank you. Your account has been verified. You may now contribute to H2O.'
       redirect_to user_path(current_user)
     elsif current_user.present?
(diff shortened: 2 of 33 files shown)


cgruppioni:
screen shot 2018-01-18 at 5 03 19 pm
screen shot 2018-01-18 at 5 02 47 pm
screen shot 2018-01-18 at 5 02 58 pm

screen shot 2018-01-18 at 5 05 25 pm ( I was verifying myself and my title is "Professor Casey Gruppioni" ) 
I added the affiliation & email address because I thought it'd help whoever know faster whether to verify this person or not. 

screen shot 2018-01-18 at 5 07 15 pm

screen shot 2018-01-18 at 5 07 20 pm

app/assets/stylesheets/users.scss
```diff
@@ -33,6 +33,15 @@ section.sign-up, section.sign-in, section.reset-password {
   .help {
     @include make-md-column(5.5);
   }
+  .verified {
+    display: inline-block;
+    width: 12px;
```
edwardloveall: I liked your `@include square(13px)` above.

app/controllers/users_controller.rb
```diff
@@ -58,14 +58,18 @@ def edit
...
-
     user_params = permitted_user_params
+
     if user_params[:password].present? && !@user.valid_password?(user_params[:current_password])
       flash[:error] = "Current password is incorrect."
       return render :edit
     end
     user_params.delete :current_password
 
+    if user_params[:professor_verification_requested] == "1" && !@user.professor_verification_requested
```
edwardloveall: Since this updates the professor user's `professor_verification_requested` attribute, does this mean a professor can't request this twice? Would that cause problems like if the email got lost or the link somehow became invalid?

Would it be better to deny them verification only if they're already verified, as opposed to requesting to be verified?

```ruby
if user_params[:professor_verification_requested] == "1" && !@user.verified_professor?
```

app/views/layouts/_header.html.erb
```diff
@@ -16,11 +16,11 @@
...
       <% if current_user.present? %>
         <div class="current-user">
-          <% if current_user.verified? %>
+          <% if current_user.verified_professor? %>
             <div class="verified">
-              <%= t 'header.user.verified' %>
+              <%= t 'header.user.verified_professor' %>
             </div>
-          <% else %>
+          <% elsif ! current_user.verified_email? %>
```
edwardloveall: Extra space. It will work with that space but it through me off for a second.

config/environments/production.rb
```diff
@@ -142,6 +142,7 @@
   # Admin email to receive the 'new user needs verification' emails
   config.user_verification_recipients = (ENV["USER_VERIFICATION_RECIPIENTS"] || '').split(' ')
 
-  config.admin_email = ENV["ADMIN_EMAIL"]
-  config.cap_api_key = ENV["CAP_API_KEY"]
+  # config.admin_email = ENV["ADMIN_EMAIL"]
+  # config.cap_api_key = ENV["CAP_API_KEY"]
+  config.professor_verification_email = "bjohnson@law.harvard.edu"
```
edwardloveall: I'd suggest setting these as environment variables. That way if someone leaves or you just want to change it, it isn't a whole redeploy process. Deploying can change overtime and doesn't always remain simple. It also removes changes like this from the commit history since it's more user data than app data.

db/migrate/20180111182959_rename_users_verified_to_email_verified.rb
```diff
@@ -0,0 +1,5 @@
+class RenameUsersVerifiedToEmailVerified < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :users, :verified, :verified_email
```
edwardloveall: Woah `:verified_email` is an interesting field type!

cgruppioni: What do you mean?

app/controllers/users_controller.rb
cgruppioni: For now this feels like the best option to start with. The verification option still valid would be confusing to professors/TAs. Certainly email failing is a concern, but I think we'll go with this and then see if it doesn't work.

db/migrate/20180111182959_rename_users_verified_to_email_verified.rb
edwardloveall: Oh pff, ignore. I read that as `add_column`, not `rename_column`.

PR #284, opened 2018-04-27
Skip links via layouts

rebeccacremona:
This puts a skip link and a default target for that link in the layout files where it looks like skipping might come in handy.

If a particular view's "main" content is actually further down the page than the location of the default target, the view can provide a custom target via two steps:
1. Override the default by adding the line `<% content_for :custom_skip_target do %>i'm-arbitrary-text<% end %>` to the view.
2.  Place `<p id="main" tabindex="-1" class="sr-only">Main Content</p>` in the desired location.

This is kind of a zillion little commits: happy to rebase and combine them if you want!

### DIFF
--- a/app/views/layouts/casebooks.html.erb
+++ b/app/views/layouts/casebooks.html.erb
@@ -42,7 +42,7 @@
       <div class="casebook-inner">
         <div class="top-strip"></div>
         <% if @casebook.present? %>
-          <% content_for :custom_skip_target do %>i'm-arbitrary-text<% end %>
+          <% content_for :custom_skip_target do %>arbitrary text, necessary so that content_for :custom_skip_target evaluates to true<% end %>
           <p id="main" tabindex="-1" class="sr-only">Main Content</p>
         <% end %>
         <%= yield %>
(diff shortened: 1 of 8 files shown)


app/views/layouts/casebooks.html.erb
```diff
@@ -41,6 +41,10 @@
     <div class="content">
       <div class="casebook-inner">
         <div class="top-strip"></div>
+        <% if @casebook.present? %>
+          <% content_for :custom_skip_target do %>i'm-arbitrary-text<% end %>
```
cgruppioni: Why are you using content_for?

rebeccacremona: This is a trick for letting a "child" view overrule something in a parent view/layout. In the layout, a default skip target is provided unless a child/grandchild/etc. has provided content for custom_skip_target (https://github.com/harvard-lil/h2o/pull/284/files#diff-e3a4a6249a8f8ae3c560f0ba99f7134bR26). So, to customize the location of the skip link, you provide some arbitrary content in the child, and then put a skip target in your customized spot. It'd be nice if it could be empty, rather than having some arbitrary text, but it looks like that doesn't work in Rails.

Issue #322, opened 2018-05-04
Can't delete items (sections/resources)


bjohnson747:
1) Created new sections in a new casebook. When i clicked the (X) on a section, it didn't remove it.

Tried to delete: 

<img width="853" alt="deletion attempt" src="https://user-images.githubusercontent.com/12850389/39642565-211e639c-4fa0-11e8-86e1-23ff47350130.png">

Clicked delete, section remains:

<img width="939" alt="section remains" src="https://user-images.githubusercontent.com/12850389/39642564-210d282a-4fa0-11e8-9fbe-3325bceb0cc4.png">

Examples from: https://alpha.opencasebook.org/casebooks/57032-fresh-casebook/layout
I was also unable to delete a resource I added.

abziegler:
Reproduced bug.

PR #336, opened 2018-05-17
Users can delete sections & resources from table of contents 

cgruppioni:
#322 

Removing turbolinks broke being able to delete a section & resource 

- Add rest delete call when submitting delete section form
- stop even propagation from the submit link 

``

- remove "sourceMappingURL" in annotator.store.js which was preventing JS from being legibly loaded in browser for debugging 
- update development config file so that `guard` automatically compiles assets


app/controllers/content/sections_controller.rb
```diff
@@ -84,8 +84,10 @@ def destroy
     if !@section.destroy
       flash[:error] = "Could not delete #{@section.ordinal_string} #{@section.title}"
     end
-    redirect_to layout_casebook_path @casebook
+#
```
jcushman: Delete extra #?

PR #337, opened 2018-05-17
Duplicate casebooks on create

cgruppioni:
#337 

When creating a new casebook from scratch, two casebooks were being created. 

The casebook methods were pretty confusingly named. So I renamed / simplified them.

### DIFF
--- a/app/controllers/content/casebooks_controller.rb
+++ b/app/controllers/content/casebooks_controller.rb
@@ -17,22 +17,15 @@ class Content::CasebooksController < Content::NodeController
     render 'content/show'
   end
 
-  # def edit
-  #   @casebook.update_attributes public: false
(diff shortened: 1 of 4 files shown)


app/controllers/content/casebooks_controller.rb
```diff
@@ -8,7 +8,8 @@ class Content::CasebooksController < Content::NodeController
   def new
     @casebook = Content::Casebook.create(public: false, collaborators: [Content::Collaborator.new(user: current_user, role: 'owner')])
     logger.debug @casebook.errors.inspect
-    redirect_to edit_casebook_path @casebook
```
cgruppioni: originally this redirected to edit_casebook_path which then redirected to layout_casebook_path

PR #345, opened 2018-05-21
Header Accessibility

rebeccacremona:
- Hover/focus styles for the logo link
- "Create casebook" button is a real button
- The modal that it opens has all the behavior described here: https://github.com/harvard-lil/accessibility-tools/tree/master/code/modals
- The buttons/links are all the same size
- User picture removed from display per team decision!
- Dropdown coded up as a touch- and keyboard-usable Bootstrap dropdown
- Dropdown made white and full-width per Adam.

image
(The screenshot removed my cursor, which was hovering over "Edit", which is why that's gray. keyboard focus is on the button, which is why that has focus styles. )

Lemme know if you don't like anything about the way this looks!

Not in this PR:
- The modal accessibility fixes will need to be applied to every modal. So, I should abstract that out into a more generic component. If you have no objections, I'd like to do that when fixing the next modal, rather than right now. (Having two examples will make it easier to get the abstraction right, I think.)
- If the user picture is not being displayed, then the ability to upload one can probably be removed from the code base, but that feels like a separate project.
- The generic focus styles should, ideally, be applied to everything, rather than being applied to a few elements at a time, like I'm doing here. But, when I tried that, weird visual relics popped up in unexpected places. I'll fix that as I go along, and then hopefully can switch over to a general declaration.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -25,22 +25,22 @@
               <%= t 'header.user.unverified' %>
             </div>
           <% end %>
-          <a class="wrapper" href="<%= user_path(current_user) %>">
-            <div class="name">
-              <%= current_user.display_name %>
-            </div>
-            <div class="portrait">
-              <%= image_tag current_user.image.url :thumb %>
-            </div>
-          </a>
-          <div class="user-links">
-            <a class="user-link" href="<%= root_path %>">
-              <%= t 'header.user.dashboard' %>
-            </a>
-            <a class="user-link" href="<%= edit_user_path(current_user) %>">
-              <%= t 'header.user.edit-profile' %>
-            </a>
-            <%= button_to t('header.user.sign-out'), user_session_path(current_user_session), method: 'delete', class: 'user-link' %>
+          <div class="dropdown">
+            <button type="button" id="dLabel" class="dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" >
+              <span class="name">
+                <%= current_user.display_name %>
+              </span>
+              <span class="caret"></span>
+            </button>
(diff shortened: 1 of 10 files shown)


cgruppioni:
I just created issue #354 to address `If the user picture is not being displayed, then the ability to upload one can probably be removed from the code base, but that feels like a separate project.`

app/views/layouts/_header.html.erb
```diff
@@ -7,7 +7,7 @@
     </div>
     <div class="links">
       <% if current_user.present? %>
-        <div class="create-casebook" data-action="show-casebook-modal"><%= t('header.links.create-casebook') %></div>
+        <button type="button" class="create-casebook" data-action="show-casebook-modal"><%= t('header.links.create-casebook') %></button>
```
cgruppioni: Is there a reason you used `button` instead of `button_to`?

rebeccacremona: Aha. Sorry, I'm not used to Rails' view helpers. Can you show me how to use `button_to` here?

rebeccacremona: Nevermind, I should be able to figure it out. Fix forthcoming.

rebeccacremona: Aha! Yes, there is a reason.  If this is the right `button_to` (https://apidock.com/rails/ActionView/Helpers/UrlHelper/button_to), it sounds ike `button_to` creates a whole form that, when you click on the button, submits the form. Here, we just need a button to trigger some javascript that opens the modal and performs related housekeeping: no HTTP request involved.

cgruppioni: Ah! Ok --- I was wrong.

Issue #301, opened 2018-04-30
Annotations: links


abziegler:
Group of bugs re annotation > links:
- No warning that prefix is required (h2o just redirects un-prefixed links to the preview mode of the resource)
- Should be able to edit links; currently the 'edit' button does not function
- Links should open in new tab

PR #350, opened 2018-05-23
Edit link annotations

cgruppioni:
TODO before peer review: move e.stopDefault() calls 

#301 

Previously clicking edit link for an annotation sent an empty request and didn't update the annotation. I created an `'edit-link'` case that renders the editLinkTemplate with the link filled in and the class has a listener `'update-link'` to update the annotation. 

I found the `Annotator` class confusing because it is so large so I reorganized the functions to be in more of an order of operations. 

The flow for this ticket is 
1. Click on ‘edit link’
2. `doAnnotationAction `triggered with case ‘edit-link’
3. Template `edit-link`
4. `editLinkTemplate()`
5. `submit .update-link'`
6. `updateAnnotation()`

Other:
* I refactored a few methods that were outside of the Annotator component into another file `lib/ui/content/annotations/placement` 
* "No warning that prefix is required (h2o just redirects un-prefixed links to the preview mode of the resource)"
** Fixed by adding `type="url"` 
* Links open in new tab

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,7 +71,6 @@ gem 'wkhtmltopdf-binary', require: false
 # gem 'minitest-rails-capybara'
 # gem 'quiet_assets'
 
-gem 'pry-byebug'
 
 group :development do
   gem 'annotate'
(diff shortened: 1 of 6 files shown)


jcushman:
Cool! Just skimmed over this, but I like the refactoring to an order that makes more sense to you.

PR #351, opened 2018-05-23
Search page tweaks

rebeccacremona:
This is dependent on https://github.com/harvard-lil/h2o/pull/345, so, the git diff includes those commits at the moment. It should automatically update appropriately to only include the most recent 4 commits (fd71d96 - 1994642) after 345 is merged. But I'll keep my eye out, and if it doesn't, will rebase!

This PR:
- swaps out select2 comboboxes for regular selects to avoid keyboard trap and other documented issues
- attempts some passable css for selects; will open an issue to tweak/re-address if we have ideas for improving
- added some aria-labels for the form elements and a bit of other labeling
- added focus styles for the search results tab links

This PR does not:
- apply focus styles to the results list items themselves (opening an issue to investigate why this didn't work as expected)
- try to apply any specialized patterns for "tabs" or for search results pages to try and polish the screen reader experience.


lib/locales/en/search.yml
```diff
@@ -5,9 +5,9 @@ en:
       n-results: "%{n} results found:"
       n-casebook-results: "Casebooks (%{n})"
       n-case-results: "Cases (%{n})"
-      n-user-results: "Users (%{n})"
+      n-user-results: "Authors (%{n})"
```
cgruppioni: Nice

rebeccacremona:
Okay, this didn't remove the extra commits after you merged #345. I'll rebase!

PR #360, opened 2018-05-29
CSS

rebeccacremona:
This is an accidentally giant PR. Sorry! All it touches are styles, and the tiniest bit, the markup, to affect the styles.

The easiest way to review is probably to pull down the code and run it locally, and see what you think. I'll include screenshots of some of the biggest changes.

Important to know: it turns out there were a couple subtle problems in the font-loading CSS, so the fonts specified by the CSS were not loading: your computer's backups were being used instead. This PR tweaks that CSS so the specified fonts load. It's a big change! You might find it jarring. If you do, we can change all these fonts to load Helvetica Neue instead, which is probably what you've gotten used to seeing. We don't yet own the font intended for the font page; TBD.

- each listing in a Tables of Contents is now a bit shorter, and the text is vertically centered, for easier readability (especially for people with eye-tracking issues)
image

- There are also now hover/focus styles for the Table of Contents links, blue on published, yellow on draft
image
image

- As best as I can tell, everything on the site now zooms cleanly up to 300% (and probably beyond that). Adding zoom styles is the same thing as adding mobile styles, so you should find the site looks a bit better on mobile now. Certainly mobile will need another pass, especially from someone with a better eye to aesthetics than me!! 

image

image

- I moved the action buttons in the markup, so that they appear right away in the tab order (and are thus more discoverable for people using screen readers, and easier to navigate to for people using their keyboard). Visually, at large screen sizes, they look the same as ever, but as your screen gets smaller, they appear towards the top. IMPORTANT: I might have missed pages where action buttons are supposed to appear. I'd appreciate a second set of eyes, to make sure I didn't miss any views.

image

image

image

image

I also added some assorted focus styles to the casebook/resource pages.

Finally, two small fixes to problems introduced by my last PR: 
- reintroduce some space at the bottom of the page, lost in https://github.com/harvard-lil/h2o/commit/7e302f4bb844e92fd94c97206135c8d71a6c8675
- I didn't test the dropdown menu for people with short names, only long ones: this works for everything but crazy long names

Sorry it's so giant!


rebeccacremona:
Man, no idea why that test fails..... can reproduce locally...... From the screenshot it's clear the element IS of course, still present... hmmmm....

rebeccacremona:
Okay, that test fails because it turns out some of the action buttons are little inline forms, not just buttons/links, and, of course, you can't nest forms inside other forms (like the edit casebook form). So, the html was invalid, so the html parser couldn't find the element.

I'll figure something out.

rebeccacremona:
Okay, on the two edit forms, I had to move the action buttons lower in the DOM for now. Which, again, still only applies to mobile/zoomed views. In prod, in those views, the action buttons are at the bottom of the contents anyways, so this doesn't really change anything.

Edit casebook page, under the headnote section, above the TOC:

image

Edit resource page, at the bottom:
image

app/assets/stylesheets/fonts.scss
```diff
@@ -22,25 +23,29 @@ $light: 300;
...
 @font-face {
-  @include load-font($font-family-sans-serif, normal, $light, 'AtlasGrotesk-Light');
+  @include load-font($preferred-sans-serif, normal, $light, 'AtlasGrotesk-Light');
 }
 
-$font-family-serif: 'Chronicle Text G3', Georgia, 'Times New Roman', Times, serif;
+
 @font-face {
-  @include load-font($font-family-serif, normal, $regular, 'ChronicleTextG3-Roman');
+  @include load-font($preferred-serif, normal, $regular, 'ChronicleTextG3-Roman');
```
cgruppioni: Love the name change

app/views/content/_edit_details.html.erb
```diff
@@ -41,3 +41,9 @@
     </div>
   <% end %>
 <% end %>
+
+<% if @casebook.present? %>
+  <aside class="casebook-actions">
+    <%= render 'content/actions' %>
```
cgruppioni: I really like this move here, makes a lot of sense

app/views/content/show.html.erb
```diff
@@ -37,7 +36,7 @@
       <%= @content.title %>
     <% end %>
   </h1>
-    <% if @section.nil? %>
+    <% if @section.nil? && @content.subtitle? %>
```
cgruppioni: nice

PR #363, opened 2018-05-30
First pass, elision accessibility

rebeccacremona:
This makes the button that reveals/hides elided text into an html button so that it is accessible by keyboard, and it adds a label for screen reader users. It also adds a little hidden note at the end of elided text notifying screen reader users that they've reached the end of elided text. 

This still lacks some important features: all the buttons have the same label (how could I make them different? Paragraph number might be good, but I can't figure out how to expose paragraph number to the annotation code....), and, the labels are quite different visually and in the markup, which can/will be confusing both for people using Dragon Speak Naturally or similar, and for screen reader users following along with the H2O docs, which will certainly reference the visible label. But, I think I got as far as we can without major surgery.

While I was in there, I took a pass at hiding the borders of blockquotes, which currently display even when elisions are hidden. If an elision starts inside an inline element, but also includes a blockquote, it doesn't quite work, due to the way elisions are recorded, but now, you'll see at most one stray border, instead of a list.

before:
image

after:
image

`<br>` and other block-level elements inside elisions might/will still leak through sometimes, but I don't think it's something I can fix without major surgery.

While I was in there, I also set paragraph numbers to display when elided text is revealed.

### DIFF
--- a/app/assets/javascripts/lib/ui/content/annotations/elide.js
+++ b/app/assets/javascripts/lib/ui/content/annotations/elide.js
@@ -19,5 +19,6 @@ delegate(document, '.annotate.elide', 'click', e => {
   for (let el of elisions) {
     el.classList.toggle('revealed');
     el.parentElement.classList.toggle('revealed');
+    el.parentElement.previousElementSibling.classList.toggle('revealed');
   }
 });
--- a/app/assets/javascripts/lib/ui/content/annotations/elide.js
+++ b/app/assets/javascripts/lib/ui/content/annotations/elide.js
@@ -18,5 +18,6 @@ delegate(document, '.annotate.elide', 'click', e => {
   }
   for (let el of elisions) {
     el.classList.toggle('revealed');
+    el.parentElement.classList.toggle('revealed');
   }
 });
--- a/app/assets/javascripts/lib/ui/content/annotations/elide.js
+++ b/app/assets/javascripts/lib/ui/content/annotations/elide.js
@@ -8,7 +8,14 @@ delegate(document, '.annotate.elide', 'click', e => {
   let annotationId = e.target.dataset.annotationId;
   let elisions = document.querySelectorAll(`.annotate.elided[data-annotation-id="${annotationId}"]`);
 
-  e.target.classList.toggle('revealed')
+  e.target.classList.toggle('revealed');
+  if (e.target.classList.contains('revealed')){
+    e.target.setAttribute('aria-expanded', 'true');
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/lib/ui/content/annotations/elide.js
```diff
@@ -8,8 +8,17 @@ delegate(document, '.annotate.elide', 'click', e => {
   let annotationId = e.target.dataset.annotationId;
   let elisions = document.querySelectorAll(`.annotate.elided[data-annotation-id="${annotationId}"]`);
 
-  e.target.classList.toggle('revealed')
+  e.target.classList.toggle('revealed');
+  if (e.target.classList.contains('revealed')){
```
cgruppioni: Would this always be true since it was set to `revealed` in the previous line?

app/assets/stylesheets/content/annotations.scss
```diff
@@ -79,6 +79,17 @@
     padding: 3px 0 3px;
   } // etc
 
+  /* hacks for misbehaving blockquotes */
```
cgruppioni: super awesome

app/models/content/annotation.rb
```diff
@@ -94,7 +94,8 @@ def exists_in_published_casebook?
   def annotate_html inner, handle: true, final: false
     case kind
     when 'elide' then
-      "#{handle ? "<span class='annotate elide' data-annotation-id='#{id}'></span>" : ''}<span class='annotate elided' data-annotation-id='#{id}'>#{inner}</span>"
+      "#{handle ? "<button class='annotate elide' data-annotation-id='#{id}' aria-label='elided text' aria-expanded='false'></button>" : ''}" +
```
cgruppioni: Is `aria-label` where you were talking about all of your labels being the same?

app/assets/javascripts/lib/ui/content/annotations/elide.js
rebeccacremona: Nope, actually! 

I love toggle: if the class is present, toggle removes it. If the class is NOT present, toggle adds it. This event handler is called every time you click the button, so it has to handle BOTH when you want to reveal the elided text and when you want to hide it. Since there isn't a toggle method for the other stuff, I had to make a separate if/else.

app/models/content/annotation.rb
rebeccacremona: Yep, exactly. Without a `<label>` for the button, and without any text inside the button, this just gets read as... silence... by screen readers. `aria-label` is a way of providing a label for screen readers without changing the visual appearance of the page.

rebeccacremona:
@cgruppioni Was your concern addressed?

cgruppioni:
@rebeccacremona Yes thanks for the bump

Issue #303, opened 2018-04-30
Annotations: replace text


abziegler:
Cursor should change to show action available or should auto-click into field

cgruppioni:
Status notes:
* cursor changed 
* I'm not sure if auto-clicking into field is possbile because it's a `<span>` not a form or input box

cgruppioni:
PR #365 solves changing the cursor and passing in the annotation_id to the URL params. 
The issue will be fully resolved in combination with another ticket

Issue #295, opened 2018-04-30
Annotations: after annotation, should not jump to top of document


abziegler:
I made an elision in a case. After submitting the annotation, page reloaded and jumped to top of document, losing my location. 

What should happen is window should stay at location of annotation after annotation is submitted

This may be related to Turbolinks and/or to this commit: https://github.com/harvard-lil/h2o/commit/51f8ecb3b98ace35620fe52419016572845204aa

cgruppioni:
- Doesn’t jump on Chrome (very smooth) (edited)
- Doesn’t jump on Safari (with some slightly weird page reload behavior) (edited)
- Does jump on Firefox Developer edition incognito
- No annotation menu at all on regular Firefox

cgruppioni:
![screen shot 2018-05-31 at 11 05 28 am](https://user-images.githubusercontent.com/14900328/40790297-8ded346c-64c2-11e8-83cf-4b345597ee1c.png)
I'm getting the menu on regular firefox

bjohnson747:
@abziegler @cgruppioni based on my work on yesterday #362 , Firefox 36.0.1 gave me no menu but when i uninstalled FF and reinstalled it from the web (v60.0.1), i got a menu. For some reason my automatic update didn't work, i had to reinstall FF.

Looks like right now, having an un-updated Firefox (and possible Safari) browser can prevent full H2O functionality.

cgruppioni:
Waiting on a PR from Becky

rebeccacremona:
As of https://github.com/harvard-lil/h2o/commit/ce09ef4a0bc5db7c78d144d76d652bd3501be5e2, this should work now...... Let's hope and let's test!

rebeccacremona:
That is to say, the focusing on replacement annotations part. No idea what's up with the missing menu on Safari, didn't look at that.

PR #365, opened 2018-05-31
Annotations replace pointer & Firefox location 

cgruppioni:
#295  & #303

### DIFF
--- a/app/assets/javascripts/lib/requests.js
+++ b/app/assets/javascripts/lib/requests.js
@@ -17,7 +17,13 @@ export function request (url, method, data = {}, options = {scroll: true}) {
       let location = response.request.responseURL;
 
       if ((window.location.href == location) || (method == 'delete')){
-        window.location.reload(true); // save scroll position for annotations
+        // saving scroll position 
+        if (navigator.userAgent.match('Firefox') != null){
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/lib/ui/content/annotations/annotator.js.erb
```diff
@@ -107,6 +107,7 @@ export class Annotator extends Component {
       { annotation: attrs }, { scroll: false })
     .then( _ => this.deactivate() );
 
+    window.location.hash = `annotation_id=${annotationId}`
```
cgruppioni: Try to add as params instead

app/assets/javascripts/lib/requests.js
```diff
@@ -17,7 +17,13 @@ export function request (url, method, data = {}, options = {scroll: true}) {
       let location = response.request.responseURL;
 
       if ((window.location.href == location) || (method == 'delete')){
-        window.location.reload(true); // save scroll position for annotations
+        // saving scroll position 
+        if (navigator.userAgent.match('Firefox') != null){
```
cgruppioni: Not sure why this works/is necessary but it does. Must be something going on deeper behind the scenes...

PR #372, opened 2018-06-01
Replace button a11y first pass

rebeccacremona:
I could not make this into a real html button. I could not get contenteditable to work reliably across browsers in combination with real buttons: hitting the spacebar presses the button, despite all the fancy javascript I could think of. It might be worth another try later... Instead, I added a bunch of stuff to make the `span` behave like a real button: added it to the tab order using tabindex, added event handlers for space and enter (which real buttons have automagically, when you assign a click handler), and added the `button` aria role. Good enough, methinks.

So, now it can receive focus. That's an accessibility win, and should help with the problem you are having related to page refreshes, until there's time to move to AJAX: we can use JS to set focus to an annotation on page load, if an annotation-id is specified in the query string. This PR doesn't do anything to get an annotation-id into the query string, but it DOES include code for setting focus, if one is there. You can add one manually in your browser's URL bar to see it work.

As with the elision PR, there's still some work to do to improve the experience for people using screen readers. It's not clear that the button, not pressed, is replacement text, but when pressed, says "hide" and is followed by original text. It would be better if this button behaved like the regular elision button, with some sort of clear label, and the button was followed by either the replacement text or the original text, depending on whether the button had been pressed or not. But that's relatively major surgery, and a design change: the whole flow + contenteditable would change. TBD when convenient, later.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,6 +16,7 @@
 //= require 'jquery-readyselector'
 //= require 'polyfills'
 //= require 'lib/bootstrap/dropdown'
+//= require 'lib/helpers'
 //= require 'lib/requests'
 //= require 'lib/ui/skip-link-focus-fix'
 //= require 'lib/ui/content'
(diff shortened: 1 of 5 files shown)


rebeccacremona:
Oh, there are merge conflicts now. Will resolve.

rebeccacremona:
Okay merge conflicts resolved.

PR #377, opened 2018-06-04
tweaked modals

rebeccacremona:
This take the code from `dashboard.js.erb` that made THAT modal accessible, and abstracts it out into its own class, a subclass of Component.  So, now you can make any modal accessible by just inheriting from ModalComponent instead of Component (and making sure your template includes a tabindex of -1 at the top, and a label). 

This PR switches the clone modal and the publish modal to inherit from ModalComponent. Unfortunately I didn't get to any of the others yet.

This also makes the export button a one-click thing: export immediately instead of triggering a modal window.

### DIFF
--- a/app/assets/javascripts/lib/ui/content/clone_modal.js
+++ b/app/assets/javascripts/lib/ui/content/clone_modal.js
@@ -1,27 +1,16 @@
 import {html} from 'es6-string-html-template';
 import delegate from 'delegate';
-import Component from 'lib/ui/component';
+import ModalComponent from 'lib/ui/modal';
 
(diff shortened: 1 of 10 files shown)


rebeccacremona:
Okay...... in order for the tests to work without using `window.open`, which has some problematic behavior, I had to restructure my earlier export refactor. It's not as clean now, but it still works with a single click, which is probably still an improvement for now. 

If, as Adam suggests, people will eventually be able to choose different layouts or whatever, the modal will come back, and this weird-looking code will be obsolete, and we can readdress testing strategy at that point!

PR #380, opened 2018-06-05
annotation-id in URL 

cgruppioni:
Create annotation as json and return annotation_id to put in href query string.

### DIFF
--- a/app/assets/javascripts/lib/ui/content/annotations/annotator.js.erb
+++ b/app/assets/javascripts/lib/ui/content/annotations/annotator.js.erb
@@ -108,8 +108,6 @@ export class Annotator extends Component {
     Axios.patch(this.annotationPath(annotationId + '.json'), 
       { annotation: attrs }, { scroll: false })
     .then( _ => this.deactivate() );
-
-    window.location.search = `annotation-id=${response.data.annotation_id}`;
   }
 
   deactivate () {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/lib/ui/content/annotations/annotator.js.erb
```diff
@@ -107,7 +109,7 @@ export class Annotator extends Component {
       { annotation: attrs }, { scroll: false })
     .then( _ => this.deactivate() );
 
-    window.location = (window.location + `?annotation_id=${annotationId}`);
+    window.location = (window.location + `?annotation-id=${annotationId}`);
```
cgruppioni: this is under updateAnnotation() --- does it need to be here?

app/assets/javascripts/lib/ui/content/annotations/annotator.js.erb
```diff
@@ -107,7 +109,7 @@ export class Annotator extends Component {
       { annotation: attrs }, { scroll: false })
     .then( _ => this.deactivate() );
 
-    window.location = (window.location + `?annotation_id=${annotationId}`);
+    window.location.search = `annotation-id=${response.data.annotation_id}`;
```
cgruppioni: if the annotation-id query string isn't needed in updateAnnotation lmk and I can remove it

PR #389, opened 2018-06-12
Remove outdated associations in RailsAdmin.

cgruppioni:


bjohnson747:
All this looked good to me - my only question, in the edit user area, there were several checkboxes, including 'print annotations' (which defaulted to unchecked). Are these just things pulled in from old H2O or do they impact user export/display experience?

screen shot 2018-06-14 at 3 05 55 pm

bjohnson747:
All this looked good to me - my only question, in the edit user area, there were several checkboxes, including 'print annotations' (which defaulted to unchecked). Are these just things pulled in from old H2O or do they impact user export/display experience? 

screen shot 2018-06-14 at 3 05 55 pm

Issue #302, opened 2018-04-30
Annotations: notes


abziegler:
- Notes should default to open / displayed
- Notes should not cause highlighting of text
- Menu should offer "Remove Note" rather than "Restore Text" and "Elide Text"
- Note should align with top of selected text rather than bottom of selected text

![h2o-notes](https://user-images.githubusercontent.com/2140548/39444661-7fb4e6fa-4c86-11e8-9b8d-33ce645c1cf4.png)

cgruppioni:
What do you want this to look like since it doesn't have highlighting? 
![image](https://user-images.githubusercontent.com/14900328/40789478-80f9ffbc-64c0-11e8-84e6-b3ff33d8eca1.png)

( the icon has the yellow background built into it, so would have to create a new icon or use something else? )

cgruppioni:
What do you want the placement of the note next to the icon to be like? 

![image](https://user-images.githubusercontent.com/14900328/40789534-9e9d8494-64c0-11e8-86ed-5eebfb21e31c.png)

![image](https://user-images.githubusercontent.com/14900328/40789542-a2329838-64c0-11e8-88f1-b0e304282c95.png)

cgruppioni:
https://slack-files.com/T02RW19TT-FB0MU26CQ-d8e979e681

https://slack-files.com/T02RW19TT-FB16A8U04-9879743f6d

cgruppioni:
![h2o-annotation-comment-closed](https://user-images.githubusercontent.com/14900328/40937201-d5e8217c-680b-11e8-8016-41dff7046c1e.png)
![h2o-annotation-comment-open](https://user-images.githubusercontent.com/14900328/40937202-d5f6b0a2-680b-11e8-9593-eda5f975cc29.png)

cgruppioni:
+ select all text with that annotation id
+ change double underline to more solid

abziegler:
Still seeing these same styling concerns:

![image](https://user-images.githubusercontent.com/2140548/41435473-e66a856a-6fec-11e8-81fa-ee6b09c8ef9c.png)

PR #394, opened 2018-06-13
Annotation note styling 

cgruppioni:
#302


bjohnson747:
This all looked good, except for one visual thing - the text block i'm typing in is just a touch larger than the box that's around it. (Same for both Safari and Chrome)
screen shot 2018-06-14 at 2 34 16 pm

PR #397, opened 2018-06-14
Annotation notes 2

cgruppioni:
- Remove double underline from text that is selected for open notes
- selected text is equally selected amongst paragraph lines
- the new annotation menu doesn't pop up when clicking the selected text


bjohnson747:
These things all acted as described here 👍

PR #422, opened 2018-06-22
Use retrofitted spans instead of buttons so docx export works.

rebeccacremona:
https://github.com/harvard-lil/h2o/commit/7c335c41a4357c2a7124ec9611cfd3810c50ade9#diff-f87f5ba9f384eda55270ae286585b500L97 switched from plain spans to html buttons so that elisions could be toggled by people using input devices other than mice.

The export code is looking for spans (https://github.com/harvard-lil/h2o/blob/master/lib/htmltoword/xslt/h2o/export.xslt#L133), so this transformation wasn't run any more. I do not know why this test (https://github.com/harvard-lil/h2o/blob/master/test/system/export_test.rb#L21) did not catch the change.

Switching the export code to look for buttons instead of spans creates corrupt docx files. 

`¯\_(ツ)_/¯`

This PR switches back to a span, but retrofits the span to behave like a button.

### DIFF
--- a/app/assets/javascripts/lib/ui/content/annotations/elide.js
+++ b/app/assets/javascripts/lib/ui/content/annotations/elide.js
@@ -4,11 +4,20 @@ import throttle from 'lodash.throttle';
 import Component from 'lib/ui/component'
 import delegate from 'delegate';
 
-delegate(document, '.annotate.elide', 'click', e => {
+// Respond to click, spacebar, or enter, like a real html button would
+delegate(document, '.annotate.elide', 'click', e => handleElideButtonPressed(e));
+delegate(document, '.annotate.elide', 'keypress', e => {
+  if (e.key=='Enter'||e.key==' '||e.keyCode==13||e.keyCode==32){
+    e.preventDefault();
+    handleElideButtonPressed(e);
+  }
+});
+
+function handleElideButtonPressed(e){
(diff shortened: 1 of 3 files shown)


rebeccacremona:
For future debugging efforts, here's what happened when I switched the xslt selector to `button`:
```
$ diff corrupt.docx valid.docx 
50,55c50,65
< </w:p><w:r xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
<   <w:rPr>
<     <w:rStyle w:val="Elision"/>
<   </w:rPr>
<   <w:t xml:space="preserve">[ ... ]</w:t>
< </w:r><!--
---
> </w:p><!--
>         In the following situation:
> 
>         div
>           h2
>           span
>             textnode
>             span
>               textnode
>           p
> 
>         The div template will not create a w:p because the div contains a h2. Therefore we need to wrap the inline elements span|a|small in a p here.
>       -->
> <w:p xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main" xmlns:pkg="http://schemas.microsoft.com/office/2006/xmlPackage" xmlns:str="http://exslt.org/strings" xmlns:fn="http://www.w3.org/2005/xpath-functions">
>   <!--span not ancestor-->
> </w:p><!--
```

Issue #285, opened 2018-04-30
Dashboard: display only title, author, school and attribution


abziegler:
Either we should not display subtitles or we should limit characters or handle overlap

![h2o-funky-dashboard](https://user-images.githubusercontent.com/2140548/39432111-a21c91c8-4c60-11e8-9cf5-ed1aac49774e.png)

leppert:
Okay to broaden the scope of this ticket to deal with text truncation generally? 
<img width="1253" alt="screen shot 2018-06-19 at 9 15 35 pm" src="https://user-images.githubusercontent.com/26643/41632155-84657938-7406-11e8-8345-3ea63b960d83.png">

leppert:
@abziegler @cgruppioni bump

abziegler:
Yes, absolutely @leppert. Let me know if you need any add'l input from us on how this might affect more than dashboard/bookshelf display.

PR #425, opened 2018-06-23
285 handle reasonable text lengths

leppert:
#285 
image

### DIFF
--- a/app/assets/stylesheets/mixins.scss
+++ b/app/assets/stylesheets/mixins.scss
@@ -65,3 +65,13 @@
   float: none !important;
   margin: auto;
 }
+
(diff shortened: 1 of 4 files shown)


leppert:
@cgruppioni Doesn't look like I have permission to add you as a reviewer, unless I'm missing something. 
image

app/assets/stylesheets/mixins.scss
```diff
@@ -65,3 +65,13 @@
   float: none !important;
   margin: auto;
 }
+
+@mixin line-clamp($lines, $max-height) {
```
cgruppioni: :+1:

Issue #325, opened 2018-05-04
New sections created in a subsection are placed at the top-level section instead


bjohnson747:
When user's view is within a subsection, clicking Add Section should add the new section to the subsection they're viewing: currently, it adds it to the top level of the casebook.

Single section in new casebook:
<img width="783" alt="single section in new casebook" src="https://user-images.githubusercontent.com/12850389/39643748-c6ff54da-4fa3-11e8-9e63-24e4cbbb7c19.png">

I attempted to add new section when in that view, but it added it to the main level of the casebook (Section Two):
<img width="830" alt="new section created is added to top-level" src="https://user-images.githubusercontent.com/12850389/39643747-c6ec3e68-4fa3-11e8-8de5-f296a5929e54.png">

PR #440, opened 2018-06-28
Remove improper @ scoping

leppert:
#325


leppert:
@cgruppioni Travis is failing because it's pointing to a nonexistent branch, despite Github using the correct info. Not sure what to do about that other than recommend you pull down locally and run the tests there.

PR #464, opened 2018-07-23
Add catchall to render 404s

### LINKED ISSUE #344: Deprecation Warning
`DEPRECATION WARNING: Using a dynamic :controller segment in a route is deprecated and will be removed in Rails 5.2. (called from block in <top (required)> at /<path-to-my-repo>/h2o/config/routes.rb:124)`

leppert:
This also resolves a deprecation warning for Rails 5.2. Closes #344.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -119,6 +119,5 @@ H2o::Application.routes.draw do
 
   get "/pages/*id" => 'pages#show', as: :page, format: false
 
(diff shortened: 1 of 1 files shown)


leppert:
@cgruppioni Unknown routes were previously throwing 500's from what I can tell. Some orgs get into a weird habit of using error reports to see what URLs people are guessing; would simply want to make sure H2O isn't depending on that. cc/ @abziegler

Issue #421, opened 2018-06-21
Fix signup and email confirmation flow


bjohnson747:
Testing user clicked the link in his email, which brought him back to opencasebook.org but it failed to verify his email address.
This is the message recieved when attempted:

<img width="1202" alt="screen shot 2018-06-21 at 6 03 49 pm" src="https://user-images.githubusercontent.com/12850389/41747883-91c822ec-757d-11e8-83ee-5d9806f2506b.png">

bjohnson747:
Also the "youve been sent a verification email" shows up half-covered, at least in Chrome 
<img width="752" alt="screen shot 2018-06-21 at 5 44 42 pm" src="https://user-images.githubusercontent.com/12850389/41747483-27058ffe-757c-11e8-89e9-847934208ad8.png">

bjohnson747:
Additionally - Display Name (attribution) is 'optional' this should be set to required. Without a display name their casebooks display with no owner.

cgruppioni:
Part 2: 

- [ ]  Clicking on the confirmation email link takes you to a view on which you create your password ( this is how perma works)

- [ ] "youve been sent a verification email" shows up half-covered, at least in Chrome https://user-images.githubusercontent.com/12850389/41747483-27058ffe-757c-11e8-89e9-847934208ad8.png

- [ ] confirmation email link doesn't do anything // says verification failed

PR #467, opened 2018-07-23
Update sign up flow 

cgruppioni:
User enters only their email address in the sign up form and then receives an email with a link to verify their account + fill in their password. 

They click the link that was emailed to them, fill in their password and password confirmation and then are logged in 

#421 

Deploy instructions: 

Run `User.where(attribution: nil).update_all(attribution: "Anonymous")` ***[Note: This has already been done in staging & prod]***


app/controllers/users_controller.rb
```diff
@@ -11,27 +11,17 @@ def index
...
-    else
-      flash[:notice] = 'Your account has not been verified. Please login and try visiting the link in the email again.'
-      redirect_to '/user_sessions/new'
-    end
-  end
-
   def create
-    @user = User.new(permitted_user_params)
+    user_params = permitted_user_params
+    temporary_password = Random.new_seed
```
cgruppioni: Authlogic has the variable `require_password_confirmation`. 

To save a user without having a password it'd have to be set to false. But then, when the user does add a password there is no such thing as the password_confirmation field. 

So I added a temp password to the newly created user account so that we still have the password + password confirmation fields

app/mailers/notifier.rb
```diff
@@ -9,13 +9,9 @@ def password_reset_instructions(user)
...
   def verification_request(user)
-    @verification_url = verify_user_url(user, token: user.perishable_token)
+    @verification_url = edit_password_reset_url(user.perishable_token)
     @user_name = user.display_name
-    mail(to: user.email_address, subject: "H2O: Verify your email address")
-  end
-
-  def verification_notice(user)
-    mail(to: user.email_address, subject: "Welcome to H2O. Your account has been verified")
+    mail(to: user.email_address, subject: "A H2O account has been created for you")
```
jcushman: Typo: should be "An" H2O account

app/views/notifier/password_reset_instructions.erb
```diff
@@ -1,8 +1,6 @@
 A request to reset your password has been made.
 If you did not make this request, simply ignore this email.
-If you did make this request just click the link below:
-
-<%= edit_password_reset_url(@token) %>
+If you did make this request <%= link_to "click here to reset it", edit_password_reset_url(@token) %>.
 
 If the above URL does not work try copying and pasting it into your browser.
 If you continue to have problem please feel free to contact us.
```
jcushman: Unrelated typo since I happen to be reading: should be "If you continue to have problems"

PR #479, opened 2018-07-30
Remove vestigial karma property

leppert:


db/migrate/20180719190825_add_default_to_users_attribution.rb
```diff
@@ -1,7 +1,9 @@
 class AddDefaultToUsersAttribution < ActiveRecord::Migration[5.1]
```
cgruppioni: Ah! Nice thank you

Issue #307, opened 2018-04-30
Export: Notes and Links


abziegler:
On export, notes and links added to a resource via annotation are not present in the export file.  They should be present, though not sure where/how. Possibly as ** following the resource?

cgruppioni:
Notes/links show up on bottom of page or bottom of document?

abziegler:
Page is preferred.

cgruppioni:
- [ ] notes and links show an asterick where the annotation lives (maybe have to do numbers?) 
- [ ] notes and links show on bottom of page, full text with the same number of astericks

show up on bottom of the case

cgruppioni:
/Users/cgruppioni/lil/h2o/lib/htmltoword/xslt/h2o/export.xslt
/Users/cgruppioni/lil/h2o/lib/htmltoword/xslt/base.xslt

https://www.mulberrytech.com/quickref/XSLT_1quickref-v2.pdf
https://docstore.mik.ua/orelly/xml/xmlnut/ch08_07.htm
https://www.w3schools.com/xml/xsl_elementref.asp

Issue #367, opened 2018-05-31
AJAX to make saving things and annotations more fluid 


cgruppioni:

cgruppioni:
https://github.com/harvard-lil/h2o/pull/372

cgruppioni:
One example: Editing link annotations (e.g. by changing the URL) is a little clunky. When you enter the new URL, then click Enter, the form disappears, but you have to manually refresh to confirm that the new URL was saved.

PR #484, opened 2018-08-01
Export notes and links

cgruppioni:
#307 

- Note and link 'footnotes' are shown at the bottom of a resource (this goes for a single resource exported file, a section or a casebook)

- created the applyAnnotationToParagraphs service class to remove complicated logic from the annotation model and make the code a lot more readable
- deleted and/or moved outdated export css and js files to the old_tree folder. All of the export css code wasn't actually being used, only xslt stylesheets are used for word exports.
- export_footnote_index (in the resource model) keeps track of the number of asterisks an annotation should have 
- changed Annotation column names to be more verbose
-- start_p is now start_paragraph
-- end_p is now end_paragraph

Example: 
screen shot 2018-08-01 at 5 18 16 pm
^ Editor view so you can see where the notes and links are 

screen shot 2018-08-01 at 5 17 29 pm
^ 2 notes and 1 link annotation are shown and asterick'd 

screen shot 2018-08-01 at 5 17 39 pm
^ bottom of the resource


app/models/content/resource.rb
```diff
@@ -61,4 +66,18 @@ def annotated_paragraphs editable: false
   def title
     super || resource.title
   end
+
+  def footnote_annotations
```
jcushman: This is not specific to this code review, but I have a general suggestion. What if you pick a commenting style and start adding comments for new code?

I like [the policy here](https://github.com/airbnb/ruby#commenting) in AirBnB's style guide, which basically recommends a comment describing the purpose of each class, a comment describing the purpose and input/output of each function, and a comment describing the purpose of any complicated block of code. That's a lot like what I do in CAP -- [here's an example function](https://github.com/jcushman/capstone/blob/5c39f23f6b723dc2f923795ae87ac40ac37c1fed/capstone/scripts/make_pdf.py#L49).

The idea here is if you know the few lines of code that are causing a bug, it's usually obvious how to fix it. But if you have 200 lines of code you haven't looked at in a year, you don't want to slog through and figure out the point of them all in order to find the few lines that matter to the bug you're fixing. You want section headings to scan through.

So here I might add a comment before the function like

```
# Get string representation of note and link annotations, 
# formatted like "* Note 1 ** Note 2 *** Note 3 ".
```

When I do this I tend to do it as the very last step before sending a pull request, with only minimal commenting as I go along, because I don't want to rewrite stuff as I mess around. But then before committing, I go through and make sure that the new code is sorted in blocks that make sense, variable names still make sense (because often the meaning changes as I go along), and each function or block has a short comment that will explain what it's for when I come back in a year.

Totally up to you how much of that is helpful -- but my hunch is that adding some block-level comments would pay off as you go along.

cgruppioni:
Testing Plan

First test that all annotation functionality is still fully functional
- [ ] Compare a casebook from opencasebook.org to one on staging and make sure all annotations still look the same
- [ ] Create new annotations and they show up on the correct lines that you intended to annotate. 
- [ ] All annotation interactions are still functional, ie: editing, deleting, etc. 

Added functionality

- [ ] In export: lines that have note and link annotations show up with an asterisk following it 

- [ ] If you scroll to the bottom of the resource the corresponding astericks are there (like a footnote) with the correct content

- [ ] 'Footnote annotations' show up on the bottom of the individual resource, not the bottom of the entire section or casebook

abziegler:
@cgruppioni I tested annotation comparison between staging and prod, and it fails

This is prod (https://opencasebook.org/casebooks/47933-criminal-law-spring-2016-h2o-export-test-version/resources/2.1.4-brian-valenti-on-his-own-behalf-and-on-behalf-of-a-class-of-those-similarly-situated-plaintiff-v-hartford-city-indiana-defendant)

h2o-annotations-prod

This is staging (../casebooks/47933-criminal-law-spring-2016-h2o-export-test-version/resources/2.1.4-brian-valenti-on-his-own-behalf-and-on-behalf-of-a-class-of-those-similarly-situated-plaintiff-v-hartford-city-indiana-defendant) 

Not only are the annotations different, some of the text on staging is rendering in duplicate

h2o-annotations-staging

bensteinberg:
@abziegler and @cgruppioni, I will try to confirm that the data move to stage was correct.

abziegler:
@cgruppioni I've tested all of this (wow!), and only one thing looks a little funky ... and was already the case in opencasebook

- Editing link annotations (e.g. by changing the URL) is a little clunky. When you enter the new URL, then click Enter, the form disappears, but you have to manually refresh to confirm that the new URL was saved.

Everything else looks good.

cgruppioni:
@abziegler Awesome!! I'm going to make a note of that and I can fit it in with the next round of annotation work I'm doing. Thank you!

cgruppioni:
@abziegler talked to Becky and she said it would be a good thing for Ajax to fix so I added it to our running ticket https://github.com/harvard-lil/h2o/issues/367

Issue #478, opened 2018-07-30
Subsections displaying at incorrect 'depth' when in draft mode


bjohnson747:
When viewing a section layout (rather than entire case layout) for a draft casebook, sub-section indentation is not correct. In published mode, sub-section indentation is correct:

Published mode:

<img width="752" alt="screen shot 2018-07-30 at 3 12 33 pm" src="https://user-images.githubusercontent.com/12850389/43418212-476594ee-940b-11e8-9bfb-9de5105ad732.png">

But when in draft mode, subsections that are actually the same level appear to be subsections of each other. 
Here's the same sections in draft mode:
<img width="755" alt="screen shot 2018-07-30 at 3 12 55 pm" src="https://user-images.githubusercontent.com/12850389/43418264-5ea71bf0-940b-11e8-888a-4773c8336048.png">

abziegler:
@bjohnson747 can you link to casebook or provide ID?

bjohnson747:
@abziegler Not limited to just this casebook but these examples are from casebook #28258

abziegler:
@bjohnson747 Happening only with imported/migrated casebooks, or also with new casebooks?

bjohnson747:
-Only happens when viewing a specific section, not at full casebook view.

-Only occurs if there is a resource between one section and the next.
CORRECT:
<img width="734" alt="screen shot 2018-07-31 at 2 19 42 pm" src="https://user-images.githubusercontent.com/12850389/43480073-287e1342-94d0-11e8-822d-fa8a491c2085.png">

INCORRECT:
<img width="705" alt="screen shot 2018-07-31 at 2 20 33 pm" src="https://user-images.githubusercontent.com/12850389/43480079-2ceec7dc-94d0-11e8-818a-b0a381602efa.png">

-I can only re-create this at the first-subsection level. Red-boxed below are the ones that are wrong but the rest are correct.
<img width="700" alt="screen shot 2018-07-31 at 3 58 44 pm" src="https://user-images.githubusercontent.com/12850389/43484124-759c22f8-94db-11e8-99a2-00c46487c975.png">

PR #493, opened 2018-08-07
Fix section depth on layout view for draft casebooks

abziegler:
This PR attempts to address #478. I'm a bit out of my comfort zone on this one @cgruppioni, so you may want to give this one a very close look when you can. It does work locally for me.

- Rewrote the layout/edit view from HAML to ERB
- Added the section depth logic from the layout/show view
- Modified the main layout template to pull in the right data (I think)

Note: there is one bit of code in the old view that I couldn't make sense of. On Line 2 of the old view there was `- content = edit`. I did not include that in the ERB and cannot tell if what it is doing.


app/views/content/table_of_contents/_edit.erb
```diff
@@ -0,0 +1,84 @@
...
+            <button class="action-delete"></button>
+          </div>
+        <% end %>
+      </div>
+    <% end %>
+  <% elsif content.is_a? Enumerable %>
+      <div class="section-wrapper">
+      <% if editable %>
+        <div class="section-contents editable">
+          <%= render 'content/table_of_contents/edit', collection=content %>
```
cgruppioni: collection: content ( and on line 73)

app/views/content/table_of_contents/_edit.erb
```diff
@@ -0,0 +1,84 @@
+<% editable = true %>
+<% start_depth = contents.first.try(:ordinals).try(:length) %>
+<% current_depth = start_depth %>
+
+<% contents.each do |content| %>
+  <%# render separate opening and closing tags around sections %>
+  <% (1..content.ordinals.length - current_depth).each do %>
```
cgruppioni: I pulled the code down to try to figure this out more. If it's the first resource in a section insert the <div> tags on top of it?

cgruppioni:
Do you know why this was only happening for sections and not for casebooks?

abziegler:
@cgruppioni I don't know why. Good question. 

What's happening is, in draft mode, sections are progressively nesting / increasing their depth, except when the full casebook contents are displayed.

PR #500, opened 2018-08-13
Revise button yml - published casebook without a draft

cgruppioni:


bjohnson747:
This looked good to me on staging

PR #501, opened 2018-08-13
143 capapi integration

leppert:


leppert:
A few things to note:

- Rails and Puma have been updated
- A bunch of old_tree JS has been deleted; I left the code that was referenced by export (for now)
- There are data migrations in with the schema migrations. Different folks have different feelings about this — where to put data migrations — but IMO they work here because they're tightly coupled to the schema changes and are largely reversible.

abziegler:
Couple of notes about cases successfully imported from CAPAPI:

- Case text isn't annotatable -- highlighting, elisions, replacement text, comments, links all fail
- Paragraph numbering isn't being applied across entire case text.

app/assets/javascripts/lib/ui/content/add_content.js.erb
```diff
@@ -81,7 +79,20 @@ class AddResourceModal extends Component {
           e.stopPropagation();
           e.preventDefault();
 
-          post('<%= sections_path('$CASEBOOK_ID') %>'.replace('$CASEBOOK_ID', this.casebookId()), {resource_id: e.delegateTarget.dataset.resultId, parent: this.sectionId()}, {modal: this})
+          const add = id => {
```
cgruppioni: circle back

app/views/shared/objects/_pending_case.html.erb
```diff
@@ -2,7 +2,7 @@
   <a href="#" class="tooltip list-icon icon icon-case" title="Pending Case"></a>
```
cgruppioni: This file could probably be deleted?

config/environments/development.rb
```diff
@@ -60,7 +60,7 @@
   config.user_verification_recipients = ['cgruppioni@law.harvard.edu']
 
   config.admin_emails = ['cgruppioni@law.harvard.edu']
-  # config.cap_api_key = '2c62c54b47e507b2eee20a70f29f1b4ae0ccd1a3'
```
cgruppioni: hah

leppert:
Here's the Travis-CI that should be checked for this PR since Travis doesn't update the LIL version of this PR with new commits once the PR is created.

https://travis-ci.org/leppert/h2o

test/controllers/searches_controller_test.rb
```diff
@@ -0,0 +1,36 @@
+# coding: utf-8
+require 'test_helper'
+
+class SearchesControllerTest < ActionDispatch::IntegrationTest
```
cgruppioni: 🙏

leppert:
@abziegler Annotating should now be fixed in capapi cases, and I believe this branch is ready to merge. The fix is that I'm unwrapping some of the new capapi HTML tags here:
https://github.com/leppert/h2o/blob/143-capapi-integration/lib/html_helpers.rb#L10

Previously we only unwrapped divs:
https://github.com/harvard-lil/h2o/blob/master/app/models/content/resource.rb#L31

There is, however, a change I'm unsure about. After the various HTML processing, I found that capapi cases would sometimes have tags containing only whitespace that caused extra paragraph numbering. As a fix, I'm now removing tags that are empty or contain only whitespace:
https://github.com/leppert/h2o/blob/143-capapi-integration/lib/html_helpers.rb#L34

This replaces previous logic that would only delete nodes that were completely empty, leaving nodes containing only whitespace:
https://github.com/harvard-lil/h2o/blob/master/app/models/content/resource.rb#L17-L21

The rub is that, because this has an effect on paragraph numbering (a positive effect, in the case of capapi cases) I'm not sure if it'll mess up any existing annotations. I'm thinking I'll write a script to check existing cases for any html differences between the new formatter and the old one; any testcases you might have will be welcomed.

abziegler:
Annotations work great. I tried all different types, viewed them in all modes, modified them various ways, with multiple cases imported from CAPAPI.

abziegler:
@leppert I don't know if we have existing cases with whitespace-only tags, but I would assume so because all existing cases were manually copy-and-pasted from 3d party web sources ... meaning we likely have all the possible weird HTML structures present somewhere in our cases. Defer to @cgruppioni on this, but it's probably worth it to comb through the DB to look for (1) cases with white-spaces only tags that (2) have been annotated. 

It's certain that removing a `<p>` tag will have a cascading, shifting effect on all later annotations, so if there are any whitespace tags out that precede an annotation, we'll have some annotation slippage.

leppert:
@abziegler I ran a query against cases that are annotated and it looks like 295 will be affected by the whitespace change I mentioned above. Going to spend tomorrow writing a migration script for those annotations.

leppert:
@abziegler I've reverted my whitespace filtering improvement since, after looking through the DB, the whitespace problems I was seeing in capapi imports also appears to be a semi-common issue in existing entries. For example:

https://h2o-dev-test.lil.tools/cases/291
image

I'll bring this whitespace improvement back in a separate branch when I've had time to write a proper migration; no need to delay this PR any further IMO.

There is, however, one case with annotations that my query surfaced that will be affected by this PR as it stands:

https://h2o-dev-test.lil.tools/cases/7958

The HTML for this case contains a `<section>` element that my branch properly unwraps. I may just migrate the annotations for this one case by hand in the name of moving this PR forward.

leppert:
@abziegler I refined my query and it looks like none of the cases I have in my dump, including case `7958`, will be an issue. An error in my query was pulling all cases that were part of a casebook instead of those that also have an annotation of some sort. We should check against a production dump, though, before deploying.

leppert:
I've moved the paragraph numbering improvements to this branch:
https://github.com/leppert/h2o/tree/373-overlapping-paragraph-numbers

leppert:
When deploying this branch, the following should be done:

- Add `CAPAPI_KEY` env var containing a valid capapi key
- Reindex the solr indexes for Cases (can be done using `Case.reindex` from a Rails console)
- Run the `rake h2o:correlate_cases` task to pull in capapi_ids for existing cases

These should be done on both staging and production

PR #502, opened 2018-08-13
Preview button label

abziegler:
A few more tweaks to this button labeling issue, to ensure that when previewing any draft casebook/section/resource, the button label says "Return to Draft" rather than "Revise."

"Revise" button applies only when something is published and the author wants to revise it, and that button triggers build of a draft.


bjohnson747:
This looked good to me on staging (I construed this and #500 as the same thing i was testing)

Issue #511, opened 2018-08-20
Remove carriage returns from case content text 


cgruppioni:
### **Remove carriage returns from case content text** 

1. Database has `\r\n`, Javascript only sees `\n\` 
`\r\n` length is 2 characters, `\n\` is 1

- [x] Update case text in database to have `\n` tags instead of `\r\n` https://github.com/harvard-lil/h2o/issues/511#issuecomment-415152310
- [x] Make sure that cap cases use `\n`

2. Inside a block quote, or inside of a node that doesn’t have ( guess) the resource span tag as the top level parent, it doesn’t count the text between the p tags 
which is a text node with`\r\n\`

<hr>  
Testing: 

https://h2o-dev-test.lil.tools/casebooks/81800-zittrain-torts-playlist-spring-2018/resources/7.1.2-sarah-wood-linda-wood-and-warren-wood-appellants-cross-appellees-v-derry-groh-and-choon-groh-appellees-cross-appellants/annotate?annotation-id=450608

-in a regular paragraph with italicized text, this issue seems to be corrected. but:

-paragraph 23 (quote text): when I make an elision on this text, it's moved back one character.

-paragraph 77 (h2-tagged text): first elision on this is fine:
<img width="414" alt="screen shot 2018-08-17 at 4 25 07 pm" src="https://user-images.githubusercontent.com/12850389/44287268-2ce7ed00-a23a-11e8-9bc7-43b7c9f0862d.png">
but when i make a second elision on the same line it's moved over one character:
<img width="392" alt="screen shot 2018-08-17 at 4 25 15 pm" src="https://user-images.githubusercontent.com/12850389/44287271-31140a80-a23a-11e8-97a9-4932ca94c017.png">

cgruppioni:
```
Case.all.each do |kase|
  if Content::Resource.where(resource_id: kase.id).empty?
    kase.destroy!
  else
    content = kase.content
    content.gsub(/\r\n/, "\n")
    kase.update!(content: content)
  end
end
```

cgruppioni:
Test Plan

- [ ]  quote text elides correctly
- [ ] h2 text elides correctly
- [ ] elisions look correct in draft, preview and published mode 
- [ ] Previously made annotations still look normal ( compare opencasebook.org to staging )

PR #516, opened 2018-08-22
If the parentElement is a blockquote, count the newlines between para…

cgruppioni:
…graphs in the nodeOffset
 
#511


bjohnson747:
By and large this is looking A+  - the only issue i found is below:

in H2 tagged text, when an annotation is already applied to H2 tagged text (ex., the text is highlighted), and an elision-annotation is applied to it, it shows up one character over from than where it ought to be. 
In this screen-capture, i 
(1) make two elisions to h2-tagged text without a problem, then remove them, 
(2) highlight the h2 tagged text, then 
(3) make an elision to the highlighted text, which is shows up off a character. 
-> https://drive.google.com/file/d/1fk8Z4NGTC98bnuMTM31ZdnIYzcFX0hzr/view?usp=sharing
I could not re-create this in blockquotes.

PR #521, opened 2018-08-23
Render list of casebooks that use a case in rails admin 

cgruppioni:


bjohnson747:
@cgruppioni let me know when this is ready to test! (right now i can't go to the Edit button on cases, as you know, where'd i'd need to go to test this).

cgruppioni:
@bjohnson747 I think you already have tested this, but if not you can now

bjohnson747:
@cgruppioni this looks great , and accessible through the case -> edit area.
it's also showing up at the list of cases at https://h2o-dev-test.lil.tools/admin/case, i don't think we need it at this screen too? IMO clicking into a case to view the list of casebook's its in is the only place this list needs to live.

bjohnson747:
screen shot 2018-08-24 at 11 54 22 am

cgruppioni:
@bjohnson747 Thanks! Yes that was accidentally left in there, I'm deleting it in the current branch I am working on

PR #522, opened 2018-08-24
Rename admin reference from case.resources -> case.resource_links

leppert:


config/initializers/rails_admin.rb
```diff
@@ -208,7 +208,7 @@ class ShowInApp < RailsAdmin::Config::Actions::Base
     end
 
     list do 
-      field :resources
+      field :resource_links
```
cgruppioni: Thank you ah

Issue #296, opened 2018-04-30
Annotations: add "show all" toggle to reveal full text of a resource


abziegler:
Faculty  want to CTRL-F across full text of case, including hidden/elided text, so need to be able to reveal all text

abziegler:
Also, this is important for export ... authors should not be forced to manually toggle/delete all the elisions to get to an unelided resource

PR #552, opened 2018-10-04
"show all elisions" button toggles elide & replacement. Choose to export without annotations

cgruppioni:
#296 

Add second set of xslt files for htmltoword. Not able to choose between them with the htmltoword gem, so need to manually set the path. 

Add back the export modal - with two buttons -> "with annotations" "without annotations"


rebeccacremona:
Hi! I saw the word "modal" go by in Slack, so thought I'd pop by this PR :-)

Want to use the `Modal` component instead of the generic `Component`component (https://github.com/harvard-lil/h2o/blob/master/app/assets/javascripts/lib/ui/modal.js) like the publish, clone, and new casebook modals do? (e.g. https://github.com/harvard-lil/h2o/blob/master/app/assets/javascripts/lib/ui/content/clone_modal.js) 

The Modal component comes with all the accessibility work done already. I know I'm supposed to be doing another round of a11y fixes some time, so I can switch this over then if you'd rather not. But, since you are in this code already, I thought it might be convenient for you to do it now. No pressure.

bensteinberg:
Merging and rebasing in order to generate artifacts at Travis...

bjohnson747:
It's working on elisions on a case, but I'm *not* getting the toggle to appear when making edits on a draft or published text block.

No toggle:
(from: https://h2o-dev-test.lil.tools/casebooks/66901-criminal-law-slap-chop/resources/1.3-test-text-tesxt-texst)
screen shot 2018-10-05 at 4 29 40 pm
screen shot 2018-10-05 at 4 26 50 pm

Export of all types (resource, section and casebook) exported showing (or not) annotations correctly.

Issue #462, opened 2018-07-20
Add ability to 'edit' replacement text annotations


abziegler:
Authors should be able to edit any replacement text annotations. Currently, we permit authors only to remove replacement text or to reveal the original text.

![h20-edit-replacement-text](https://user-images.githubusercontent.com/2140548/43006362-835c05ce-8c03-11e8-94af-1abdf08d82ba.png)

Issue #461, opened 2018-07-20
Clicking 'enter' on a "replacement text" annotation should save annotation


abziegler:
When an author is doing a "replacement text" annotation and clicks `Enter,` we treat it as a normal carriage return that causes a line break. 

![h2o-enter-on-replace-text](https://user-images.githubusercontent.com/2140548/43005717-b38c2ece-8c01-11e8-81e2-2e75d65964b4.png)

Clicking `Enter` should save the annotation rather than create a line break

PR #557, opened 2018-10-09
Save replace annotation on enter & annotation loses focus.

cgruppioni:
#461 & #462

### DIFF
--- a/app/assets/javascripts/lib/ui/content/annotations/replace.js
+++ b/app/assets/javascripts/lib/ui/content/annotations/replace.js
@@ -20,7 +20,8 @@ delegate(document, '.annotate.replacement', 'keypress', e => {
 // Pressing enter or spacebar in the contenteditable region shouldn't press the containing button
 delegate(document, '.annotate.replacement .text', 'keypress', e => {
   if (e.key=='Enter'||e.key==' '||e.keyCode==13||e.keyCode==32){
-    e.stopPropagation();
+    document.querySelector('[data-annotate-action=save-changes]').click()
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/lib/ui/content/annotations/replace.js
```diff
@@ -20,7 +20,8 @@ delegate(document, '.annotate.replacement', 'keypress', e => {
 // Pressing enter or spacebar in the contenteditable region shouldn't press the containing button
 delegate(document, '.annotate.replacement .text', 'keypress', e => {
   if (e.key=='Enter'||e.key==' '||e.keyCode==13||e.keyCode==32){
-    e.stopPropagation();
+    document.querySelector('[data-annotate-action=save-changes]').click()
+    e.target.blur();
```
rebeccacremona: Hi! I saw the word "focus" go by on Slack, so I thought I'd pop by this PR. 

Calling `blur()` drops keyboard focus, which means that anybody navigating via a keyboard or assistive technology will have to start again from the top of the page after this code runs: not a great user experience, exactly like when making an annotation caused the screen to scroll all the way to the top. 

Instead of calling `blur()` to unset the keyboard focus entirely, you can call `focus()` on the element you'd like to set focus to.

rebeccacremona: Also noticed: this code is listening for spacebar too (`e.key==' '`, `e.keyCode==32`); I'm guessing you still want people to be able to press spacebar without triggering that new click event?

PR #566, opened 2018-10-23
Fix i18n reference in export modal

leppert:


leppert:
@cgruppioni This bug was the result of my rebase, but it reminds me to mention that `js.erb` files that go through webpacker (vs sprockets) are interpreted in the top level scope of the Rails environment rather than within the view scope. The practical effect of that is that you have to include many of the helpers that you normally get for free (ex: https://github.com/harvard-lil/h2o/blob/master/app/assets/javascripts/lib/ui/content/delete_modal.js.erb#L1-L2) and when using I18n you need to reference it off of that object (`I18n.t`) rather than using the naked `t` helper.

PR #576, opened 2018-10-25
Use vuejs for annotation handles

leppert:
This does the basic job of getting Vue on the page and rendering a component, the AnnotationHandle component (the little pencil icons for deleting annotations etc). There are a large number of files that have been touched in order to better deal with paths in (S)CSS files, but the core of the branch is relatively small.

In terms of what this accomplishes on the Vue front, it's still rather anemic but it's a good beachhead for further development. The backend now outputs `<annotation-handle/>` tags which Vue picks up and renders out. The state in the app uses Vuex, which is the standard state management library for Vue. The Vue app is bootstrapped onto a `<div id="app">` element that wraps the whole page, meaning we can gradually add additional Vue component tags peppered throughout without needing to render everything else within that `div` through Vue. A prerequisite for this, though, is that you can't have `<script>` tags within that `div` and you'll see I've moved some around in this branch to avoid that.

On the road to a full featured Vue app, the caveats in this branch are that the state is bootstrapped from ruby through a `<script>` tag and the AnnotationHandle components map their `annotation-id` property to the appropriate item in the Vuex data store when initially rendered. In an ideal world, the state would be fed by an API and the handles would be instantiated and placed into the page based on a for-loop that cycles through the annotation data that's returned. That'll be a step for further down the road.

@cgruppioni, this could use your eagle eye, especially where I've changed the ruby annotation offset parsing we talked about yesterday to rely on an explicit property rather than implicit css classes. I did my best to test that, I think it's working equivalently, but I wasn't sure what good test cases would be so I'm hesitate to call it 1-for-1.

@rebeccacremona Per our conversation, I changed these handles to use buttons and added an aria property. It'd be helpful to give it a look and make sure I have that right.


app/controllers/content/resources_controller.rb
```diff
@@ -11,6 +11,7 @@ def show
 
   def annotate
     @editable = true
+    @include_vuejs = true
```
cgruppioni: - [ ] 1.

leppert:
@cgruppioni The save/cancel options should now be present when editing replacement text. I also removed functions for "converting" annotations because I didn't see anything in the UI that utilized it, though I also didn't entirely know what it was used for to begin with. If you happen to know, I'd love to understand it better and restore the code if needed.

PR #587, opened 2018-11-08
Check all ancestor elements for exclusion flag

leppert:


cgruppioni:
@leppert could you explain why this solves it? :)

leppert:
@cgruppioni So the offsets are calculated by using a tree walker to grab all text nodes within an element and then count them, excluding any that are part of annotation handles, buttons, etc. The old code did this by checking to see if the parent of the text node was of a certain CSS class (such as “annotation-button”) and then skipping the count on that one if so. I changed it to use an explicit data attribute instead so that we didn’t need to keep a running list of classes. However, I failed to account for the fact that the new Vue annotation handles have text nodes that are more deeply nested than simply the first parent, so that data attribute wasn’t being found. Now the code looks to see if any ancestors has the attribute and, if so, skips as expected.

cgruppioni:
@leppert Awesome! Thank you

PR #592, opened 2018-11-16
change links in footer

abziegler:
just hardcoded links to the new guide and blog

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -5,10 +5,11 @@
   <div class="layout">
     <div class="link-group">
       <div class="link">
+
         <%= link_to t('footer.links.home'), root_path %>
       </div>
       <div class="link">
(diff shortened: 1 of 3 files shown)


app/views/layouts/_footer.html.erb
```diff
@@ -17,7 +20,7 @@
         <%= link_to t('footer.links.about'), page_path('about')%>
       </div>
       <div class="link">
-        <%= link_to t('footer.links.user-guide'), page_path('docs/guide')%>
+        <%= link_to t('footer.links.help-guide'), 'https://about.opencasebook.org'%>
```
cgruppioni: Have you tried something like `root_path(subdomain: 'about')`? https://stackoverflow.com/questions/42014992/rails-link-to-passing-a-subdomain

cgruppioni: If that doesn't work - feel free to merge

PR #593, opened 2018-11-16
Admin can delete collaborators and add new collaborators to a casebook

cgruppioni:
- Admin can search for users by email_address or name. 
- Admin can select a role and add that user as a collaborator
- Admin can delete collaborators from the casebook
- Only the owners of the casebook show as the authors in the user view


cgruppioni:
Closed - rebased in #594

Issue #604, opened 2018-11-28
Annotator issues in Firefox


bjohnson747:
(Testing on: https://opencasebook.org/casebooks/73025-constitutional-law-demo-casebook-2/resources/1.5-childs-et-al-v-catlin-et-al/annotate)

1) Already having an annotation in a paragraph causes additional annotations to be placed incorrectly: http://recordit.co/ySQpI1BLEq
2) Can elide an individual paragraph, but attempting to elide more than one causes incorrectly placed annotation: http://recordit.co/s2TTqcAZkK
3) Annotation sidebar should follow the selected text - or jump down - as it's being clicked+dragged and the browser window scrolls; in Firefox it does not: http://recordit.co/5OtcsVvXS8

Issues not occurring in Chrome or Safari.

leppert:
A little more info on No.2: the exception is thrown when bridging paragraphs because range.endContainer returns the outer div.case-text rather than the paragraph’s text node.

leppert:
PR #607 fixes items 1 and 3. Item 2 has been spun out into its own ticket here https://github.com/harvard-lil/h2o/issues/609

PR #607, opened 2018-12-03
Disable prefixing for user-select to sidestep Firefox bug

leppert:
Before switching to webpack, this property was not vendor prefixed, leading to
it having an effect in Chrome and Safari but not Firefox. After the
switch, the vendor prefix meant Firefox was correctly disabling select
on paragraph numbers but this lead selections that spanned
paragraphs to register their end range as the enclosing `div.case-text`
rather than the actual paragraph in which the selection ended. Not
sure why Firefox behaves this way -- maybe there's a proper work
around to that behavior -- but this commit restores the pre-webpack
behavior for better or worse (seems to be for better).

Side note: I couldn't get `/* autoprefixer: ignore next */` to
work as outlined in their readme, hence the `off` then `on` statements.

### DIFF
--- a/app/assets/stylesheets/casebook.scss
+++ b/app/assets/stylesheets/casebook.scss
@@ -542,7 +542,9 @@ form.edit_content_resource, form.edit_content_section, form.edit_content_caseboo
   .p-handle {
     @include size(0px, 0px);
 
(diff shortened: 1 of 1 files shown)


cgruppioni:
#604

Issue #532, opened 2018-08-29
Add User Create and Casebook Author Edit to Rails Admin


abziegler:
1. Admins should be able to make accounts in Rails Admin
2. Admins should be able to transfer ownership/authorship of a casebook from one user to another in the Rails Admin

cgruppioni:
- Roles should travel if anyone in that collaborator publishes then create draft
- cloning should have no previously attached collaborators

bjohnson747:
updates per our Roles discussion:

Only two roles: Owner, and Editor. Owner role remains as-is, Editor can:
-Revise casebook (aka create new draft of existing casebook)
-Publish casebook
-Annotate items in casebook
-Edit casebook contents (add/remove/edit resources/sections, edit headnote fields + titles, etc)

Admins need ability (via checkbox, for example) to select if Editors get attribution and are displayed as additional authors.

Casebook (+ associated drafts) should appear in both Owners and Editors' dashboards.

cgruppioni:
In staging and prod

```
Content::Collaborator.all.each do |collaborator|
  if collaborator.role == 'owner'
    collaborator.update(has_attribution: true)
  end
end
```

bensteinberg:
@cgruppioni, did you run that, or do you want me to run it?

cgruppioni:
@bensteinberg It's notes for the next iteration of this issue that hasn't been deployed yet. I'll run it in prod console when it is deployed though - thanks for checking!

PR #620, opened 2018-12-11
Admin can add and edit collaborator roles 

cgruppioni:
#532

- Admin can search for users by email_address or name.
- Admin can select a role and give attribution and add that user as a collaborator
- Admin can delete collaborators from the casebook
- Admin can edit a collaborators attribution status
- Only the attributed collaborators of a casebook show as authors in the user view. Owners show up before editors in a comma separated line. 
- All collaborators of a casebook can see the casebook on their dashboard and make edits to it, publish it.
- When a casebook is published and then revised, all collaborators see the casebook in their dashboard correctly. 

Developer note: 
Run after deploying and before testing: 
```
Content::Collaborator.all.each do |collaborator|
  if collaborator.role == 'owner'
    collaborator.update(has_attribution: true)
  end
end
```


bjohnson747:
Feedback:

Clarification re roles: the only real difference between the two roles is that an Owner is automatically given attribution (that can be unchecked by an admin), and an Editor isn’t automatically attributed when they’re added to a casebook.

-So, editors should be able to create draft (Owners can create drafts, Editors cannot currently)
As an editor, I also didn’t have access to a draft I was added to.

-If an owner/editor makes a draft, all collaborators should be given access to it. (This will also keep multiple owners from trying to create multiple drafts.) - currently as an owner, when a different owner made a draft, I didn’t have access to that draft. (Same issue when my role was Editor). 

-Anyone added to a draft should also be automatically added to the associated published version of the casebook, and vice-versa: If they’re added to a published version, they should be added to the draft version, if there is one.

SMALLER THINGS:

-There should alway be one user with attribution (perhaps whatever user is in the first slot of names on admin collaborator page, if none have their checkbox checked?)

-when I was viewing a draft that then got published by different user, then refreshed the page, I got a 500 error. Ideally in this situation, H2O will kick me to the Published version. (This might be a next-round-of-improvements tweak)

-when there were collaborators, but none with attribution, the comma after owner’s name was still visible.

-when multiple collaborators are listed as Owner, no comma appeared between them. Once I added one as an Editor, commas appeared between all.

bjohnson747:
Second round of feedback!

-When a collaborator is added to a draft casebook, they're not also added to the Published version - they should be added as a collaborator to both / both should appear in their dashboard.

-Similarly, when i was added as a collaborator to a published casebook that also had a draft, I was able to see both in my dashboard, but couldn't access the draft (got the "page is only available to casebook editors" message). 
Need ability to access any drafts that are created if I was added to the published casebook.

PR #639, opened 2019-01-22
Integrate vuejs for annotations

leppert:
@cgruppioni This should be ready for tire-kicking without really any caveats aside from the fact that a few of the interactions and styles are slightly different, intentionally, but if you find anything off-putting I'm happy to go back and match what's currently in master.


app/controllers/content/resources_controller.rb
```diff
@@ -6,6 +6,8 @@ class Content::ResourcesController < Content::NodeController
   skip_before_action :check_public, only: [:export]
 
   def show
+    @editable = false
```
leppert: Maybe could remove?

app/models/content/annotation.rb
```diff
@@ -4,6 +4,7 @@ class Content::Annotation < ApplicationRecord
   has_one :unpublished_revision
 
   validates_inclusion_of :kind, in: KINDS, message: "must be one of: #{KINDS.join ', '}"
+  validates :content, presence: true, if: Proc.new { |a| a.kind == "replace" }
```
leppert: Does this need to be true for both note and link?

leppert: Also, this should do validation on the front end to prevent this state.

app/webpacker/store/modules/annotations.js
```diff
@@ -1,47 +1,117 @@
...
+    ),
+
+  getBySectionIndex: state => index =>
+    state.all.filter(
+      obj => (obj.start_paragraph == index ||
+             obj.end_paragraph == index ||
+             (obj.start_paragraph < index && obj.end_paragraph > index))
+    ),
+
+  getSpanningOffsets: state => (index, start, end) =>
```
leppert: Add a comment here

app/webpacker/store/modules/annotations.js
```diff
@@ -1,47 +1,117 @@
...
+      obj =>
+        (obj.start_paragraph == index &&
+         obj.start_offset > start &&
+         obj.start_offset < end) ||
+        (obj.end_paragraph == index &&
+         obj.end_offset > start &&
+         obj.end_offset < end)
+    ),
+
+  getAtIndexAndOffset: state => (index, offset) =>
```
leppert: Add comment

PR #644, opened 2019-01-25
Add JS versions of html_helpers.rb

leppert:
Some of the Ruby functions aren't needed because of the way we're
selecting elements in the JS. For instance, we don't need to filter
out comment nodes because we're using el.children which only returns
element nodes.

### DIFF
--- a/app/webpacker/components/TheResourceBody.vue
+++ b/app/webpacker/components/TheResourceBody.vue
@@ -14,6 +14,11 @@
 </template>
 
 <script>
+import { unwrapUndesiredTags,
+         emptyUlToP,
+         wrapBareInlineTags,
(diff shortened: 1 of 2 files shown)


app/webpacker/components/TheResourceBody.vue
```diff
@@ -37,18 +42,22 @@ export default {
   computed: {
     sections() {
       const parser = new DOMParser();
-      let children = parser.parseFromString(this.resource.content, "text/html").body.children;
+      let doc = parser.parseFromString("<div>This is something<div></div><li></li><a><b></b></a></div>"+this.resource.content, "text/html");
```
cgruppioni: What does adding the `<div>` string do?

PR #666, opened 2019-02-12
Remove cruft

### LINKED ISSUE #660: Exporting resource note annotation text shows inline
Note annotations should show up like link annotations, with an asterick as a footnote indicator and the content bottom of the resource. 

screen shot 2019-02-11 at 2 16 47 pm
screen shot 2019-02-11 at 2 17 00 pm
screen shot 2019-02-11 at 2 17 07 pm

----------------------------------

Currently they show up as inline
screen shot 2019-02-11 at 2 04 39 pm
screen shot 2019-02-11 at 2 05 09 pm
https://h2o-dev-test.lil.tools/casebooks/101502-regulation-of-financial-institutions-fall-2015/resources/1.1.2-state-ex-rel-duffy-atty-gen-v-western-auto-supply-co/annotate#

-----------------------

Here is the PR where this feature was originally set #492 

This is what it used to look like for example: 
screen shot 2019-02-11 at 1 48 32 pm
screen shot 2019-02-11 at 1 48 34 pm

leppert:
Since this code is now only used for exports, we can remove much of the logic and markup associated with front-end rendering. This also fixes #660.

@cgruppioni I wanted to remove the `include_annotations` flag from `apply_annotation_to_paragraphs` (and a few other places; not everywhere) but couldn't totally figure out why there are two export code paths there to begin with. Why not just bypass applying annotations entirely rather than execute `apply_annotation_to_paragraphs` while passing a flag that actually doesn't apply annotations to paragraphs?

### DIFF
--- a/app/services/apply_annotation_to_paragraphs.rb
+++ b/app/services/apply_annotation_to_paragraphs.rb
@@ -23,7 +23,8 @@ class ApplyAnnotationToParagraphs
   def perform
     if (paragraph_index != start_paragraph && paragraph_index != end_paragraph) || (paragraph_index == end_paragraph && paragraph_index != start_paragraph && end_offset == paragraph_node.text.length)
       # wrap entire p node in annotation
-      paragraph_node.children = annotate_html(paragraph_node.inner_html)
+      # if it's an elision, exclude it entirely
+      paragraph_node.children = kind == 'elide' ? '' : annotate_html(paragraph_node.inner_html)
       return
     end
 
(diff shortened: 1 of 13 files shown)


app/services/apply_annotation_to_paragraphs.rb
```diff
@@ -24,15 +21,9 @@ def initialize(params)
...
 
   def perform
-    # Adds data-p-idx to paragraph HTML tag 
-    paragraph_node['data-p-idx'] = paragraph_index
-
     if (paragraph_index != start_paragraph && paragraph_index != end_paragraph) || (paragraph_index == end_paragraph && paragraph_index != start_paragraph && end_offset == paragraph_node.text.length)
       # wrap entire p node in annotation
-      paragraph_node.children = annotate_html(paragraph_node.inner_html, handle: false)
-      if kind.in? %w{elide replace}
-        paragraph_node['data-elided-annotation'] = id
```
cgruppioni: I see this still being used in ResourceSectionWrapper, TheResourceBody, with-annotations/cleanup and with-annotations/h2o/export -- should the references be deleted there too?

PR #671, opened 2019-02-14
Add style for asterisk footnotes in word export

leppert:


leppert:
@bjohnson747 This changes the footnote asterisks into superscripts for better readability.

Issue #618, opened 2018-12-10
unintentional casebook clones appearing


bjohnson747:
Casebook that myself and others have been working on has had clones being created, even though (best that we can tell) the clone button has not been being hit.

The original casebook being worked on: https://opencasebook.org/casebooks/96420-open-source-property/layout
the clone that appeared: https://opencasebook.org/casebooks/109233-ignore-error-unwanted-clone-open-source-property/layout
@cgruppioni

cgruppioni:
```

  def edit
    # editing a section takes you to a cloned casebook and
    # the original casebook stays published
    @casebook = @casebook.clone(owner: @casebook.owner, draft_mode: true)
    @section = @casebook.contents.find_by(copy_of_id: @section.id)
    @decorated_content = @content.decorate(context: {action_name: action_name, casebook: @casebook, section: @section, type: 'section'})
    redirect_to layout_section_path(@casebook, @section) 
  end
```

PR #672, opened 2019-02-19
650 deleting resource in draft breaks published

cgruppioni:


cgruppioni:
Can fix #618

cgruppioni:
Notes from Greg: 

I think this is headed in the right direction and would improve even more if it leaned into a functional style. I’d change `ActionButtonBuilder` to simply be a namespace for the action functions named something like `ActionButton`, and pass the needed `casebook`, `resource`, and `section` as params. (edited) 
So I’d delete all of this:

  ```attr_reader :action, :casebook, :section, :resource
    
      def initialize(casebook, section, resource, action)
        @action = action
        @casebook = casebook
        @section = section
        @resource = resource
      end
    
      def perform(actions)
        # flatten the array b/c sometimes are pushing one array onto another 
        actions = actions.flatten
        # Turns the action symbols into methods and collect all the results
        transcribed_actions = actions.map { |action| self.method(action).call}
      end

    private```
(edited)
And then the function definitions become something like this:

  ```def create_resource_draft(casebook, resource)
        { link_to: true, text: I18n.t("content.actions.revise"), path: create_draft_resource_path(casebook, resource), class: "action edit one-line create-draft" }
      end

  def preview_section(casebook, section)
    { link_to: true, text: I18n.t("content.actions.preview"), path: section_path(casebook, section), class: "action one-line preview" }
  end```
(edited)

leppert [10:03 AM]
And then in the node_decorator you’d have this:

  ```def action_buttons
    case self
    when Content::Casebook
      casebook_actions.flatten.map {|a| ActionButton.send(a, casebook) }
    when Content::Section
      section_actions.flatten.map {|a| ActionButton.send(a, casebook, section) }
    else
      resource_actions.flatten.map {|a| ActionButton.send(a, casebook, resource) }
    end
  end```
You could refine this further by doing your flattening within the `*_actions` methods, assuming that’s a reasonable place to do it. (edited) 
BTW, not sure if this is a bug but I couldn’t find where `draft` is defined: https://github.com/harvard-lil/h2o/blob/48694f9211d6577405b99e0a0e6c39d32d53f994/app/services/action_button_builder.rb#L28

PR #718, opened 2019-03-20
Latest Bundler (~2.x); upgrade guard-bundler

rebeccacremona:
This seems fine. No idea if it's desirable 😉. Feel free to close if unwanted.


cgruppioni:
Will be implemented in the future

Issue #673, opened 2019-02-20
Next Steps (1): GUID based annotation insertion


cgruppioni:
Originally brought up in #652 

Annotations are rendered by Vue in reaction to annotation objects being [appended](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/store/modules/annotations.js#L115-L116) to the Vuex annotations store. For the initial page render, [TheResourceBody](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/components/TheResourceBody.vue#L80) calls the annotation API and gets a list of the existing annotations, including their `id`s, and appends them to this list. These `id`s are then used to instantiate Vuex [annotations_ui](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/store/modules/annotations_ui.js) objects that hold UI state for the annotations on the page, mainly whether the annotations should be expanded or not. This creates a clean separation between the data that is saved to the DB as an "annotation" and the data that's only needed for the state of the UI. Various components on the page will use the annotation ID to tie these two data representations together.

However, **this means that `id`s, which are managed by the DB and only exist after creation of the annotation in the DB, become critical for inserting annotations into the page.** This makes it difficult to, for instance, allow optimistic insertions of annotations on the page to be displayed while the API `create` action is called in the background. We cheat this a little bit when it comes to `ReplacementAnnotation`s, which are created inline in the page, by [inserting an annotation](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/components/TheAnnotator.vue#L150-L157) with `id=null`, where `null` is its own sort of unique `id`, and then updating the UI state with a proper id [when it comes back from the API](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/store/modules/annotations.js#L101-L107). This works, but only when there's a single unsaved annotation on the page.

[...]

PR #719, opened 2019-03-22
673 - GUID based annotation 

cgruppioni:
#673 

- Save temporary ids in store as negative numbers
- Notes and Links are saved optimistically in $store and destroyed when the box loses focus.

### DIFF
--- a/app/webpacker/components/ContextMenu.vue
+++ b/app/webpacker/components/ContextMenu.vue
@@ -73,6 +73,13 @@
         if(e.relatedTarget != this.$el &&
            !this.$el.contains(e.relatedTarget)){
             this.close(true);
+
+            // When clicking away from note or link annotation during creation, destroy temporarily stored annotation in store 
+            let annotation = this.$store.getters['annotations/getById'](this.$tempId);
+            let uiState = this.$store.getters['annotations_ui/getById'](annotation.id);
+
+            this.$store.commit('annotations/destroy', this.annotation);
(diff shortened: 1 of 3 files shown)


app/webpacker/components/TheAnnotator.vue
```diff
@@ -27,7 +27,6 @@
   <ContextMenu ref="linkMenu"
                :closeOnClick="false">
     <form @submit.prevent="submit('link', content)"
-          @keyup.esc="close"
```
cgruppioni: This wasn't actually being used. the scrollEventListener in ContextMenu automatically looks for focus changes and calls the close method

app/webpacker/components/ContextMenu.vue
```diff
@@ -69,27 +69,36 @@
...
-                    this.$emit('close');
-                }
-            },
+      onFocusout (e) {
+        if(e.relatedTarget != this.$el &&
+           !this.$el.contains(e.relatedTarget)){
+            this.close(true);
+
+            // When clicking away from note or link annotation during creation, destroy temporarily stored annotation in store 
+            let annotation = this.$store.getters['annotations/getById'](this.$tempId);
```
leppert: @cgruppioni I would leave knowledge of annotations out of ContextMenu so it's usable as a general UI element and instead handle this sort of thing in either a `close` handler that's passed in or modify the ContextMenu component to accept an `onFocusout` handler if it needs to be handled on that event specifically.

bjohnson747:
staging testing per the above looked good to me

Issue #674, opened 2019-02-20
Next Steps (2): Shift annotation saving out of TheAnnotator


cgruppioni:
Originally brought up in #652 

Once GUIDs have been worked into the annotation system, this should allow us to factor [the logic for saving annotations](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/components/TheAnnotator.vue#L137-L148) to the API out of TheAnnotator and into the respective annotation types, [as it is with the ReplacementAnnotation](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/components/ReplacementAnnotation.vue#L81-L89). For annotations without `content`, like `HighlightAnnotation`s, the API logic would simply be triggered in the `created` hook from the `head` annotation template instance. **For annotations with content, like `LinkAnnotation`s, this would allow us to move the [link creation context menu](https://github.com/harvard-lil/h2o/blob/master/app/webpacker/components/TheAnnotator.vue#L27-L35) out of `TheAnnotator` and into `LinkAnnotation`, thereby having all of a link's creation and editing functions happen within that component.**

PR #727, opened 2019-03-29
674 Move the annotator logic into single components

cgruppioni:
#674 

**Testing plan**
There are no noticeable changes for users, but a lot of the functionality under the hood has changed. So compare annotating to production and make sure that everything is still fully functional

- [ ] can create, edit and delete all forms of annotations correctly
- [ ] All menus ( main annotation menu, note, link, replace ) look the same
- [ ] when making a note or link annotation, click away makes the annotation disappear. Including the underscored text that was to be annotated and the text input box

### DIFF
--- a/app/webpacker/components/AnnotationBase.vue
+++ b/app/webpacker/components/AnnotationBase.vue
@@ -54,6 +54,12 @@ export default {
     ...mapActions(['destroy'])
   },
   created() {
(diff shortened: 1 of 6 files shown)


bjohnson747:
One issue with notes:

-when in edit mode, can't click on either the note-annotated text (underlined text) or the paperclip to close the note:
Screen Shot 2019-04-01 at 2 57 48 PM

When in view/preview,the note can't be closed, and the text that should be underlined isn't showing:

Issue #558, opened 2018-10-09
Publishing draft casebook sometimes instead creates separate casebook


bjohnson747:
Started with this casebook: https://opencasebook.org/casebooks/65582-contracts-reading-group-for-export-only/

Created a draft: https://opencasebook.org/casebooks/90960-contracts-reading-group-for-export-only/layout

Elided item: https://opencasebook.org/casebooks/91086-contracts-reading-group-for-export-only/resources/2.1-dougherty-v-salt/annotate?annotation-id=493116

Then when i clicked Publish, it took me to a separate, newly created version of the casebook (that included the change I just made): https://opencasebook.org/casebooks/91086-contracts-reading-group-for-export-only

The original casebook and casebook draft (listed at the top ^ ) remain. I made no other changes to these items, for troubleshooting's sake.

--------

I tried to re-create this issue with a couple of other casebooks - this happened on staging originally with a different casebook - but have yet to do so successfully; i.e., just eliding a item in a draft was not enough to make this issue recur. I don't know if there's a something specific that makes this happen every time, or if there's something in the publish-casebook code that's being intermittently funky.

cgruppioni:
WIP: https://github.com/cgruppioni/h2o-1/tree/618-cloning-bug

Issue #443, opened 2018-07-10
Disable multiple clicks on 'create casebook' + adding case/resource


bjohnson747:
Currently when user clicks the create casebook button (from scratch) multiple times, it creates multiple casebooks. Same with when they're searching + adding a case: clicking the case name multiple times adds it multiple times to a casebook.

Button should be set to not being able to be clicked repeatedly. (Becky mentioned HTML function 'disabled')

cgruppioni:
Set `disabled` on button 

Perma stage button to create a permalink has all the features

rebeccacremona:
rcremona [2:45 PM]
Re: making sure people can only click a button once:
https://www.w3schools.com/jsref/prop_pushbutton_disabled.asp
https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state (edited)

If it isn’t a true button (e.g. a span, a link, etc.), let me know. (edited)

PR #730, opened 2019-04-07
Refactor of node decorator to make editing action buttons more understandable

cgruppioni:
- Add tests for all action button sets on pages 
- Solve #618, #558, #443 which were all caused by buttons not being disabled after the original click
- Also fixes the scenario if a user has deleted a resource in a draft that is in a published version of the casebook. Clicking "Return to Draft" redirects user to draft casebook (instead of causing a 500 error because it is trying to return to the draft version of that resource which no longer exists )  @bjohnson747 Is there a ticket attached to this issue that we can close out?


bjohnson747:
@cgruppioni Looking good! Most of these tasks seem to be completing faster than they used to be. Notes/issues:

1. I can't click out of the 'creating draft' box, but i can click out of the 'Publishing' box still - which returns my view to the normal view for a few moments before it takes me to the Published book.

This isn't a bug per se, but it would be better to make it not-click-outside-of-able, a la the "creating draft' box, because I was able to click out of it (while it was actively 'Publishing') then mess around briefly: for example, I clicked out of it, and clicked Add Section and then got a 500 error for trying to do both of those at once.

2. Created a new casebook, tested it and published it - then went to click "revise" at the resource screen, and i got a 500 error.
Tested on this resource - got the error when I clicked "Revise" on it: https://h2o-dev-test.lil.tools/casebooks/99448-testbook-2/resources/1-zurich-insurance-v-killer-music-inc

3. when i clicked 'Export' it exported successfully but wouldn't let me export again until i refreshed the screen. (happened @ casebook view/section view / resource view, and with all types of users)

bjohnson747:
These ^ look good!

PR #754, opened 2019-04-18
Fix export tests

rebeccacremona:
This PR gets the export tests up and running again.

The export tests were designed to find the "href" of the Export link, and download the file at that url. I'm not sure exactly when, but the behavior of that link was changed in a way that broke the tests. (I originally thought https://github.com/harvard-lil/h2o/commit/ac7c04aebe2e2521b3d95d7b222cff6d2782abce#diff-53c7fe43dc207eaaa85a86e9125cf263, part of the Vue integration work, but it must have been before that.... Wrong month!). Since the "href" of the link was now simply "#" instead of the url of the docx, the test downloaded a copy of the very page it was on!

But, instead of fixing the tests, the files in the `test/files` directory were swapped out, replacing all the docx files with html files with a `.docx` extension. Example https://github.com/harvard-lil/h2o/blob/master/test/files/osx/test_export_casebook.docx, originating in the commit https://github.com/harvard-lil/h2o/commit/f2ca183bb3707cc817df6db2be07667ae8e5ce0c#diff-73b96e33aa453ef9eb39042032c6cd00). So, now the tests were actually just comparing two versions of the rendered page. Later commits continued updating those html files, as the markup of the rendered page changed (new css files, etc.): https://github.com/harvard-lil/h2o/commit/d4d24bb85cf83530ad22327a89ffb3303e305add#diff-73b96e33aa453ef9eb39042032c6cd00

This PR reverts all the commits related to the html "docx" files, and reworks the test to inspect the downloaded export files

Since these tests were commented out last October (https://github.com/harvard-lil/h2o/commit/3430ba2810e6fda85cc525a455511d4d4729fc8d), two other changes were made that affected them: a switch to Chrome, and a new test fixture. 
- A small problem with the test fixture (duplicate name) caused `:verified_student`'s dashboard to 500 (casebook with no owner)
- Headless Chrome's download behavior is a little finicky.... there are several tickets about problems with `window.open`: I could not get it working despite much much trying LOL. So, I switched to `window.location.assign`, which so far as I can tell, doesn't change the user experience at all.
- In order to wrest the downloaded files from Chrome's clutches, we're using some fancy features, which are not supported by Chromium-Chromedriver, but only by recent versions of Google's Chromedriver. So! I updated that on Travis. If you are using Chromium-Chromedriver locally, your tests will error. (The Docker setup uses Chromedriver.)

Finally, I updated the test fixture docx files: it looks like there have been some changes to the export code since the old ones were produced. I make no claims that the files I am including in this PR are correct! They are just the docx currently being produced by the app.

To assist in seeing how the files change over time, I updated the test to unpack the zip and compare all the files within. It should print a pretty diff when it fails... presuming Travis can handle colors :-) Even though we're diffing some fairly inscrutable xml, I'm hopeful this will help us more easily analyze what has changed when the tests report that an export file has changed.

I've also removed the separate files for OSX and Travis.... they don't seem to be necessary.


test/system/export_test.rb
```diff
@@ -1,26 +1,66 @@
...
+      end
+      files = Dir[@download_dir + '/*']
+      assert_equal files.length, 1
+      files.first
+    end
+
+    if format == 'Word'
+      Zip::File.open(downloaded_path) do |downloaded_docx|
+      Zip::File.open(expected_file_path(file)) do |expected_docx|
+        downloaded_docx.each do |downloaded_file|
```
cgruppioni: Save you a LOC here  

```
if format == 'Word'
    Zip::File.open(expected_file_path(file)) do |expected_docx|
      Zip::File.open(downloaded_path).each do |downloaded_docx|
        expected_file = expected_docx.glob(Regexp.escape(downloaded_file.name)).first
        assert_equal expected_file.get_input_stream.read, downloaded_file.get_input_stream.read
      end
    end
  end
```

test/system/export_test.rb
```diff
@@ -1,26 +1,66 @@
...
+          expected_file = expected_docx.glob(Regexp.escape(downloaded_file.name)).first
```
cgruppioni: This is REALLY smart

test/system/export_test.rb
rebeccacremona: Ooh fancy. Will do.

rebeccacremona:
Thanks for reading @bensteinberg and @cgruppioni. I'll merge after these tests pass. 

@cgruppioni would love if, after this is merged in, you would try running the full test suite on your local installation and let me know if everything passes there. (See the conversation tab for more details why)

rebeccacremona:
Oh, for what it's worth, I did find the commit that caused html to be downloaded instead of the docx: https://github.com/harvard-lil/h2o/commit/01129c04fe452e3a0fbf5755e554ba4161f57fe7#diff-53c7fe43dc207eaaa85a86e9125cf263

By switching 
```
import 'lib/ui/content/export_modal';
// import 'lib/ui/content/export';
```
where [`export` alters the `href`](https://github.com/harvard-lil/h2o/blob/01129c04fe452e3a0fbf5755e554ba4161f57fe7/app/assets/javascripts/lib/ui/content/export.js.erb) and [`export_modal` uses `window.open`](https://github.com/harvard-lil/h2o/blob/01129c04fe452e3a0fbf5755e554ba4161f57fe7/app/assets/javascripts/lib/ui/content/export_modal.js.erb).

PR #760, opened 2019-04-24
Correct contact email address; fix a couple typos.

rebeccacremona:


bensteinberg:
Want to change the email addresses in app/views/pages/privacy-policy.html.erb and app/views/pages/terms-of-service.html.erb?

rebeccacremona:
Whoops. Yep.

bensteinberg:
I also see an h2o.law URL in app/views/notifier/playlist_push_completed.erb, but I bet we're not using that anymore.

rebeccacremona:
I'm not going to change `app/views/notifier/playlist_push_completed.erb` because we should probably delete that file....

PR #768, opened 2019-04-25
Factor out need for el.innerText property

leppert:
This will allow for more consistent parsing between front and back
ends since backend solutions can't calculate element length based on CSS
rendering


leppert:
@rebeccacremona There's a performance case ([see here](https://vue-loader.vuejs.org/guide/scoped-css.html#also-keep-in-mind)) to made for using a css class here, so I've pushed an update with one. In terms of making the markup more easily inspectable, I'd advocate for using the [Vue Dev Tools](https://github.com/vuejs/vue-devtools), which has both Chrome and FF extensions.

leppert:
@rebeccacremona as for factoring out the need for a full browser, I believe this is it — fingers crossed. I'll need to write some more tests to confirm, but worth moving forward for now with the assumption that we're in a good spot.

rebeccacremona:
> @rebeccacremona There's a performance case ([see here](https://vue-loader.vuejs.org/guide/scoped-css.html#also-keep-in-mind)) to made for using a css class, so I've pushed an update with one. In terms of making the markup more easily inspectable, I'd advocate for using [Vue Dev Tools](https://github.com/vuejs/vue-devtools), which has both Chrome and FF extensions.

Vue Dev Tools! Awesome!!

rebeccacremona:
> @rebeccacremona as for factoring out the need for a full browser, I believe this is it — fingers crossed. I'll need to write some more tests to confirm, but worth moving forward for now with the assumption that we're in a good spot.

Wow. Unexpected and delightful. Fingers crossed too!

Issue #755, opened 2019-04-22
Export tests: both with and without annotations


rebeccacremona:
Best as I can tell, the export tests have only ever tested export without annotations. Let's test both options. Should close https://github.com/harvard-lil/h2o/issues/423

rebeccacremona:
@rebeccacremona, you goober, you only changed `window.open` to `window.assign.location` in half the necessary spots, in `downloadFile`, but not in `showExportModal` https://github.com/harvard-lil/h2o/commit/7601bc62e4a74041ca3b5276050bfc95c7baa285#diff-53c7fe43dc207eaaa85a86e9125cf263

(to investigate: why are these functions so different? Naively, I would have imagined `showExportModal` would itself call `downloadFile`.

Issue #680, opened 2019-02-25
Vue strips out whitespace from annotations that contain a single space


leppert:
There's a test for this, that's currently being skipped, here:
https://github.com/harvard-lil/h2o/blob/master/test/javascript/components/ResourceSection.test.js#L90-L101

This is an issue primarily when an annotation bridges elements, though it would also be an issue in the rare case that someone annotated a space character by itself. For example:

```html
<em>foo</em> <em>bar</em>
```

Creating a highlight across "oo ba" in the markup above creates results in the following pseudo code:

```html
<em>f<highlight>oo</highlight></em><highlight></highlight><em><highlight>ba</highlight>r</em>
```

Notice that the space within the second `<highlight>` is missing.

Changing the [whitespace compiler option](https://github.com/harvard-lil/h2o/blob/master/config/webpack/loaders/vue.js#L9) to its default of "preserve" would seem like the logical way to fix this, but this also removes the space. Converting the space to a `&nbsp;` while also setting `whitespace: 'preserve'` does render the space, though this will potentially introduce other whitespace problems in the templates that would need to be sorted through carefully; this might be our only option. The only other escape hatch with Vue and whitespace is with `<pre>` tags, the contents of which are left alone; perhaps there's a solution there.

Complicating the issue is that the `vue-jest` plugin doesn't let you set Vue template compilation options, so it's not currently possible to run tests with `whitespace: 'condense'` i.e. the test env doesn't match the prod output. Yet.

cgruppioni:
"It’s just a hack and it adds characters to the resource which always feels dicey. It works because the browser collapses the double space back down to a single space.

Or rather, that’s why it doesn’t destroy offset counts for subsequent annotations."

PR #769, opened 2019-04-29
Export tests with annotations

rebeccacremona:
This is a first pass at testing `.docx` export with annotations. Closes https://github.com/harvard-lil/h2o/issues/755

I left a fair bit of work undone:
- I created my own very basic annotation fixtures for the purpose. While developing, I commented out the already-present annotation fixtures and ran the full test suite: all tests passed anyway, with them absent. I did not investigate. But, this suggests that there's some tidying to be done
- I did not attempt to add any interesting annotations: the existing case text contained a single paragraph with text reading "content to elide" and similar, so I just followed instructions. Soon-ish, we should add better annotation fixtures, covering some of our more complex cases (multi-paragraph elisions, annotations that do fun things with whitespace ala https://github.com/harvard-lil/h2o/issues/680, etc.)
- ~We should probably tweak the test design so that the same casebook/section/resource can be exported with and without annotations.~ just did this.... because I introduced a bug, and this would have caught it......


test/system/export_test.rb
```diff
@@ -25,27 +25,41 @@ def teardown
...
 
-  def export_test path, format, file
+  def export_test path, format, file, include_annotations
     sign_in user = users(:verified_student)
     visit path
-    click_link 'Export'
 
+    if include_annotations
+      click_link 'Export'
+      click_button 'Without annotations'
```
cgruppioni: I'm shocked that you got clicking in a modal to work so easily. I feel like I've tried this in the past and had to try a bunch of things and was unsuccessful 🤷‍♀️

app/assets/javascripts/lib/ui/content/export_modal.js.erb
```diff
@@ -26,35 +27,25 @@ function export_casebook_path(casebookId, format = 'docx') {
...
-  let includeAnnotations = <%= Rails.application.config.export_annotations_by_default %>;
   if (pageInfo.resourceId)  {
-    window.location.assign(resource_export_path(pageInfo.resourceId) + (includeAnnotations ? '': '?annotations=false'));
+    window.location.assign(resource_export_path(pageInfo.resourceId) + (includeAnnotations ? '?annotations=true' : '?annotations=false'));
   } else if (pageInfo.sectionId)  {
-    window.location.assign(section_export_path(pageInfo.sectionId) + (includeAnnotations ? '' : '?annotations=false'));
+    window.location.assign(section_export_path(pageInfo.sectionId) + (includeAnnotations ? '?annotations=true' : '?annotations=false'));
   } else {
-    window.location.assign(export_casebook_path(pageInfo.casebookId) + (includeAnnotations ? '' : '?annotations=false'));
+    window.location.assign(export_casebook_path(pageInfo.casebookId) + (includeAnnotations ? '?annotations=true' : '?annotations=false'));
```
cgruppioni: Great point with the window.location.assign

test/system/export_test.rb
rebeccacremona: Wait a minute... if `include annotations` click `without annotations`????? Something is wrong here.

cgruppioni: Ahhhh yes

rebeccacremona:
Okay: reopening.

Now this tests all three possibilities: export of things without any annotations, export of annotated things using the "with" option and using the "without" option.

If I had done that the first time around, I would have noticed that I introduced a bug: `includeAnnotations` is a string, not a `Boolean` and would have noticed that same bug in the existing ruby-interpolated default value.

This should be all set now. 

If you are comfortable with this version, @casey, please merge at your convenience.

rebeccacremona:
(public shame on me for not writing all 6 tests the first time)

app/controllers/content/sections_controller.rb
```diff
@@ -109,8 +109,9 @@ def reorder
 
   private
 
-  def export_filename format
-    helpers.truncate(@section.title, length: 45, omission: '-', separator: ' ') + '.' + format
+  def export_filename format, annotations=false
```
cgruppioni: You can move this method and the section & casebook one into 1 method in the node model. 

Ex: 
`@section.export_filename('docx', @include_annotations)`

cgruppioni:
@rebeccacremona Just one small thing I commented on

Also, I usually like to make a test plan for whoever does it on staging. Just something simple like -- 

Testing Plan: 
- [ ] Take a resource, section and casebook that has annotations and export them with `with annotations` selected
- [ ] Take a resource, section and casebook that has annotations and export them with `no annotations` selected
- [ ] Take a resource, section and casebook that doesn't have annotations and export them
Make sure exports export successfully, inside content should not be changed by this ticket 

Do you think this covers everything that could be effects?

app/controllers/content/sections_controller.rb
rebeccacremona: Yeah, it seems goofy to have duplicate export logic in each of the three classes. Is it okay with you if I do not refactor that existing export logic as part of this PR? I just added a parameter to already-existing methods in the three classes. If you WOULD like that refactoring to happen at the same time, no problem, I'll close this and can add that later this week or early next week.

rebeccacremona:
@cgruppioni Yep, that sounds right, thank you for writing it out! I can definitely start including steps for manual testing in future PRs. (FWIW, I usually pop over to stage as soon as something gets merged to have a check personally.)

> Do you think this covers everything that could be effects?

app/controllers/content/sections_controller.rb
cgruppioni: That's fine!

cgruppioni:
Just tested, good to be merged in tomorrow

PR #777, opened 2019-05-08
Ensure action methods always return an array

leppert:


cgruppioni:
Why is this necessary?

leppert:
@cgruppioni it's often good to provide a consistent return type from a function. In this case, the lack of an array coming back was causing an error on line 18 when visiting a resource preview while logged out since `actions` was `nil` in that case: https://github.com/harvard-lil/h2o/blob/531939e98cbcf36a2f29c340fc4af8db1b158a22/app/decorators/content/node_decorator.rb#L18

Also, one of the `else` statements around line 51 appeared to be an error.

cgruppioni:
Good catch with #51. 

Hmm a logged out user shouldn't be able to see a preview of a resource

leppert:
> Hmm a logged out user shouldn't be able to see a preview of a resource

@cgruppioni possible the logic to prevent this would work if the `node_decorator` wasn't causing an exception to be thrown and an error page displayed.

cgruppioni:
@leppert That's a good point

PR #781, opened 2019-05-10
Pin version of Chrome and Chromedriver.

rebeccacremona:
I cannot imagine how I landed on version 2.41 of Chromedriver in my initial installation.... this pins Chrome to the version being distributed today via `http://dl.google.com/linux/chrome/deb/ stable main` and the matching version of Chromedriver. Works in Docker; syntax may be incorrect for Travis.

We're not positive this is precisely correct for the long run; I'll definitely readdress if this technique proves to be brittle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
 addons:
   apt:
     packages:
-      - google-chrome-stable
+      - google-chrome-stable=74.0.3729.131-1
       - wdiff
   postgresql: "9.4"
   artifacts:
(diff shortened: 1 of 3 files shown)


rebeccacremona:
(are there enough superlatives in that description? nah, needs a few more)

cgruppioni:
This is amazing

rebeccacremona:
LOL so terrible, right? I must have cut and pasted v. 2.41 from Stackoverflow or something.

cgruppioni:
@rebeccacremona Lol I feel like that came off as sarcasm but it really wasn't meant to be :) - Maybe this is great would've been a better comment

PR #782, opened 2019-05-13
Upgrade webpacker to proper release, 4.0.2

leppert:
This also removes items from the package.json file that should be left
out (despite "unmet dependency" warnings from yarn) so that their
versions can be derived from rails/webpacker


leppert:
@rebeccacremona they're dependencies beneath the webpacker JS component. See: https://github.com/rails/webpacker#upgrading

rebeccacremona:
@leppert Aha! Of course, thanks. 

Noted for my future self: if we update the webpacker gem, we must also update the `@rails/webpacker` node dependency.

Issue #814, opened 2019-06-07
Staging Testing Plan 6/10


cgruppioni:
Staging to Prod diff: https://github.com/harvard-lil/h2o/compare/v20190516-01...master

Testing plan: 
- [x] Clicking through search ( page 1, 2 of casebook search results ) does not lose filtering ( sorting by date)
- [x] Private cases can not be adde to casebooks 
- [ ] Previously if you put in the citation for a private case, that case gets added as a resource, rather than pulled from CAP
- [x] Sort casebook & case results by date and alphabetical order work 
- [x] When switching tabs ( casebooks, cases, users ) the count on each tab remains the same https://github.com/harvard-lil/h2o/issues/712
- [x] On user tab, the author casebook count should only be for published casebooks
- [x] The export modal no longer says “Confirm Publish” in the title, it says “Export Casebook”

--------------------------------------------------------------------------------------
Note for Casey: when deploying to Prod run these queries: 
https://github.com/harvard-lil/h2o/issues/804

bjohnson747:
@cgruppioni all looked good besides checkbox #3, some notes on that:

I tested the step as written, with a non-CAP case: I set it to private, searched for it by citation, and instead of adding the exsting non-CAP case it added the CAP case to the DB. <-looks good. However, i then tried setting an existing CAP case to private... see below:

-I set an existing CAP case in the DB to private (Walker v Keith, case #8435)
-Did Add Resource -> searched by name, didn’t pull the case up. (correct so far)
-Searched via Add Resource by case citation: it pulled it from CAP because the existing CAP case in the DB was set to private. Then added it to the casebook (and DB). The CAP case is now in the DB twice.

I think we shouldn't allow the case to be added again from CAP if the CAP case that already exists in the DB has been set to private. (I can’t think of any use cases where we’d set a CAP case to private.)

I think if a CAP case has been set to private, and a user tries to add it again by citation, it should refuse to add it, maybe they should get a message saying “there’s been an error trying to add this case, please contact H2O support at info@opencasebook.org for more information.”

In the immediate term, if you want to deploy this, one option could be: IF CAP case existing in DB is set to private, set it so users can't add the CAP case again to the DB (Add Resource just returns "Case not found.") - Then in the near future we can add the above error message, or something similar.

---

bjohnson747:
Separately, when I try to delete the extra versions of Walker v Keith, it says the following, even though it's not actually in these other casebooks:

[...]

PR #816, opened 2019-06-19
Rails admin allows you to delete a case when there are no casebooks that use it 

cgruppioni:
Deleted has_casebooks? method in cases, textblocks, links because it's redundant logic from associated_resources


cgruppioni:
#814

PR #837, opened 2019-07-29
Sanitize or escape user-supplied values

rebeccacremona:
Subsequent to the recent discussion about user-supplied ``, Jack was wondering how we handle other user-supplied values throughout the app. Investigation proved we could benefit from more careful handling of headnotes, and more careful handing of titles, usernames, and email addresses in a handful of spots. 

This PR runs headnotes through [Rail's sanitizer](https://github.com/rails/rails-html-sanitizer/blob/master/lib/rails/html/sanitizer.rb#L61) before rendering, and adds html-escaping in other places where user-supplied text is subsequently marked as `html_safe`. 

Experimentation suggests that Vuejs is stripping `script` tags from cases and textblocks before rendering. It would be good to get another set of eyes on this, especially from a dev more experienced with Vue to verify that we're okay. I'll plan to do more reading soon, to see if we can pin down its behavior.

Going forward, we should think about running headnotes, textblocks, and cases through `sanitize` on save. (We can't now, without shifting annotations, but once 812 lands....)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ addons:
 cache:
   bundler: true
   yarn: true
+jdk:
+  - openjdk8
 rvm:
   - 2.4.1
 before_install:
--- a/app/controllers/password_resets_controller.rb
+++ b/app/controllers/password_resets_controller.rb
@@ -6,25 +6,21 @@ class PasswordResetsController < ApplicationController
     @user = {email_address: params.fetch(:email_address, '')}
   end
 
-  def html_safe x
-    x.html_safe
-  end
-
   def create
     account_email = params.fetch(:user, {}).fetch(:email_address, nil)
     if account_email.blank?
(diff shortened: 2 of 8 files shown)


rebeccacremona:
(whoops looks like I have a test to fix, will do shortly)

leppert:
@rebeccacremona to answer you question above, in the Vue rendering I remove anything that's not a layout element here:

https://github.com/harvard-lil/h2o/blob/5cf6c867cf65e600e0c3be29a2c0ff2335e3ca21/app/webpacker/libs/resource_section_parsing.js#L167-L169

The whitelist of layout elements is here:

https://github.com/harvard-lil/h2o/blob/5cf6c867cf65e600e0c3be29a2c0ff2335e3ca21/app/webpacker/libs/html_helpers.js#L52

As for what defines a "layout" element, that was mostly just me going through this list and whitelisting what I thought was either appropriate for H2O's current use cases.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element

PR #844, opened 2019-08-06
Admin can add admin privileges to another user  

cgruppioni:
Run `Role.where.not(name: ["superadmin", "case_admin"]).destroy_all` for leftover roles from old h2o ( questions, askers )


rebeccacremona:
This is now implemented via the Django admin.

PR #846, opened 2019-08-07
Add nginx container

bensteinberg:


bensteinberg:
Rails is still at localhost:8000, Django at localhost:8001; nginx is at localhost:8002.

bensteinberg:
I'm not doing any special handling of assets here...

PR #852, opened 2019-08-09
Static pages content and urls

abziegler:
Used same static page pattern we used in Perma.

### DIFF
--- a/_python/main/urls.py
+++ b/_python/main/urls.py
@@ -2,6 +2,7 @@ from django.urls import path, re_path, register_converter
 from rest_framework.urlpatterns import format_suffix_patterns
 
 from . import views
+from .views import DirectTemplateView
 
 #
 # Converters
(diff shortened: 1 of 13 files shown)


rebeccacremona:
Awesome!! Quick tweak: looks like a find-replace about => about.html hit more files than was intended.

_python/main/urls.py
```diff
@@ -67,4 +68,9 @@ def to_url(self, value):
     path('users/<int:user_id>', views.dashboard, name='dashboard'),
     path('casebooks/<idslug:casebook_param>/sections/<ordslug:ordinals_param>/', views.section, name='section'),
     path('casebooks/<idslug:casebook_param>/', views.casebook, name='casebook'),
+    re_path(r'^about/?$', DirectTemplateView.as_view(template_name='pages/about.html'), name='about'),
+    re_path(r'^privacy-policy/?$', DirectTemplateView.as_view(template_name='pages/privacy-policy.html'), name='privacy-policy'),
+    re_path(r'^terms-of-service/?$', DirectTemplateView.as_view(template_name='pages/terms-of-service.html'), name='terms-of-service'),
+    re_path(r'^faq/?$', DirectTemplateView.as_view(template_name='pages/about.html'), name='faq'),
```
rebeccacremona: It doesn't really matter, but I don't think we need the more-advanced "re_path" here; regular path with "about/", "privacy-policy" etc. ought to do the job.

PR #873, opened 2019-10-01
Use our updated capapi gem.

rebeccacremona:
(see https://github.com/harvard-lil/capapi-ruby/pull/1)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ gem 'bcrypt'
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.1.0', require: false
 gem 'cancancan'
-gem 'capapi', git: 'https://github.com/leppert/capapi-ruby.git'
+gem 'capapi', git: 'https://github.com/harvard-lil/capapi-ruby.git'
 gem 'ckeditor'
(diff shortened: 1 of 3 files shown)


rebeccacremona:
LOL. This solves the
```
A NoMethodError occurred in searches#show:

  undefined method `[]' for nil:NilClass
  app/controllers/searches_controller.rb:14:in `show'
```
That was happening during search.... exposing the 
```
NoMethodError (undefined method `with_indifferent_access' for nil:NilClass):
  
app/controllers/content/cases_controller.rb:37:in `from_capapi'
```
That now happens during the actual case creation.

In the meantime, CAP has restored slashes to all its URLs, so prod is fine for the moment.
We should still try to find the other spot(s) that are looking for slashes. See rcremona/bsteinberg/jcushman Slack for most recent state of the hunt. Current tree we are barking up: `capapi_obj.court` is being converted into a `Capapi::CapapiObject` instead of `Capapi::CapapiCourt` and so is falling through this switch: https://github.com/harvard-lil/h2o/blob/master/lib/capapi/model_helpers.rb#L41

rebeccacremona:
We may or may not look into this further... might address during python port.

PR #874, opened 2019-10-02
First pass at whitelisting Django routes

bensteinberg:
This exposes the application locally at port 8003, manually whitelisting routes we want to send to Django in `docker/h2o.conf`. To try it, run, in three separate terminals,
```
docker-compose up
bash docker/run.sh
docker-compose exec python ./manage.py runrver 0.0.0.0:8000
```

### DIFF
--- a/_python/config/settings/settings_dev.py
+++ b/_python/config/settings/settings_dev.py
@@ -1,6 +1,6 @@
 from .settings_base import *  # noqa
 
-ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]', '.local', 'backend']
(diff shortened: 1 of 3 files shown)


bensteinberg:
@jcushman, do you think it's preferable to enumerate all the Rails routes as well, so we have to explicitly move or delete them?

jcushman:
Aw, this is super cool! No, I don't think we want to enumerate Rails routes, because we want to make sure we're not breaking anything that currently works. What will work similarly to that though is getting an ongoing report of which Rails routes are still being hit in the access log.

PR #881, opened 2019-10-05
Python: serve Rails static assets in Django using whitenoise

jcushman:
This avoids the temporary hack we had of loading static assets from production URLs. Django will now use `whitenoise` to serve static assets, serving the Django assets under `/static/` and also anything in the Rails public/ folder under `/`. So Rails files will now be available as `http://h2o-dev.local:8001/robots.txt`, `http://h2o-dev.local:8001/packs/css/main-1a2b3c4d.css` etc, and Django assets for the admin and django-debug-toolbar and anything else will also work.

This opens up the ability to gradually migrate any assets to `/static/` if they are only referenced from the Django side.

Rails assets have hashes in the filenames that might change and shouldn't be included in our templates. So this also includes a `rails_static` templatetag that looks up hashed filenames, inserting "/packs/css/main-a4a68e68.css" for `{% rails_static "packs/css/main.css" %}`.

### DIFF
--- a/_python/config/settings/settings.py.example
+++ b/_python/config/settings/settings.py.example
@@ -4,8 +4,6 @@
 from .settings_dev import *  # noqa
 
 
-STATIC_URL = '/static/'
-
 # log all SQL statements:
 # LOGGING['loggers']['django.db.backends'] = {
 #     'level': 'DEBUG',
--- a/_python/config/settings/settings_base.py
+++ b/_python/config/settings/settings_base.py
@@ -12,13 +12,15 @@ https://docs.djangoproject.com/en/2.2/ref/settings/
 
 import os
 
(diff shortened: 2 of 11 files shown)


bensteinberg:
@jcushman -- I have nginx config to let Django handle `/static`, but I can't let Django handle

> and also anything in the Rails public/ folder under `/`

without suppressing the ability of Rails to handle routes we're not whitelisting for Django. I added `/robots.txt` to the whitelist; I don't see anything else in the source for the front page that needs a whitelist entry, since links to assets in `packs/` start with `/static/`. Are there other things I should whitelist?

jcushman:
Having Django just handle `/static/` is great! The other stuff can be served as it is now on stage/prod -- serving `/` assets through whitenoise is more helpful for dev.

(Sidenote, links to assets in packs/ should not be starting with /static/ -- if they do, let's talk ...)

PR #893, opened 2019-10-10
Merge draft casebooks

jcushman:
- Add a casebook.merge_draft function, not yet exposed through views (though I do have some WIP locally on the view to go along with this). This uses the strategy Becky and I discussed where casebook attributes are copied over, old casebook assets are deleted, and draft content nodes are pointed to the published casebook. I'm sure I missed something!

- Coming along for the ride: `fix_after_rails("some todo message")` and `fix_before_deploy("some todo message")`, to use in place of our current `# TODO` comments. `fix_before_deploy` asserts that `settings.NOT_ON_PRODUCTION` has been set to true, so it will make a fuss if we whitelist a route we shouldn't. `fix_after_rails` is just a no-op, but making it a function ensures that we're consistently using the same string and can find them all later.

### DIFF
--- a/_python/config/settings/settings_base.py
+++ b/_python/config/settings/settings_base.py
@@ -196,3 +196,8 @@ WHITENOISE_USE_FINDERS = True
 
 CAPAPI_BASE_URL = 'https://api.case.law/v1/'
 CAPAPI_API_KEY = ''
+
+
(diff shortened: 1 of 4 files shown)


rebeccacremona:
Looks great! Per discussion, I'll add handling for `copy_of` tomorrow.

PR #901, opened 2019-10-23
Remove max bounding in search function to handle impossible offsets

leppert:
This is meant to address the following error that was being thrown in production:

```ruby
undefined method `[]' for nil:NilClass
​
  lib/differ.rb:69:in `get_delta_at_offset
```

I believe this was caused by annotations with "impossible" offsets (i.e. greater than the length of the text). This PR causes those offsets to be left unmodified when calculating new offsets on text updated.

### DIFF
--- a/lib/differ.rb
+++ b/lib/differ.rb
@@ -60,9 +60,7 @@ module Differ
 
     def get_delta_at_offset diffs, offset
       # Reverse the list so we find the last diff that includes the offset and for which the offset isn't at the start of the diff (i.e. before the changes in that diff would have any effect)
-      diff = diffs.reverse.find { |diff|
-        diff[2].max >= offset && diff[2].min < offset
-      }
+      diff = diffs.reverse.find { |diff| diff[2].min < offset }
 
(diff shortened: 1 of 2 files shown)


rebeccacremona:
@leppert I regret to report that I:
1) spun up `develop` after merging this in; 
2) loaded in a fresh production database, as of 10/7
3) Attempted to merge the draft casebook that was generating this error.....

and now, instead of 
```
undefined method `[]' for nil:NilClass
​
  lib/differ.rb:69:in `get_delta_at_offset
```
it fails with
```
<NoMethodError: undefined method `-' for nil:NilClass>
```
with a gruesomely formatted list-in-email stacktrace, summarized by
```
"/h2o/app/services/annotation_converter.rb:11:in `global_offsets_= to_node_offsets'", "/h2o/app/models/content/annotation.rb:6= 1:in `update_paragraph_based_offsets'"
```

I can investigate shortly; as background, was just wondering: were you able to successfully merge casebook `166815` locally, or were you looking primarily at the differ, rather than the end-to-end process?

PR #902, opened 2019-10-24
Action buttons; refactoring; partial "edit" pages.

rebeccacremona:
The goal of this PR was to port over the logic for which "action buttons" (e.g. Save, Publish, Clone, Export, etc.) to display on which pages.

This proved a very useful approach for starting to build out the rest of the app: action buttons appear on virtually every page. So, this PR includes a number of partially-implemented views, and, by way of the buttons, exposes much as-yet-unported functionality. For instance, you can click the "publish" button... but it doesn't publish yet! There are no form fields on the pages that should have forms. Etc. I would have preferred not to merge in these half-finished pages... But, I think that's probably impossible short of finishing the port in one go :-) I believe it is unambiguous what has been built and what hasn't. I've included lots of TODOs and topics for discussion. A PR to come soon will address some of the TODOs, and will convert the rest of them to use the better "todo" utility functions.

Working on action buttons also proved a useful project for exposing opportunities for better abstractions in the code base. THANKS @jcushman for the ContentNode db method that arranges for ContentNode querysets to contain Casebook, Section, and Resource instead of ContentNodes. That turns out to make a lot of things cleaner.... We're also starting to tease out the similarities and differences between Casebooks, Sections, and Resources, in anticipation of eventually re-imagining this data model, using less polymorphism and more consistent tree handling. So, there's a lot of refactoring in models.py. I'm trying to make it easy to see at a glance which methods are common to all node types, which methods are implemented by all node types but with differing logic, and which methods are unique to specific node types. We'll see how it holds up.

I've included tests for the action buttons specifically, and a few other tests here and there.... I plan to continue filling in tests, as I fill in more TODOs.

### DIFF
--- a/_python/conftest.py
+++ b/_python/conftest.py
@@ -245,6 +245,33 @@ def full_casebook(casebook_factory):
     ResourceFactory(casebook=casebook, ordinals=[1, 4, 3], resource_type='Default', resource_id=DefaultFactory(user=user).id)
     return casebook
 
+@pytest.fixture
+def full_private_casebook(full_casebook):
+    """
+    >>> private, published = [getfixture(f) for f in ['full_private_casebook', 'full_casebook']]
+    >>> assert published != private
+    >>> assert private.is_private
+    >>> assert all(node.is_private for node in private.contents.all())
+    """
+    casebook = full_casebook.clone()
+    return casebook
+
+@pytest.fixture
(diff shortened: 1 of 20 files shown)


rebeccacremona:
Why yes, I AM going to merge my own PR.

PR #903, opened 2019-10-25
Modify differ to handle negative offsets

leppert:
Annotations may have negative offsets as a result of the text beneath
them being deleted.


jcushman:
I think this is outmoded by differ.py -- let's close for now!

PR #905, opened 2019-10-27
Update scrypt for OS X Mojave compatibility

holtchesley:
The particular commit that this needs to pull in is https://github.com/pbhogan/scrypt/commit/8f94ec5ba7c3b59844d12fc4480430c2b7889dc7 , which targets x86_64 instead of i386 arch.


bensteinberg:
Thanks, @holtchesley! I'm closing this in favor of #915, in which I'm updating this and a couple of other gems; in addition, I'm bumping the image version number in `docker-compose.yml`, so as to force a rebuild of the container. We'd do the same thing for the python service if we were to update `requirements.txt`.

PR #906, opened 2019-10-30
Edit pages!

rebeccacremona:
This PR adds forms to the edit casebook, edit section, and edit resource pages, allowing authorized users to alter the titles, subtitles, and headnotes of casebooks/sections/resources. It also allows authorized users to update the URL of Link Resources in their casebooks. It includes the code for allowing authorized users to update the contents of TextBlock Resources, but disables it for now: we cannot safely update TextBlocks until we have migrated over the code that re-positions annotations when HTML changes.

(Note for testing in browser: we haven't built "publish" yet, so don't expect that to work :-))

Some details:

- I've added the django-crispy-forms library, which I have not used before, but which made the customization of Django model forms quite convenient, and took care of adding bootstrap's form classes without any fuss. I tentatively think it's great! If folks think adding a dependency is unnecessary, we can remove.

- I made a change that affects the ruby side as well: turns out we have been hiding the form labels using `display: none`. That actually hides them from screen readers too: these form fields effectively have had no labels. I tweaked the scss to include Bootstraps `.sr-only` class instead; I don't see a visual difference, and I now hear the labels for these fields.

- Totally unexpectedly, adding the `cdkeditor` class to a `<textarea>` JUST WORKS!!!!! I did not anticipate that AT ALL. We're definitely going to have to look into how that's being achieved. (Note: adding the class to an `<input>` does NOT work: the wysiwyg is loaded, but the pre-existing text is not loaded; the editor is always blank.)

- This PR makes it particularly obvious that there is much overlap between the casebook, section, and resources views.... I haven't put a lot of effort into figuring out what abstractions might clean things up. For now, it's here in all its repetitive glory.


rebeccacremona:
Oh! I forgot! This PR also re-introduces the HTML sanitization that I removed like an idiot in https://github.com/harvard-lil/h2o/commit/088e21478ee00aac052c58cc47a52a804e08817a#diff-bf70808cb1950d7a43d1618577271610

Greg switched the ruby application to sanitize when saving to the DB https://github.com/harvard-lil/h2o/commit/91f74da550096a7fbbb9df4c3bdf5bc6436c5de1. I just.... stopped using the sanitizer on the python side.

This PR replicates, hopefully closely, the cleansing applied on the ruby side, before save.

rebeccacremona:
What fun, a test that passes sometimes but not others. Debugging.

rebeccacremona:
This smells to me like... expecting a dict to be an ordered dict in 3.5. (30+ successful runs, followed by 15+ failures, then another 10 successes). When it fails, these so-called "sections" are actually resource objects. This is the only test experiencing failures.

rebeccacremona:
Jack figured it out: there is no default `ordering` on Casebook objects `contents`.

PR #907, opened 2019-10-31
Can move nodes in the content tree

jcushman:
- Add the `ContentNode.content_tree__move_to` method, as well as a bunch of supporting methods
- Refactor some ancestry methods to start with `version_tree__` (I don't think I hit all of the methods that could use this new naming scheme though)
- Add `-v -v` option for `assert_num_queries`

### DIFF
--- a/_python/main/models.py
+++ b/_python/main/models.py
@@ -1,3 +1,4 @@
+from os.path import commonprefix
 from urllib.parse import urlparse
 
 from django.contrib.auth.models import AnonymousUser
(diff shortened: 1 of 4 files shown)


_python/main/models.py
```diff
@@ -404,49 +406,289 @@ def type(self):
...
+        if len(new_ordinals) < 1:
+            raise ValueError("Cannot move node to root")
+        if type(self) is Casebook:
+            raise ValueError("Cannot move casebook node")
+        if new_ordinals[:len(self.ordinals)] == self.ordinals:
+            raise ValueError("Cannot move a node inside itself")
+
+        # find common grandparent node for old and new location
+        old_ordinals = self.ordinals
+        common_prefix = commonprefix((old_ordinals, new_ordinals))
```
rebeccacremona: commonprefix! great idea; didn't know about this method.

rebeccacremona:
This is so intense!! Awesome. I'm going to merge now and study later, because I think I need to read this with Monday brain... but I want to use it in MY branch now :-)

PR #909, opened 2019-11-04
Shift annotations when underlying text is edited

jcushman:
This ports over the code from differ.rb and adds code to update annotations when Case or TextBlock content is modified.

There are two new sources of error introduced during the port (in addition to the usual may-not-have-done-it-right source):

(1) lxml may extract text from html differently from the Ruby html parser
(2) I modified the internal data structure from differ.rb so it looks like:

```
offsets = [0, 5, 13, 23]  # the offset beginning each range in the first string
deltas = [RelativeDelta(0), AbsoluteDelta(5), RelativeDelta(-8), RelativeDelta(-2)]  # how to adjust offsets within each range 
```

This basically precalculates the delta to apply for every given offset. The `diffs` data structure in the current ruby instead looks like this:

```
[(0, 'Kids ', Range(min=0, max=5), Range(min=0, max=5)),
 (-1, 'happily ', Range(min=5, max=13), Range(min=5, max=13)),
 (0, 'eat their ', Range(min=13, max=23), Range(min=5, max=15)),
 (1, 'happy ', Range(min=23, max=29), Range(min=15, max=21)),
 (0, 'meals.', Range(min=23, max=29), Range(min=21, max=27))]
```

... which I find a little less intuitive as far as knowing how to adjust a given offset.

So, given the various things that can break here, we'll definitely want to do some checking of this one on real data! To make things a little easier, I added some test helpers that take readable text input and output, so it's easy to collect edge case examples.

---

A major question I'm not clear on, but you might be, is what to do with annotations for text that is deleted. The current ruby code sets the start and stop offsets to -1, so this does too. I think that implies deleting them. So:

(a) how and when should annotations with -1 offsets be filtered out or removed?
(b) is that actually correct? like it makes sense to remove a "highlight" or "delete" annotation, but what about a note?

---

Also in this PR:

- revert a find-replace mistake from the version_tree PR
- Add an `EditTrackedModel` model that represents my newest swing at tracking edited fields on Django models

### DIFF
--- a/_python/conftest.py
+++ b/_python/conftest.py
@@ -14,15 +14,18 @@ from django.db.backends import utils as django_db_utils
 
 from main.models import ContentNode, User, Casebook, Section, Resource, ContentCollaborator, Role, Default, TextBlock, \
     Case, CaseCourt, ContentAnnotation
+from main.utils import re_split_offsets
+
+from test_helpers import dump_casebook_outline
 
 
 # This file defines test fixtures available to all tests.
 # To see available fixtures run pytest --fixtures
 
-
 ### internal helpers ###
 
 # functions used within this file to set up fixtures
+
 
 def register_factory(cls):
     """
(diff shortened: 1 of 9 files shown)


leppert:
Sweet! Not sure if it's in here, but I'd advocate porting over the tests as well unless the test helpers you mentioned are more comprehensive. https://github.com/harvard-lil/h2o/blob/develop/test/lib/differ_test.rb

jcushman:
Yep, differ_test.rb is covered as far as I know! I started by porting over differ_test.rb into docstring tests, and then refactored some stuff to be more compact using test helpers. In particular, this part tests the core adjust_offset logic:

```
>>> assert_offsets_adjusted("fo*o b*uzz", "fo*o bar b*uzz")  # adding word
>>> assert_offsets_adjusted("fo*o b*ar b*uzz", "fo*o *b*uzz")  # deleting word
>>> assert_offsets_adjusted("fo*o l*orem b*uzz", "fo*o *ipsum b*uzz")  # editing word
...
```

Each asterisk here represents a separate `adjust_offset` test case -- like the first line is the equivalent of:

```
assert AnnotationUpdater("foo buzz", "foo bar buzz").adjust_offset(2) == 2
assert AnnotationUpdater("foo buzz", "foo bar buzz").adjust_offset(5) == 9
```

Likewise, the `adjust_range` tests from differ_test.rb end up being tested by the helpers `annotations_factory` and `dump_annotated_text`, which read and write text strings like this:

`'<p>\n  <em>[note]Keep foo[/note] [highlight]delete bar[/highlight] [elide]keep baz[/elide] buzz</em>\n</p>'`

`annotations_factory` turns that string into a case with note, highlight, and elide annotations; we can then change the html, save, and pass to `dump_annotated_text` to confirm the expected output.

PR #910, opened 2019-11-04
Publishing casebooks

rebeccacremona:
This PR adds the ability to publish a private casebook, both a) a newly-composed casebook that has never been published, and b) a draft of an already-published casebook, whose publication == "merging" the draft into the parent.

It's pretty simple, but includes some extras:
- some JS that pulls Django's CSRF token out of the page for use in AJAX requests
- some middleware that re-creates Rails' strategy of listening for a "X-CSRF-TOKEN" in requests, and if present, using it's value ("PATCH", "PUT", "DELETE", etc.) in place of the truly received HTTP verb (POST) when routing requests.
- a temporary duplication of the `casebooks` route in `urls.py` to work around a regrettable situation with trailing slashes.....
- a tweaking of the Webpack dev server, to smooth things while we need functional Rails and Django setups.

### DIFF
--- a/_python/main/views.py
+++ b/_python/main/views.py
@@ -329,8 +329,10 @@ class CasebookView(View):
         casebook = get_object_or_404(Casebook, id=casebook_param['id'])
 
         # check permissions
-        if not casebook.viewable_by(request.user):
+        if not casebook.editable_by(request.user):
             raise PermissionDenied
+        if casebook.is_public:
+            raise PermissionDenied("Only private casebooks may be published.")
 
         if casebook.draft_mode_of_published_casebook:
             casebook = casebook.merge_draft()
--- a/_python/main/views.py
+++ b/_python/main/views.py
@@ -256,17 +256,10 @@ def dashboard(request, user_id):
     return render(request, 'dashboard.html', {'user': user})
 
 
+@require_http_methods(["GET", "PATCH"])
 def casebook(request, casebook_param):
     """
         Show a casebook's front page.
-
-        TODO: slashes.
-        > RuntimeError: You called this URL via POST, but the URL
-        > doesn't end in a slash and you have APPEND_SLASH set. Django
-        > can't redirect to the slash URL while maintaining POST data.
-        > Change your form to point to localhost:8001/casebooks/157662/
-        > (note the trailing slash), or set APPEND_SLASH=False in your
-        > Django settings.
 
         Given:
         >>> casebook, casebook_factory, client, admin_user, user_factory = [getfixture(f) for f in ['casebook', 'casebook_factory', 'client', 'admin_user', 'user_factory']]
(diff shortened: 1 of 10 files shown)


_python/main/views.py
```diff
@@ -256,75 +260,134 @@ def dashboard(request, user_id):
...
+            ...     content_includes=private_casebook.title,
+            ...     content_excludes="You are viewing a preview"
+            ... )
+            >>> private_casebook.refresh_from_db()
+            >>> assert private_casebook.is_public
+
+            Drafts of already-published casebooks, when published, replace their parent.
+            # NB: calculating expected_url this way is a hack, since we don't know it
+            # when check_response is called.... can we think of a better strategy? Do we
+            # want `check_response` to return the response, for optional further testing?
```
jcushman: I briefly had `check_response` return response, but I ended up preferring this:

```
>>> response = client.patch(draft_casebook.get_absolute_url(), as_user=user, follow=True)
>>> check_response(response, ...)
>>> ...
>>> assertURLEqual(response, casebook.get_absolute_url())
```

... so you don't need to remember `assert check_response()` in the typical case. This often ends up as short or shorter as check_response needs less whitespace. I'd be fine with requiring `assert check_response()` though if you like that approach.

_python/test_helpers.py
```diff
@@ -1,7 +1,27 @@
+from django.test.testcases import SimpleTestCase
 from rest_framework.response import Response
+from urllib.parse import urljoin, urlsplit
 
 
-def check_response(response, status_code=200, content_type=None, content_includes=None, content_excludes=None):
+def assertURLEqual(response, expected_url):
```
jcushman: Nit -- does this want to be `assert_url_equal`?

_python/test_helpers.py
```diff
@@ -14,6 +34,11 @@ def check_response(response, status_code=200, content_type=None, content_include
                 content_type = "text/html"
         assert response['content-type'].split(';')[0] == content_type
 
+    if expected_url:
+        if callable(expected_url):
```
jcushman: This workaround is probably not needed given the above?

app/assets/javascripts/lib/requests.js
```diff
@@ -1,8 +1,24 @@
...
+    modal.destroy();
+  }
+}
+
 export function request (url, method, data = {}, options = {scroll: true}) {
-  const csrf_el = document.querySelector('meta[name=csrf-token]');
-  let headers = csrf_el ? {'X-CSRF-Token': csrf_el.getAttribute('content')} : {};
+
+  // For now, separate handling of csrf in the Rails and Django apps.
+  const rails_csrf_el = document.querySelector('meta[name=csrf-token]');
```
jcushman: Django recommends getting the token from the cookie instead of from the DOM, though it's not super persuasive about why:

https://docs.djangoproject.com/en/2.2/ref/csrf/#acquiring-the-token-if-csrf-use-sessions-and-csrf-cookie-httponly-are-false

(see especially the Note, but then sort of undermined by the Warning). Any feeling here about why to prefer the DOM element in our case?

_python/main/views.py
rebeccacremona: Ah! Totally. For whatever reason, it just did not occur to me to call `client` separately first. I'll revise this to do that.

_python/test_helpers.py
rebeccacremona: yep.

_python/test_helpers.py
rebeccacremona: yep.

app/assets/javascripts/lib/requests.js
rebeccacremona: 10% I wasn't totally confident we'd stick with "CSRF_USE_SESSIONS and CSRF_COOKIE_HTTPONLY are False"; 10% I was lazy; 80% I was unduly influenced by the knowledge this particular case probably wanted to be a normal form submission in the future.

Happy to use the cookie instead.

PR #912, opened 2019-11-06
Toc drag and droc

jcushman:
- add client-side html and server route for dragging and dropping TOC items
- fix some edge cases in node reordering function
- add hydrate_params and user_has_casebook_perm view decorators to remove some boilerplate
- remove some permissions doctests -- I want to do something more comprehensive here but saving it for a separate PR
- add fabfile.py, currently with a single command `fab run`
- also threw in `pytest-mock` but didn't end up using it yet
---
- Add `replaceSelector` option for ajax requests to avoid page reload. This should be tested before the next prod deploy as it affects the current TOC as well.


_python/main/templates/casebook_edit.html
```diff
@@ -13,7 +13,11 @@
 {% endblock %}
 
 {% block mainContent %}
-  <header class="casebook-draft casebook" data-casebook-id="{{ casebook.id }}">
+  <header class="casebook-draft casebook" data-casebook-id="{{ casebook.id }}"
+          data-section-id=""
+          data-section-ordinals=""
+          data-resource-id=""
+  >
```
rebeccacremona: Interesting. These empty data attributes are used to make the editable TOC work?

_python/main/views.py
```diff
@@ -38,6 +43,60 @@ def login_required_response(request):
...
+            key = param.split('_', 1)[0]
+            kwargs[key] = get_object_or_404(ContentNode.objects.select_related('casebook'), casebook=casebook_param['id'], ordinals=param_value['ordinals'])
+            kwargs['casebook'] = kwargs[key].casebook
+        if 'casebook' not in kwargs:
+            kwargs['casebook'] = get_object_or_404(Casebook, id=casebook_param['id'])
+        return func(request, *args, **kwargs)
+    return wrapper
+
+
+def user_has_casebook_perm(casebook_method):
```
rebeccacremona: Nice. This is so much tidier.

_python/main/views.py
```diff
@@ -38,6 +43,60 @@ def login_required_response(request):
         return redirect_to_login(request.build_absolute_uri())
 
 
+def hydrate_params(func):
```
rebeccacremona: Interesting! I don't know if you saw: the Rails app also handles this at a higher level: https://github.com/harvard-lil/h2o/blob/develop/app/controllers/content/node_controller.rb#L6

_python/main/views.py
```diff
@@ -361,7 +407,7 @@ def patch(self, request, casebook_param):
             ...     content_excludes="You are viewing a preview"
             ... )
             >>> with assert_raises(Casebook.DoesNotExist):
-            ...     draft_casebook.refresh_from_db();
```
rebeccacremona: Doh

PR #918, opened 2019-11-12
Fix logout link for python pages

jcushman:
This will fix the Sign Out link that is currently broken on python pages on h2o-experiment and prod:

image

When deploying, you can add `/user_sessions/logout/` to the nginx config -- the ruby pages use a different route, so won't be affected.


bensteinberg:
> When deploying, you can add /user_sessions/logout/ to the nginx config -- the ruby pages use a different route, so won't be affected.

Just to confirm: you mean, let Django handle that route?

jcushman:
Yep! Django should handle that route for all methods, not just GET.

PR #921, opened 2019-11-13
Re-add /pages/ to template urls for now

jcushman:
Note that because these were permanent redirects, you may have a /pages/about -> /about redirect hanging around in your browser that confuses things when testing.


bensteinberg:
I believe this means I can remove the whitelist entry for `~^/about/?$` but don't strictly need to? We'd need it back again after the eventual removal of `/pages` from these routes.

jcushman:
That sounds right!

PR #924, opened 2019-11-18
Casebook export

jcushman:
This PR adds initial support for exporting docx files from Python. I expect we'll find some edge cases that need fixing!

For testing, we should check out some real-life casebooks, but (like the Rails side) there are also a bunch of test exports in test/files/exports. This is a great place to start kicking the tires -- either checking for problems with the current exports, or adding edge cases to test_export.py and running `pytest -k test_export --write-files` to check the results.

As a performance note, export now uses O(1) sql queries -- casebooks take 5 sql queries to render regardless of casebook size. On my Docker environment large casebooks now spend about 1.5 seconds in the generating-pandoc-html phase, and 25 seconds in the running-pandoc phase. We might want to push on whether pandoc can be faster, as those two phases should be roughly equally complex under the hood.

Some other misc refactoring and features and stuff that came along for the ride:

- Resource.resource is now cached on the model instance on first access. This means this works as expected, when otherwise it would not:
    ```
    node.resource.content = 'new html'
    node.resource.save()
    ```
    I don't believe there are downsides, but flagging it in case it causes anything surprising.
- tweak @perms_test so it can handle valid Django template values like `'"some string"'` as well as pytest fixtures like `'casebook'`
- add rails_routes.txt file, useful both for finding URLs when porting new routes, and tracking which routes are left to port
- continuing the experiment with URL converters, I added a way to register model url converters like `<casebook:my_casebook>`. This is a URL converter that automatically provides a model instance to the view, or raises 404 if the ID isn't found. In `reverse()`, the url will only match if provided an object of the right type, which allows us to have multiple routes with the same name that will resolve `{% url "export" some_case_or_section_or_resource %}` correctly. This goes on my "possibly too clever, may rip out later" pile ...
- generalize `@user_has_casebook_perm(foo)` to `@user_has_perm('casebook', foo)` so we can also use it for views that take a resource or section instead of a casebook


rebeccacremona:
So jazzed. Pulling this down and kicking the tires.

jcushman:
Pulling in while Becky continues review ...

PR #927, opened 2019-11-18
Fix TOC depth in export template.

rebeccacremona:


rebeccacremona:
Oh right, I have to update the test files.

PR #935, opened 2019-11-21
Disable rails sunspot indexing

jcushman:


bensteinberg:
Maybe lose https://github.com/harvard-lil/h2o/blob/develop/.travis.yml#L37 FWIW

PR #936, opened 2019-11-22
Delete Casebooks, Sections, and Resources

rebeccacremona:
Delete sections and resources from casebooks, and fix up the ordinals of the remaining sections and resources as necessary.

To do this with the fewest number of DB queries, making some unusual moves (thanks for suggesting the technique, Jack!): I'm setting `on_delete` to `DO_NOTHING` in a number of places, to prevent Django from making unnecessary SELECTs and DELETEs in situations where we know it makes no sense... for instance.... when deleting a section or a resource, there is no need to adjust the ContentCollaborators table.

There's definitely a future where this is done more cleanly, by dropping this ContentNode + proxy class-based data model..... I think this ought to hold in the meantime.


rebeccacremona:
It occurs..... TextBlock objects should be deleted when their corresponding Resource objects are deleted. I don't see that in the Rails source, but I'll verify on Monday... and either way, will add that to this PR.

rebeccacremona:
Confirmed: the Rails app does NOT delete TextBlocks or Defaults/Links when their corresponding resources are deleted.

PR #937, opened 2019-11-25
Admin query optimization and casebook permissions copying

jcushman:
The headline features here are to have user permissions for casebooks also get applied to their drafts (and vice versa) in the admin, and to improve the performance of some admin screens. This moved a bunch of stuff around under the hood but hopefully didn't change much actual functionality. Details:

- Add cached properties `draft` and `draft_of` to `Casebook`, which work more or less like they will if we turn the relationship into a one-to-one field in the future
- Drop the existing `drafts()` method
- Saving a casebook in the admin edit screen copies its user collaborators to its draft/draft_of
- Refactor admin.py a bit with an edit_link() helper
- Optimize some admin.py screens with get_queryset 
- Have the casebook collaborator methods (like `.owner`) always use `contentcollaborator_set.all()` so `.prefetch_related('contentcollaborator_set__user')` will work on a collection of casebooks to prevent further queries
- for convenience, `clone_model_instance` helper now accepts arbitrary properties to apply to the new instance

### DIFF
--- a/_python/conftest.py
+++ b/_python/conftest.py
@@ -385,7 +385,7 @@ def full_casebook_with_draft(full_casebook_parts_factory):
         >>> assert has_draft.has_draft and not draftless.has_draft
         >>> assert all(node.has_draft for node in has_draft.contents.all())
         >>> assert has_draft.is_public
-        >>> assert has_draft.drafts().is_private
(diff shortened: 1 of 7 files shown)


jcushman:
The admin screen now shows clone/draft/draft_of relationships like this:

image

PR #953, opened 2019-11-27
Quick admin fix

rebeccacremona:


jcushman:
Lol, the result of a last-minute find-replace after flake8 complained that I was using `lambda`s for those. Thanks!

PR #957, opened 2019-11-27
Sanitize headnote and textblock html

jcushman:
- Update the `sanitize()` method to allow precisely the html that either (a) can be created by ckeditor or (b) is already in the database. The already-in-the-database part is probably over-inclusive and we could start stripping some of it out. For now the lists in `sanitize.py` are good documentation of where we're starting from.
- Update `SanitizingMixin` to only run if the value has changed.
- Add test to confirm that annotation shifting and sanitizing work together on TextBlock.
- Add a couple of fab tasks, `report_tags` and `compare_sanitized_html`, that may be primarily of historical interest. Not sure how we want to handle this -- it's kind of nice to keep around examples of how problems were solved, especially since this isn't the last time we'll have to do something like this, but it's also kinda noisy. Maybe there's somewhere else one-off fab tasks should live.


_python/main/sanitize.py
```diff
@@ -0,0 +1,140 @@
...
+    
+    - ones can be created by ckeditor, our frontend wysiwyg editor.
+    - "legacy" ones that already exist in legacy HTML in the browser and that we don't want to strip yet.
+    
+    The second source could be reduced as we decide that certain things don't have to be preserved.
+"""
+
+## helpers ##
+
+def get_words(s):
```
rebeccacremona: This helper is a great idea! I definitely would have just made a list of individual strings.... This is SO much easier to read!

_python/main/sanitize.py
```diff
@@ -0,0 +1,140 @@
...
+    # styles created by ckeditor
+    allowed_styles = (
+        # ckeditor image tags
+        get_words("border-width border-style margin float width height")
+        # ckeditor table tags
+        | get_words("width height")
+    )
+
+    # legacy styles
+    allowed_styles |= get_words("""
```
rebeccacremona: Wow. I had no idea there were all these nice operators for working with sets (`|` for union, `|=`, etc.) Huzzah!

PR #966, opened 2019-12-05
Update pandoc

bensteinberg:
This is a work in progress; export tests currently fail.


bensteinberg:
Note that export tests on the Rails side still fail with 99c374c -- but they're not running in Travis.

PR #973, opened 2019-12-10
Edit TextBlocks

rebeccacremona:
This PR enables the editing of TextBlock content via the UI and via the Django admin.

Doing so revealed a few hitches in the saving/annotation-moving/annotation rendering process: 
1) sanitiziation (via bleach) was taking place AFTER annotation placements were recalculated, instead of BEFORE. 
2a) CKEditor posts newlines as `\r\n`; Django admin uses `\n`; etc.
2b) bleach converts `\r` to `\n`, doubling the number of newlines on save if `\r` is present
3) The javascript that places the annotations in-browser does not expect whitespace following block-level elements. If such whitespace is present, and if it happens to be included in an annotation, the js wraps the whitespace in annotating spans, causing a re-numbering of resource paragraphs and other undesirable visual effects.

So, we decided:
1) It was simpler/more flexible to explicitly override models' `save` methods, rather than using a custom `Sanitizing*Field`
2) To convert `\r\n` and `\r` to `\n` before save
3) To strip trailing whitespace after block tags, instead of rewriting the JS.

These changes required an adaptation of our tests, to make the tests less sensitive to difference in whitespace. 

Let's definitely kick the tires on this one on stage.


_python/main/admin.py
```diff
@@ -401,7 +401,7 @@ def related_resources(self, obj):
 
 class TextBlockAdmin(NonLoggingAdmin):
     # Content is readonly until we implement the annotation-shifting logic on the python side
```
jcushman: Outdated comment

_python/main/normalize.py
```diff
@@ -0,0 +1,3 @@
+
+def normalize(html):
```
jcushman: Extra file?

PR #974, opened 2019-12-10
User auth and password flows

jcushman:
Here's a first pass at bringing the user auth views to Python, using the built-in Django views as much as possible:

- create user and confirm email address
- login
- forgot password
- change password (this was previously on the same page as user profile edit, but I made it a button from that page so we can use the built-in Django stuff)


_python/main/models.py
```diff
@@ -2620,6 +2630,19 @@ def set_password(self, raw_password):
...
+    """
+        Register signal to record user login details on successful login.
+    """
+    user.last_login_at = user.current_login_at
+    user.current_login_at = timezone.now()
+    user.last_login_ip = user.current_login_ip
+    user.current_login_ip = get_ip_address(request)
+    user.login_count += 1
+    user.save(update_fields=['last_login_at', 'current_login_at', 'last_login_ip', 'current_login_ip', 'login_count'])
+user_logged_in.connect(update_user_login_fields)
```
rebeccacremona: Nice idea to use a signal. I wonder if this level of tracking is handy, or is something we might drop going forward. Since you've done all the work, I suppose no reason to drop it.

_python/main/forms.py
```diff
@@ -1,15 +1,27 @@
...
 
 from main.models import ContentNode, Default, TextBlock, User
 from main.utils import fix_after_rails
 
 
+# Monkeypatch FormHelper to *not* include the <form> tag in {% crispy form %} by default.
+# Forms can opt back in with self.helper.form_tag = True. This is a more useful default
+# because then we can use {% crispy form %} in built-in Django views like registration/password_change_form.html
+# without having to override the form.
+FormHelper.form_tag = False
```
rebeccacremona: Ah! Good idea! Wouldn't have thought of this.

_python/main/models.py
jcushman: For documentation -- the default Django behavior is to use this same signal to set `last_login = timezone.now()`, so we could fully switch to the Django way by renaming current_login_at to last_login and dropping the signal and the other fields. I can imagine situations where more logging is handy though. 🤷‍♂️

PR #1004, opened 2019-12-17
rich text editor in admin

rebeccacremona:
This PR adds the rich text editor to headnotes (Casebook, Section, Resource) and content (Case, TextBlock) in the Django admin. The editor includes a source code editor, available by pressing the `<>` button in the menu bar. That opens a modal allowing you to edit the HTML directly:
image

The modal includes a "save" button, which means that the edits you've made in the modal are persisted to the main page, and shown in the WYSIWYG there... it does NOT save the Django model instance: you still need to press the "save" button in the Django admin. I'd like to change the modal's button to read something other than "save", for instance, "preview", but I haven't figured out a good way to do that yet.

I have not tested this much to see how saved cases are affected. We should definitely test some more.


jcushman:
Very snazzy!

PR #1005, opened 2019-12-17
Support add-user and change-password in admin

### LINKED ISSUE #999: Enable admin creation of users
Needed for international profs who do not have EDU address but who are confirmed as affiliated with educational institution

jcushman:
This lets our UserAdmin inherit from Django's built-in UserAdmin, which comes along with a free change-password form. Also overrides the new-user form to set a random password, which means the reset password form will work.

*Suggested workflow for creating accounts that don't end in .edu:*

- Create the account in the Django admin, with no password and with verified_email off
- Either use the forgot-password form or tell the user to use it, and they should then be able to verify their email, set a password, and get the welcome email as normal

Fixes #999

### DIFF
--- a/_python/main/admin.py
+++ b/_python/main/admin.py
@@ -1,6 +1,7 @@
 from django import forms
 from django.conf import settings
 from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
 from django.db.models import Q, Count
 from django.urls import reverse
 from django.utils.html import format_html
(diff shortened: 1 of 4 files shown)


_python/main/admin.py
```diff
@@ -428,12 +429,52 @@ def live_annotations_count(self, obj):
...
+    def save(self, commit=True):
+        user = super().save(commit=False)
+        self.instance.set_password(User.objects.make_random_password(length=20))
+        if commit:
+            user.save()
+        return user
+
+
+class UserAdmin(BaseAdmin, DjangoUserAdmin):
+    filter_horizontal = ('roles', 'verified_professor')
```
rebeccacremona: Whoops! I think this is a little broken. Django says it can't have a horizonal filter on `'verified_professor'` since that isn't a many-to-many field.
```
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
<class 'main.admin.UserAdmin'>: (admin.E020) The value of 'filter_horizontal[1]' must be a many-to-many field.
```

rebeccacremona:
Thanks for working on this Jack!

I don't think that suggested flow works at the moment: I don't see that there is a way for users to verify their email address. Without a verified email address, the usual "forgot password" flow does not work (you aren't sent an email).

Should we set things up so that, if you request a password reset with an existing, but unverified email address, you get the verification email?

If the admin creates an initial password, the user still won't be able to log in, unless the admin also sets `verified_email=True`, since that's a synonym for `is_active` here. So maybe that should be the standard flow (regardless of how we want the reset flow to work)?

I also notice that the embedded admin form reads "No password set" for all users... template must be pointed at the standard field, instead of ours.

I can have a go at these tomorrow if you prefer!

Issue #872, opened 2019-09-26
Running list of data questions


rebeccacremona:
Addressed, Round 1:
- Q: There are a lot of CaseCourts with 0 associated cases. Data problem, or just unnecessary objects? A: We can delete these.
- Q: Why do "Default" objects have a "created by import" field (always false, I think)? A: Let's drop this field.
- Q: There are some Casebooks with no "owners". How can that be? Maybe users were deleted, but their casebooks weren't? A: Let's assign these to an owner.
- Q: TextBlocks have a 'version' field, currently populated with integers from 1-3. What is this? A: Unknown, and in any case no longer useful. We are free to drop.

- Q: In the Rails admin, I see newly created Defaults that belong to user 0. What are those? Also, I see some TextBlocks that aren't associated with a user. Some of them have related resources. What are those? A: These user column for both these models was a mess... nulls, 0s, no-longer-existing users... and, in any case, is redundant w/the owner of the associated resource object/casebook. Deleting. If we want to track authorship of these objects independently of casebooks in the future, we'll need new code + data anyway.

- Q: There are lots of TextBlocks and Defaults that aren't associated with any Resource objects. Why? A: We're going to delete these. We think that some (or all) of them were accidentally not deleting, during the merging of drafts. Others are likely from other deletion processes.
```
>>> Default.objects.exclude(id__in=Resource.objects.filter(resource_type='Default').values('resource_id')).count()
34590
>>> Default.objects.filter(id__in=Resource.objects.filter(resource_type='Default').values('resource_id')).count()
12361
>>> 12361 + 34590
46951
>>> Default.objects.count()
46951

>>> TextBlock.objects.exclude(id__in=Resource.objects.filter(resource_type='TextBlock').values('resource_id')).count()
21783
>>> TextBlock.objects.filter(id__in=Resource.objects.filter(resource_type='TextBlock').values('resource_id')).count()
9934
>>> 21783 + 9934
31717
>>> TextBlock.objects.count()
31717
```

- Q: Roles: can we delete all the "asker" roles? How about "rep"? A: Yes.

[...]

PR #1015, opened 2019-12-19
Migrations galore

rebeccacremona:
This is a first round of DB cleanup, as described in https://github.com/harvard-lil/h2o/issues/872. It is easiest to read through commit-by-commit, I think, side-by-side with the issue. OR, for me to walk through in person, which I am delighted to do.

Marking as a WIP because... it's massive, I just did a giant rebase, and I'm moments from vacation and should not be trusted :-)

For the new year!


rebeccacremona:
We might want to conditionally create the user w/ address `orphaned...` in the migration.

rebeccacremona:
Still feeling good about this DB surgery, having now re-read everything in the new year.

I think, on reflection, that we do NOT want to create the user with `email_address='orphaned@opencasebook.org'`, but rather, want to create it manually BEFORE merging/running the migrations, to avoid password-related weirdness.

Up to the team how much anyone else wants to review before merging...

In any case, we want this in a separate deployment than the pending big Rails -> Django deployment.

PR #1016, opened 2019-12-19
Delete Rails-only files and tweak repository structure

rebeccacremona:
This PR deletes the Rails application and associated files from the repository. It simultaneously tweaks the directory structure to make H2O's structure more similar to that of other LIL projects.

### Deleted

Of the files and directories at the top-level of the repository, the following were deleted:

Files:
.browserslistrc
.env.example
.ruby-gemset
.ruby-version
Gemfile
Gemfile.lock
Guardfile
Rakefile
babel.config.js
config.ru
package.json
postcss.config.js
test_export.rb
yarn.lock

Directories:
bin/
config/
db/
log/
public/
solr/
test/
lib/ (except for lib/pandoc)

### Renamed
_python -> web
lib/pandoc -> services/pandoc

(which required minor updates to docker and django config)

### Updated
.circleci/
.dockerignore
README.md
docker/
docker-compose.yml

### Untouched
LICENSE
.gitignore

### End result
```
HLSC02VK05ZHTDG:h2o rcremona$ ls -1
LICENSE
README.md
docker/
docker-compose.yml
services/
web/
```

## Meaningful changes

This intends to make only the smallest number of real changes:

1) It changes the name and credentials of the expected database from the Rails conventions (`h2o_dev`, etc.) to the docker/postgres defaults (`postgres`) etc., to simplify config and the initialization of a new database without the assistance of Rails.

**N.B. current devs should create new databases after we merge this PR**
```
docker volume rm h2o_db_data
docker-compose up -d
bash docker/init.sh -f /path/to/a-very-recent-snapshot.dump
```

2) It renames the "python" service to "web"... which formerly described the Rails service

3) I removed the instructions about merging PRs without creating merge commits.

## How do we want to review this?

Please let me know anything I can do to make this easier, and to make us feel good about the PR.

A potential strategy: 
- do we see changes that are NOT described by the above description?
- were you expecting to see anything in the above description that you do NOT see?
- (the tests indeed all pass)
- ?

It's a ridiculous PR. I fully understand it is a ridiculous PR.


rebeccacremona:
(Yah, just flailing here with the changes to CircleCI. Obviously not right. FWIW tests pass locally.)

bensteinberg:
I think you need to put `lib/pandoc` back, or at least `lib/pandoc/table_of_contents.lua`.

rebeccacremona:
> I think you need to put `lib/pandoc` back, or at least `lib/pandoc/table_of_contents.lua`.

Ha. Right. Whoops.

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             else
               echo "Building new docker images"
               docker-compose build
-              docker save -o ~/docker-cache.tar h2o python
+              docker save -o ~/docker-cache.tar python
```
rebeccacremona: @bensteinberg Do you know what this should be? `h2o-python python postgres`?

bensteinberg: just `python`, I think -- only the containers we build

bensteinberg: Ah, no, sorry -- `web`?

rebeccacremona: Thanks! If [`docker save`](https://docs.docker.com/engine/reference/commandline/save/) takes image names as args, and if we only want only the ones we build, then I think `h2o-python` is correct... which we can rename after a time, once the ruby is long gone.

PR #1018, opened 2020-01-07
Correct DB initialization and instructions

rebeccacremona:
In Rails land, `init.sh` created test and dev databases, and then, if run with `-f` and provided a pg_dump file, it would seed the database.

Now there is no need to manually create any databases: we only need one, and we are using the default that comes with the postgres docker image.

This PR updates the README and tweaks the init script. 

It retains use of the `-f` flag, which is a slightly awkward API, now, because it is mandatory.... Shall I remove it? I am pretty rusty on bash script argument parsing, and am a little hesitant to dive back in, especially if we think we might want the existing framework to parse more flags in the future (for instance, running with or without `disable-triggers`). But, I am happy to do so if this feels too peculiar!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,18 +18,13 @@ The first time this runs it will build the Docker images, which
 may take several minutes. (After the first time, it should only take
 1-3 seconds.)
 
-Finally, initialize an empty database...
-
-    $ bash docker/init.sh
-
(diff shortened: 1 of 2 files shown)


rebeccacremona:
I will verify this actually works, once we have a new DB snapshot.

PR #1024, opened 2020-01-07
Rename 'Default' model to 'Link'

rebeccacremona:
For reasons lost to history, "Link" objects have long been called "Default" objects, in the codebase and database.

This PR renames the model to Link.

It does not alter the name of the database table: all of H2O's models' currently have `db_table` set, and it doesn't make sense to me to alter the name of this one single table. It seems to me that, instead, we should either rename all the table names at once and remove the `db_table` overrides, or make no changes at all. Does that seem right to you too, @jcushman?

Since the automated tests instantiate new fixtures, rather than using static, to-be-migrated data, we should test Link-related functionality manually on stage to make sure the migrations were sufficient: I have tested locally and believe they are working just fine:
- existing casebooks with Link resources still display those Links in the TOC
- you can navigate to them
- you can export them
- you can clone them
- you can add new Link resources to drafts
- you can update existing Link resources within drafts
- admin pages with references to Links work and appear as expected
- am I missing anything?


holtchesley:
Having just looked over the models and been confused by `Default`, I like this change. Were there already plans to change the names of some of the other tables? If there were, it makes sense to me to do it in one fell swoop, otherwise it seems fine to change it for just this table for consistency's sake.

rebeccacremona:
On reflection, I'm not sure whether we talked about renaming the tables. A quick experiment demos that Django will happily handle the renaming for us, if we just lose the:
```
class Meta:
        db_table = 'custom_name' 
```
and migrate per usual.

I like the idea of doing so, for all tables, for tidiness. I'm not thinking of any real downsides.... other than adding a layer of complexity to any archeological explorations into old DBs, which I sure hope we aren't doing with any frequency, if ever... Can you think of any?

> Having just looked over the models and been confused by `Default`, I like this change. Were there already plans to change the names of some of the other tables? If there were, it makes sense to me to do it in one fell swoop, otherwise it seems fine to change it for just this table for consistency's sake.

PR #1027, opened 2020-01-08
Rename tables to Django's defaults.

rebeccacremona:
This PR drops custom database table names for all in-use models. (I left the custom table names for the few legacy tables we've retained, pending more research into their contents.)

Locally, this seems to have worked without any manual interventions. The migration only took a couple seconds, even with the recreation of all our indexes. Tests are passing, and after migrating my local DB, all app functionality, including search, seems fine. 

I'm not sure if it will be necessary to [manually re-create the search view](https://github.com/harvard-lil/h2o/blob/develop/web/fabfile.py#L51) after deploying, but it's probably a good idea. I did so locally via the Django shell.


jcushman:
@bensteinberg I think it is in fact necessary to run `fab create_search_index` after applying this -- go ahead whenever you like.

Issue #1026, opened 2020-01-08
Migrate superadmin role to standard django user fields


jcushman:
The `Role` and `RolesUser` models currently serve only one purpose, which is to label our staff user accounts as `superadmin`. We can delete those and switch to the standard Django way of doing admins:

* add `is_superuser` and `is_staff` fields to `User`
* set those to true for current superadmins in a data migration
* delete a bunch of custom stuff from `User`: has_module_perms, has_perm, is_superuser, is_staff, is_superadmin, has_role
* replace existing uses of any of that stuff with `is_superuser`

In the future if we add more user categories, like "case admin," we can do that through the built-in django groups feature.

PR #1028, opened 2020-01-08
Switch to standard Django permissions, instead of using custom roles.

rebeccacremona:
As per https://github.com/harvard-lil/h2o/issues/1026, TODOs in the code, and a remark about Roles in https://github.com/harvard-lil/h2o/issues/872, this PR switches us over from custom permission-related code to Django's standard `is_superadmin`, `is_staff`, and `is_active`.

We set `is_active` to False by default, and toggle it to True when users verify their email addresses and set a password. We won't use it instead of deleting once-active users, as recommended by Django's default helptext for that field, so I have not included that helptext!


rebeccacremona:
Hmm. I seem to have broken `pyflakes` linting.  Investigating.....

rebeccacremona:
Ooh, terrible, `pyflakes` was borking on `from django.utils.translation import gettext_lazy as _`, which I took right out of the Django source code.... I don't need it, so just removed it, but not good!

web/main/models.py
```diff
@@ -2431,18 +2401,19 @@ def related_resources(self):
...
-    # used to assign super_admin or case_admin status
-    roles = models.ManyToManyField(Role,
-        through=RolesUser
-    )
+    is_staff = models.BooleanField(default=False)
+    is_active = models.BooleanField(default=False)
+
+    # differentiate between real User model and AnonymousUser model:
+    is_authenticated = True
+    is_anonymous = False
```
rebeccacremona: @jcushman can you say a word about why we have to set `is_authenticated` and `is_anonymous` manually? I should be able to figure out, but I'm not getting it.

jcushman: Looks like we don't need to anymore! When those two lines were added we weren't inheriting from `AbstractBaseUser` yet.

rebeccacremona: Aha! Thanks!! I'll delete these lines too!

rebeccacremona: Deleted. And, corrected the reverse migration per discussion in Slack.

PR #1067, opened 2020-01-31
Try buster

bensteinberg:


bensteinberg:
Hm, is that all there is to it?

Note that although buster uses Postgres 11, this PR does not update the database container, which is still 9.6.

bensteinberg:
Also note that once we merge this, there will be a brief delay before we're running buster minions and hence ready to deploy new code.

holtchesley:
This looks good locally with my initial testing. I haven't noticed anything misbehaving with basic edit/clone/view/export.

PR #1088, opened 2020-02-20
Dismiss note annotations in FF friendly way

holtchesley:
Firefox doesn't handle relatedTarget for focusOut events in a way we are trying to use it.
Use vue directive for detecting clicks outside the note annotation save box.


holtchesley:
Superseded by #1089

PR #1110, opened 2020-03-09
Delete CaseCourt associated indexes and columns

holtchesley:


holtchesley:
I liked dropping this table so much that I decided to do it twice. In "learning experience" notes, django doesn't handle references to models that don't exist very well, and it's important to remove all references(column, field, index) to a table before the table, as Django's migrations do a bit of validation before they run, and recovery becomes convoluted.

Issue #968, opened 2019-12-06
Enable large paraphrases of text


abziegler:
E.g. summarizing findings of fact or a dissenting opinion. Our "replace text" feature works for short paraphases of elided text, but not for long summaries.

abziegler:
Here's an example ... the "replacement" text extends for another words or so, but everything after "for" is not displaying.

![image](https://user-images.githubusercontent.com/2140548/80371425-e5fc2880-885f-11ea-806d-77ce0239f813.png)

holtchesley:
This was an odd issue with FFDevEdition oninput events interacting in an odd way with content editable. Work-around introduced.

PR #1143, opened 2020-04-27
Workaround for FFDevEdition contenteditable oninput issue

holtchesley:
Also properly throw 404 instead of 500 when looking up nonexistent casebooks.


holtchesley:
Should address #968

PR #1144, opened 2020-04-27
Whitelist University of Nairobi

holtchesley:


bensteinberg:
LGTM; we could put this in settings and manage the list outside the repo if that makes more sense.

holtchesley:
That makes sense, I just hadn't prioritized it yet.

Issue #941, opened 2019-11-25
Add resource failure should be noisy


abziegler:
When author searches for a case via Add Resource, and that case can't be found, H2O should say so explicitly and should nudge author to check accuracy of citation, retry, etc.. Right now failures are silent and mysterious.

This matters because the CAP integration requires precise citation accuracy or it won't match and import case, so there's likely to be some failures.

Issue #830, opened 2019-07-22
Adding a new resource with no name throws an error


bensteinberg:
There should be a user-facing validation step rather than
```
An ActiveRecord::RecordInvalid occurred in sections#create:

  Validation failed: Resource name can't be blank
  app/controllers/content/sections_controller.rb:17:in `create'
```

rebeccacremona:
In the Django app, this no longer 500s: instead, it catches it and returns 400 Bad Request, plus a validation error message... but the JS doesn't yet handle it, and the modal just closes. Obviously we should just display the error message, instead :-)

PR #1164, opened 2020-05-27
Add resources backend improvements

holtchesley:


holtchesley:
Addresses #830 and #941 as well as a few fix_after_rails fns.

Issue #948, opened 2019-11-27
Disable draggable for contents of casebook in published state


abziegler:
Currently, contents of published casebooks can be dragged (but not dropped). Contents of published casebooks should not be draggable.

abziegler:
Going to re-open this b/c we had some user confusion, and it seems worth doing.

PR #1200, opened 2020-07-28
Disable DnD on published ToCs

holtchesley:


holtchesley:
Addresses #948 in a more resilient fashion.

PR #1207, opened 2020-08-06
Provide error views context.

rebeccacremona:
I missed an important step in https://github.com/harvard-lil/h2o/pull/945.

Some of the default error handlers render their templates without a request context.... so our custom context processor does not get called, so the template variables we inject via `TEMPLATE_VISIBLE_SETTINGS` are not available. As a result, things like footer links and "contact us" links are broken, when the pages are actually rendered in production, unless we provide our own custom error handling views.

This has historically proved super finicky and hard to remember. So, I've added tests to make sure the context variables are actually available. These tests required grabbing a copy of the Django test client from Django 3.0, which adds an argument, `raise_request_exception`, letting you toggle whether or not exceptions raised during the request should also be raised in the test. For these tests, we need `raise_request_exception=False`. I also had to upgrade `pytest-django`. (With the older version, these tests succeed when the `test_errors` module is run alone, but fail when the entire suite is run.)

For good measure, I also tested that error emails are sent as expected... which has also proved finicky. That required adding the `mail_admins` setting to testing log config.

I attempted to set things up so that error emails could be inspected in dev as well, but got into a tangle where BOTH the custom error email AND Django's standard error email were being sent. In Perma and CAP, that's solved by a very different base logging setup involving `deepcopy`.... a more extensive change than I felt confident making to H2O.... If you aren't getting dropped or duplicate error emails in prod now, I don't want to mess that up LOL.


rebeccacremona:
Oh. CircleCI is set up to use the system's python for flake8, which is `PYENV_VERSION=3.5.2`.... which means flake8 is going to bork if you use any f-strings anywhere, even though the app is running python 3.7.

Bummer.

PR #1313, opened 2021-03-03
Update to welcome email

cath9:


holtchesley:
Looks like we had a test checking for some of the old copy. I'll fix that up when I get a chance.

bensteinberg:
I believe this is a plain-text email, not HTML, so `<a>` and `<b>` tags won't work as is. I think we chose plain text over HTML mail as being more deliverable.

cath9:
My bad. Deliverable as in won't get caught by filters?

bensteinberg:
Not at all! Yeah, HTML in email is one of many signals that spam filters observe, but I don't know how heavily weighted it is, or whether its weight is conditioned on other signals, etc.

It's also possible that we made the decision on some other basis, like aesthetics or maintainability, or some combination.

cath9:
After 20 minutes of Google research, it seems like there are some pros to plain text around delivery. So let me go back and see about making changes to the email to strip some formatting.

PR #1413, opened 2021-10-06
Enable Codecov

holtchesley:


bensteinberg:
Weird. Do we need to specify a version of codecov? Or maybe just set it up with an orb like in https://github.com/harvard-lil/perma/blob/develop/.circleci/config.yml

PR #1415, opened 2021-10-07
Cap export attempts

holtchesley:


holtchesley:
Allow for 3(configurable) attempts to export a casebook and then lock further attempts to export the casebook. 
Update admin panel to allow locking/unlocking exports + searching.
Only meters full casebook exports for now. We've not had non-content related section export issues yet.

holtchesley:
Addresses #1414

PR #1416, opened 2021-10-07
Ensure updated CA certs

bensteinberg:


bensteinberg:
Another fix might be to update the machine image.

PR #1434, opened 2021-10-13
Fixup export authorship ordering, show year in casebook history

holtchesley:


web/main/models.py
```diff
@@ -1999,11 +1999,11 @@ def export(self, include_annotations, file_type='docx', export_options=None):
         children = list(self.contents.prefetch_resources().prefetch_related('annotations')) if type(
             self) is not Resource else None
 
-        current_collaborators = self.casebook.primary_authors
+        current_collaborators = set(self.casebook.primary_authors)
         cloned_from = {cn.casebook for cn in self.ancestor_nodes.prefetch_related('casebook')
                                                    .prefetch_related('casebook__contentcollaborator_set')
                                                    .prefetch_related('casebook__contentcollaborator_set__user')
-                         if cn.casebook.primary_authors ^ current_collaborators}
+                         if set(cn.casebook.primary_authors) ^ current_collaborators}
```
holtchesley, line 2006: This fixes a bug that would prevent export for cloned casebooks. I introduced it https://github.com/harvard-lil/h2o/commit/6ed792207e364f22ecab9f4af00281f07de0cc29 when I added support for orderable authors (and changed the primary author ds from a set to a list).

PR #1445, opened 2021-10-28
Retrieve legacy casebook nodes as ContentNode, else Section or Resource

### LINKED ISSUE #1444: Cast ContentNodes to subclasses with old and new-style casebooks
`Casebook`s used to be `ContentNode`s, along with `Section`s and `Resource`s. They aren't anymore: they are an independent model.

`Casebook` objects that were created from the old model have a reference to the ContentNode they were created from in `old_casebook_id`, which have been kept. 

```
>>> con_law = Casebook.objects.get(id=531)
>>> con_law.old_casebook_id
128853
>>> ContentNode.objects.get(id=128853)
<ContentNode: Constitutional Law (128853)>
```

That ContentNode does not itself have an `old_casebook_id` or a `casebook_id`:
```
>>> ContentNode.objects.get(id=128853).old_casebook_id
>>> ContentNode.objects.get(id=128853).casebook_id
```

The casebook's component sections and resources have copies of both:
```
>>> all([(c.old_casebook_id, c.casebook.id) == (128853, 531) for c in con_law.contents.prefetch_resources()])
True
```

The sections and resources for new books, of course, have no `old_casebook_id`:
```
>>> all([c.old_casebook_id is None for c in new_torts_book.contents.prefetch_resources()])
True
```

So, [the casting function](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1642) isn't quite working as desired: all new `ContentNode`s are getting caught by the [first branch](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1656), which only wants to catch retired Casebook-flavor `ContentNode`s like `128853`.

This should be easy to tweak; PR forthcoming.

rebeccacremona:
I think this fixes https://github.com/harvard-lil/h2o/issues/1444

```
>>> from main.models import Casebook, ContentNode

>>> new_torts_book = Casebook.objects.get(id=2566)
>>> set([type(c) for c in new_torts_book.contents.prefetch_resources()])
{<class 'main.models.Resource'>, <class 'main.models.Section'>}

>>> con_law = Casebook.objects.get(id=531)
>>> set([type(c) for c in con_law.contents.prefetch_resources()])
{<class 'main.models.Resource'>, <class 'main.models.Section'>}

>>> ContentNode.objects.get(id=con_law.old_casebook_id)
<ContentNode: Constitutional Law (128853)>
```


rebeccacremona:
Hmm, interesting test failures, especially the ones referencing `HistoricalCasebook` and `HistoricalContentNode`. I'll dig in in the AM.

holtchesley:
I think the `Historical*` is just an unfortunate side effect of the Django-simple-history and doctest interacting. The line numbers should reference the actual written classes though.

rebeccacremona:
Okay, this makes a meaningful difference to the [clone_to](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L2479) method on casebooks. 

Code on develop now:
`from_casebook` looks like this:
```
['Casebook<1>: Some Title 0'
' Section<1>: Some Section 0'
'  ContentNode<2> -> TextBlock<1>: Some TextBlock Name 0'
'  ContentNode<3> -> LegalDocument<1>: Legal Doc 0'
'   ContentAnnotation<1>: highlight 0-10'
'   ContentAnnotation<2>: elide 0-10'
'  ContentNode<4> -> Link<1>: Some Link Name 0'
'  Section<5>: Some Section 4'
'   ContentNode<6> -> TextBlock<2>: Some TextBlock Name 1'
'   ContentNode<7> -> LegalDocument<2>: Legal Doc 1'
'    ContentAnnotation<3>: note 0-10'
'    ContentAnnotation<4>: replace 0-10'
'   ContentNode<8> -> Link<2>: Some Link Name 1'
' Section<9>: Some Section 8']
```
We then clone its second section, 
```
'  Section<5>: Some Section 4'
'   ContentNode<6> -> TextBlock<2>: Some TextBlock Name 1'
'   ContentNode<7> -> LegalDocument<2>: Legal Doc 1'
'    ContentAnnotation<3>: note 0-10'
'    ContentAnnotation<4>: replace 0-10'
'   ContentNode<8> -> Link<2>: Some Link Name 1'
```
to `to_casebook`, whose original structure looks like:
```
['Casebook<2>: Some Title 1'
' Section<10>: Some Section 9'
'  ContentNode<11> -> TextBlock<3>: Some TextBlock Name 2'
'  ContentNode<12> -> LegalDocument<3>: Legal Doc 2'
'   ContentAnnotation<5>: highlight 0-10'
'   ContentAnnotation<6>: elide 0-10'
'  ContentNode<13> -> Link<3>: Some Link Name 2'
'  Section<14>: Some Section 13'
'   ContentNode<15> -> TextBlock<4>: Some TextBlock Name 3'
'   ContentNode<16> -> LegalDocument<4>: Legal Doc 3'
'    ContentAnnotation<7>: note 0-10'
'    ContentAnnotation<8>: replace 0-10'
'   ContentNode<17> -> Link<4>: Some Link Name 3'
' Section<18>: Some Section 17']
```

The result is: 
```
['Casebook<2>: Some Title 1'
' Section<10>: Some Section 9'
'  ContentNode<11> -> TextBlock<3>: Some TextBlock Name 2'
'  ContentNode<12> -> LegalDocument<3>: Legal Doc 2'
'   ContentAnnotation<5>: highlight 0-10'
'   ContentAnnotation<6>: elide 0-10'
'  ContentNode<13> -> Link<3>: Some Link Name 2'
'  Section<14>: Some Section 13'
'   ContentNode<15> -> TextBlock<4>: Some TextBlock Name 3'
'   ContentNode<16> -> LegalDocument<4>: Legal Doc 3'
'    ContentAnnotation<7>: note 0-10'
'    ContentAnnotation<8>: replace 0-10'
'   ContentNode<17> -> Link<4>: Some Link Name 3'
' Section<18>: Some Section 17'
' Section<19>: Some Section 4']

(Pdb) Section.objects.get(id=19)
<ContentNode: Some Section 4 (19)>
(Pdb) Section.objects.get(id=19).contents
<ContentNodeQueryset []>
(Pdb) 
```
....which I believe is incorrect. We want the section AND its contents, do we not?

With the change made in this PR, we get the section and its contents:
```
['Casebook<2>: Some Title 1'
' Section<10>: Some Section 9'
'  ContentNode<11> -> TextBlock<3>: Some TextBlock Name 2'
'  ContentNode<12> -> LegalDocument<3>: Legal Doc 2'
'   ContentAnnotation<5>: highlight 0-10'
'   ContentAnnotation<6>: elide 0-10'
'  ContentNode<13> -> Link<3>: Some Link Name 2'
'  Section<14>: Some Section 13'
'   ContentNode<15> -> TextBlock<4>: Some TextBlock Name 3'
'   ContentNode<16> -> LegalDocument<4>: Legal Doc 3'
'    ContentAnnotation<7>: note 0-10'
'    ContentAnnotation<8>: replace 0-10'
'   ContentNode<17> -> Link<4>: Some Link Name 3'
' Section<18>: Some Section 17'
' Section<19>: Some Section 4'
'  ContentNode<20> -> TextBlock<5>: Some TextBlock Name 1'
'  ContentNode<21> -> LegalDocument<2>: Legal Doc 1'
'   ContentAnnotation<9>: note 0-10'
'   ContentAnnotation<10>: replace 0-10'
'  ContentNode<22> -> Link<5>: Some Link Name 1']

(Pdb) Section.objects.get(id=19).contents
<ContentNodeQueryset [<Resource: Some TextBlock Name 1 (20)>
<Resource: Legal Doc 1 (21)>
<Resource: Some Link Name 1 (22)>]>
```

So, I think the change is a good thing.

Oh cool, and yes, with this change we revert the last line of the test to [how it appeared before the casebook model was extracted](https://github.com/harvard-lil/h2o/blame/038bd62cfe378abe060ce540bb59cb85007506cf/web/main/models.py#L1482): that sounds right. It's also clear in this form why we are looking at `dump_casebook_outline(to_casebook)[-7:]`: that's just the new section with its contents and the new resource.

This method's docstring begins, "This is currently called only manually, for extraordinary customer service situations, but would ideally be exposed through the frontend", and, indeed, it doesn't seem to be called anywhere, so if this was indeed a bug, it is unlikely it affected anything.

rebeccacremona:
This also -does- change the creation of HistoricalContentNode objects during deletion: it makes fewer of them. Looking into whether that matters.

rebeccacremona:
Okay, with the code on `develop`, when you [delete a section](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L2636), like section 1.4 of this test fixture casebook, a HistoricalContentNode record is created for ordinals 1, 1.4.1, 1.4.2, and 1.4.3 (4 rows), and then one historical record each for the related text block, link, and two annotations, for a total of 8 inserts. That doesn't seem right: the section's own ordinals, `ARRAY[1,4]`, is absent.

With the code of this PR as is, at this moment, the above only creates the HistoricalContentNode record for `ARRAY[1]` and then the text block, link, and annotations, not for any of the subsections (5 rows): that is clearly wrong.

If, however, I change `HistoricalRecords(inherit=True)` (which isn't exactly how it's intended to be used, but is close...), we then get 9 rows: 
- with the code as on develop, a HistoricalContentNode record for 1, 1.4.1, 1.4.2, and 1.4.3 (4 rows), a HistoricalSection record for 1.4, and the others as usual. The trace for the HistoricalSection row points to `/app/web/main/models.py:2731: return_total, return_dict = super().delete(*args, **kwargs)`, and the trace for the nested HistoricalContentNodes points to `/app/web/main/models.py:2726: child_total, child_deletes = self.contents.delete()`
- with this PR's change, a HistoricalContentNode record for `ARRAY[1]`, a HistoricalSection records for `ARRAY[1,4]`, and HistoricalResource records for 1.4.1, 1.4.2, and 1.4.3, and the others as usual. The trace for 1.4, 1.4.1, 1.4.2, and 1.4.3's records here points to `/app/web/main/models.py:2731: return_total, return_dict = super().delete(*args, **kwargs)`.

I... I -think- that's a step in the right direction: I think, if we want historical records to be created at all, we want a row for `ARRAY[1,4]`?

How any of this affect ACTUALLY restoring casebook contents for people if they are accidentally deleted or changed is unknown to me, though.

If we change `HistoricalRecords(inherit=True)`, new tables are created for HistoricalSection and HistoricalResource, and these are tracked separately. That will probably make the existing historical records harder to use, since they are HistoricalContentNode records. But I'm not really sure how we intend to use the historical records for deleted stuff anyway: not via the django admin.... Since it would be a manual restoration process anyway, maybe it's fine?

To be discussed.

Perhaps we should reconsider whether this `from_db` behavior is desired at all: if it hasn't been working for this long, and no harm done, maybe the level of intricacy it adds isn't worth it...

I'll push the changes w/ `inherit=True`, and bring to the group for discussion.

rebeccacremona:
Brennan thinks the smarter design move here is actually to move away from having proxy models at all, and just deal with ContentNodes directly. If we're going to do that, this change doesn't make sense: better to remove `from_db` instead.

PR #1455, opened 2021-11-09
Never cast ContentNode -> Section/Resource; effectively eliminate Section/Resource proxy models

rebeccacremona:
## Down with proxies and casting

This PR is the opposite of https://github.com/harvard-lil/h2o/pull/1445: instead of fixing `from_db` so that developers only ever work with `Section` and `Resource` proxy objects and not `ContentNode` objects, this PR removes any casting. 

While casting was not working, it proved necessary/convenient to duplicate logic that was provided via `Section`/`Resource` directly on `ContentNode`; this PR removes all copied logic from the proxy model definitions, so that in all cases, logic is defined directly on `ContentNode`. If the required behavior differs for section- and resource-type nodes, a switch is used. 

This effectively empties `Section` and `Resource`: they are now entirely synonymous with `ContentNode` and can be removed shortly. I have not taken that additional step here, to avoid touching too much of the codebase at once.

## ....Down with proxies and casting?

Do we like the direction that this is going, or, having seen it, are we missing the elegance of the proxy models?
- if we lose the proxy models and casting, then we should also lose the URL converters that, when casting was working, ensured you couldn't access a section via a resource URL, or a resource via a section URL, and we should put some other guard in place for any route currently using `section_param` or `resource_param` (probably new route decorator) OR make sure that all routes that handle content nodes can handle content nodes of any variety
- `hydrate_params` similarly would want rethinking
- There's a code duplication issue we'll want to figure out: some behavior that is common to section-type nodes and casebooks is now defined separately for each; some materialized tree behavior is now duplicated on the Casebook model.

## Further work

- This does not change the fact that when deleting a section, no HistoricalContentNode record is created for the object itself, only its contents
- Legacy casebook nodes are still in the DB; our methods don't generally check for them, but some do. We should probably remove them from the DB, negating the need for the checks we have. That would require migrating `ContentNode.old_casebook` from a ForeignKey field to a BigInt field, but I believe otherwise would not affect the application.
- With them gone, probably ALL content nodes should be updated to have a `resource_type` with allowed values `{'Temp', 'Section', 'Link', 'LegalDocument', 'TextBlock')`, probably enumerated via Django `choices` (like annotation types are), and the column should probably be renamed to `type` or `kind`...and then we can depend on that field.
- Followup on the docstring in `ContentNodeQueryset`, "Fetching all resources normally will take a linear number of queries -- each c.resource hits the DB:"

## Oh, and.....

While I was in there, I deleted some old references to `Case` objects that I saw, stuff inside `ContentNode` that checked to see if the node was of the old proxy type `Casebook`, and updated some doc strings.


holtchesley:
> ....Down with proxies and casting?
Down!

> Do we like the direction that this is going, or, having seen it, are we missing the elegance of the proxy models?

I do, but I'm certainly open to not if you think we're losing something important. I think the failure modes are clearer without the proxy models, at the very least (which doesn't speak directly to their elegance, granted). 

 > if we lose the proxy models and casting, then we should also lose the URL converters that, when casting was working, ensured you couldn't access a section via a resource URL, or a resource via a section URL, and we should put some other guard in place for any route currently using section_param or resource_param (probably new route decorator) OR make sure that all routes that handle content nodes can handle content nodes of any variety

I'd be happy putting a bit more login in the url/path converters to do this. I think we could also get away with calling everything a 'section' in the urls by default and it would get very little notice (unless it broke links). 

 > hydrate_params similarly would want rethinking

I like the idea of pushing more of this logic into the url/path converters. Perhaps make a CasebookConverter that produces a casebook, and a ChildNodeConverter that just gives us a lambda for filtering down `casebook.contents`? 

 > There's a code duplication issue we'll want to figure out: some behavior that is common to section-type nodes and casebooks is now defined separately for each; some materialized tree behavior is now duplicated on the Casebook model.

There are slight differences between each of the functions for Casebooks and Sections. It might be nice to have them consolidated more than they currently are, which I think is possible.

> Further work
 > This does not change the fact that when deleting a section, no HistoricalContentNode record is created for the object itself, only its contents.

✔️ I'm okay with this being future work.

 > Legacy casebook nodes are still in the DB; our methods don't generally check for them, but some do. We should probably remove them from the DB, negating the need for the checks we have. That would require migrating ContentNode.old_casebook from a ForeignKey field to a BigInt field, but I believe otherwise would not affect the application.

I agree, I think that's all it will take. As is, it's only really used in hydrate_params right now (to make sure old urls don't break). 

 > With them gone, probably ALL content nodes should be updated to have a resource_type with allowed values {'Temp', 'Section', 'Link', 'LegalDocument', 'TextBlock'), probably enumerated via Django choices (like annotation types are), and the column should probably be renamed to type or kind...and then we can depend on that field.

💯 

 > Followup on the docstring in ContentNodeQueryset, "Fetching all resources normally will take a linear number of queries -- each c.resource hits the DB:"

✔️ 

> While I was in there, I deleted some old references to Case objects that I saw, stuff inside ContentNode that checked to see if the node was of the old proxy type Casebook, and updated some doc strings.

🙌

PR #1465, opened 2021-11-29
Fab task for node export

holtchesley:


holtchesley:
Do we want to add any other boilerplate to this to be common across explorations?

web/fabfile.py
```diff
@@ -305,4 +305,33 @@ def casebook_garbage_collect(older_than_days=180, dry_run=False):
...
+    content_node = None
+    if node_id:
+        content_node_id = int(node_id)
+        content_node = ContentNode.objects.get(id=content_node_id)
+    else:
+        casebook_id = int(casebook_id)
+        ords = list(map(int, ordinals.split(".")))
+        content_node = ContentNode.objects.get(casebook_id=casebook_id, ordinals=ords)
+    if not content_node:
+        print(F"Couldn't find content node with node_id={node_id} or casebook_id={casebook_id} and ordinals={ordinals}")
```
bensteinberg: I did not know f-strings could start with a capital F.

holtchesley: Hah! They can, but I've fixed this so they're uniform.

web/fabfile.py
```diff
@@ -305,4 +305,33 @@ def casebook_garbage_collect(older_than_days=180, dry_run=False):
...
+    from main.models import ContentNode
+    content_node = None
+    if node_id:
+        content_node_id = int(node_id)
+        content_node = ContentNode.objects.get(id=content_node_id)
+    else:
+        casebook_id = int(casebook_id)
+        ords = list(map(int, ordinals.split(".")))
+        content_node = ContentNode.objects.get(casebook_id=casebook_id, ordinals=ords)
+    if not content_node:
```
rebeccacremona: `ContentNode.objects.get` will raise `main.models.ContentNode.DoesNotExist` if no such node exists, so I think you want a `try`/`except` here for error handling rather than `content_node = None` and `if not content_node`

holtchesley: Changed.

rebeccacremona:
> Do we want to add any other boilerplate to this to be common across explorations?

LGTM!

Thoughts on extras: 
- could possibly be worth it to add more errors re: supplying the right args... but I mean really, if we use this simple thing wrong in our own terminals, duh, shame on us :-)
- could be interesting to have it print the file size produced
- could be interesting to have it print something about memory usage, ala [`tracemalloc` or sampling](https://medium.com/survata-engineering-blog/monitoring-memory-usage-of-a-running-python-program-49f027e3d1ba), but that might be more work than it's worth....
- could be interesting to include the docx diff utils from the test suite, to optionally compare against another file.... but, of course we can always do that manually if we want

But obviously none of that is essential. Would gladly use it as is!

holtchesley:
- [ ] could possibly be worth it to add more errors re: supplying the right args... but I mean really, if we use this simple thing wrong in our own terminals, duh, shame on us :-)
> I think this is probably a good enough start for error checking since it's just for us. 
- [x] could be interesting to have it print the file size produced
- [x] could be interesting to have it print something about memory usage, ala tracemalloc or sampling, but that might be more work than it's worth....
- [ ] could be interesting to include the docx diff utils from the test suite, to optionally compare against another file.... but, of course we can always do that manually if we want
> I think another fab task, or maybe a visual differ might be useful for this?

ChefAndy:
> * [ ]  could possibly be worth it to add more errors re: supplying the right args... but I mean really, if we use this simple thing wrong in our own terminals, duh, shame on us :-)
> 
> 
> > I think this is probably a good enough start for error checking since it's just for us.
> 
>     * [x]  could be interesting to have it print the file size produced
> 
>     * [x]  could be interesting to have it print something about memory usage, ala tracemalloc or sampling, but that might be more work than it's worth....
> 
>     * [ ]  could be interesting to include the docx diff utils from the test suite, to optionally compare against another file.... but, of course we can always do that manually if we want
> 
> 
> > I think another fab task, or maybe a visual differ might be useful for this?

the only other 'at a glance' thing I could think of along these lines would be a node count of the docx doc (did my change lose something? add way more than I thought?,) but we could always just merge this and someone could throw in a quick PR if they add something they think might be useful to others.

Issue #1459, opened 2021-11-29
Create an endpoint for alternate python-docx export experiments


holtchesley:
Create a django-superuser only endpoint that we can hit to export a high level structure for docx with TOC and sections with lore-ipsum-like filler instead of content.

PR #1467, opened 2021-11-29
Experimental export with py-docx: casebook structure

rebeccacremona:
Set `FORCE_EXPERIMENTAL_EXPORT = True` in `settings.py` to make the normal in-browser export button for casebooks use the experimental logic. Or, from the shell, call `casebook.export(experimental=True)`.

This produces more of a proof-of-concept than anything polished.

Presently included:
- a placeholder for the instructive [preamble](https://github.com/harvard-lil/h2o/blob/develop/web/main/templates/export/about.html)
- a title page, bottom- and right-aligned for fun (odd page)
- a placeholder copyright statement on the title page's reverse
- "Chapters" (top-level sections, or top-level TextBlocks) that begin on odd pages, with large titles, with text that starts mid-page for fun (as do seemingly most of the books I took off my shelf at random).
- Otherwise, sections begin on their own pages; subsections and resources do not.
- The body of resources where `has_body` is truthy is dumped in, text-only, as a placeholder. (Justified, auto-hyphenation off). 

Example output: 
[Torts!.docx](https://github.com/harvard-lil/h2o/files/7621750/Torts.docx)

You can recompile the styles as in the [README](https://github.com/harvard-lil/h2o/blob/develop/services/pandoc/README.md), subbing in `template_docx` for `reference_docx`:
```
$ cd services/pandoc/template_docx/
$ zip -r ../template.docx *
```


rebeccacremona:
Oh, will also expose to superusers in a view without the feature flag on, as per https://github.com/harvard-lil/h2o/issues/1459.

rebeccacremona:
Okay, so now you have all the options for playing with this: a feature flag, an arg for the command line, and a super-admin only button+route :-)

image

When `FORCE_EXPERIMENTAL_EXPORT` is `False`, the "experimental export" button will appear for superadmins only, for now on casebook pages only. Superadmins may also directly curl URLs of the form `/casebooks/{ID}/export.docx?annotations={true|false}&experimental=true`.

PR #1473, opened 2021-12-02
Include TinyMCE icons after update

holtchesley:


rebeccacremona:
Ah!!!!! I saw my icons were missing locally and I figured it was something I flubbed with `fab run_frontend`, had not investigated yet..... Am really pleased something was legitimately busted lol

PR #1475, opened 2021-12-06
Optionally export casebooks using AWS Lambda

rebeccacremona:
(These commits probably want to be squashed during the merge. I'm leaving them in the PR in case anyone is curious.)

This PR adds an admin-only button for exporting whole casebooks or individual sections or resources via an AWS lambda function, instead via the pandoc executable installed on the servers. It also adds a feature flag that can direct all exports to the lambda.

The lambda is already deployed to AWS.

For more details on local setup, see the included [README](https://github.com/harvard-lil/h2o/pull/1475/files#diff-a419bc23c8219f4993d2934d30b4fb0a6b895133acec64746e79ffc8de38f50f)

Straightforward next steps would include:
- saving the export back to the bucket for use in a simple caching setup (if `obj.export_cache_valid` and `file_in_bucket()`, serve file, else export and cache; invalidate cache on edit; clean up bucket every so often)
- add a simple export job table and make async; lambda triggered by s3 object upload, saves back relic; UI polls for completion of job and shows a spinner while waiting


rebeccacremona:
~Note to self: don't forget to make sure these files are private, not public, if you try to access straight from the bucket instead of via the app.~
Done. (Thanks, Ben!)

docker/pandoc-lambda/function/app.py
```diff
@@ -0,0 +1,42 @@
...
+        s3 = boto3.resource('s3',
+          endpoint_url=os.environ['S3_ENDPOINT_URL'],
+          aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
+          aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
+        )
+        s3.Bucket(os.environ['EXPORT_BUCKET']).download_fileobj(input_s3_key, pandoc_in)
+        pandoc_in.seek(0)
+
+        # convert to docx with pandoc
+        html = '<p>hello world</p>'
```
bensteinberg: This is probably a relic of experimenting?

web/main/utils.py
```diff
@@ -514,3 +530,231 @@ def get_link_title(url):
...
+                    Payload=bytes(json.dumps({"filename": filename, "is_casebook": export_type == 'Casebook'}), 'utf-8')
+                )
+                response = {
+                    'status_code': raw_response['ResponseMetadata']['HTTPStatusCode'],
+                    'headers': raw_response['ResponseMetadata']['HTTPHeaders'],
+                    'content': raw_response['Payload'],
+                    'get_text': lambda: raw_response['Payload'].read()
+                }
+                lambda_log_str = str(base64.b64decode(raw_response['LogResult']), 'utf-8').strip().replace('\n', '; ').replace('\t', ', ')
+                logger.info(f"Exporting Casebook 4227: Lambda logs \"{lambda_log_str}\"")
```
rebeccacremona: LOL: yes, I intended to log "Exporting Casebook 4227" every time, no matter what we are exporting...... Will fix.

web/frontend/legacy/lib/ui/content/export_modal.js
```diff
@@ -6,12 +6,13 @@ import urls from 'libs/urls';
 delegate(document, 'a.action.export', 'click', (e) => {
   e.preventDefault();
   const experimental = e.target.classList.contains('export-experimental');
+  const aws_lambda = e.target.classList.contains('export-lambda');
```
holtchesley, line 9: I always narrate this practice of javascript with a "tuck tuck".

If you wanted to inject as a javascript variable, I'm doing a couple of those at https://github.com/harvard-lil/h2o/blob/a4885ccf918c0deca3bf0f217eb893eebcbf77e8/web/main/templates/base.html#L14 . It's probably just fine to keep it here though.

bensteinberg: Is "tuck tuck" like "tsk tsk" or is it something else?

holtchesley: Like you're tucking the data into the dom.

rebeccacremona: Ah, that is nice and tidy! I like it! A thought: this whole file is full of class list / data attribute retrievals. Would it be better to refactor the whole file at once.... maybe even porting it out of our legacy JS and into Vue? Happy to change now if you prefer.

holtchesley: I think it makes sense to do it on another pass, perhaps when we add options to the export process.

rebeccacremona: Cool. Thanks for reviewing! FYI, I am about to push one last commit, a tweak to setting handling subsequent to a convo with Ben.

rebeccacremona: (pushed!)

Issue #1427, opened 2021-10-13
Make an ideal docx example export


holtchesley:
Use the examples from #1422 to craft a docx that can serve as a template for exports.

holtchesley:
- Roman numerals for non-numbered sections preceding numbered sections (Preface, Foreward, Introduction). 
- Casebook Headnote becomes block (prior to TOC) or preface?
- Footnotes, no endnotes.

rebeccacremona:
More verbiage, from Slack:
> 1) Make us a gorgeous DOCX casebook (https://github.com/harvard-lil/h2o/issues/1427), with all the bells and whistles, informed by aesthetics, printing conventions, Jordi's feedback about our DOCX export and the PDF of JZ's book, and maybe https://github.com/harvard-lil/h2o/issues/1422, using the Microsoft Word GUI.
>
> 2) Then, changing https://github.com/harvard-lil/h2o/blob/develop/services/pandoc/reference_docx/word/styles.xml until the styles are present that are necessary for said gorgeous casebook. And, learning how to add all the other features and change all the necessary settings using a combo of python-docx (where supported) and plain old etree/lxml (per CAPnow example, and etc.). https://github.com/harvard-lil/h2o/pull/1467 may be useful too.
>
> These will help us make an objectively better exported casebook docx.
>
> There's a hope that this will speed up the docx -> indesign process for MIT press books, with the assistance of macros.

Link to the thread in Slack re: feedback https://hlslil.slack.com/archives/C4FE9FL56/p1635180657021800

ChefAndy:
## Andy's Checklist

- **Visuals**
  - [x] Get basic headline and body text style, visual hierarchy, letter/line spacing, space relationships styles squared away
  - [x] Make the TOC more readable
  - [x] Restyle introductory page
  - [x] Get footers set up
      - [x] Styled? 
      - [x] Final Layout
      - [x] Different Visual Presentations hooked up to different tracks
          - [x] put together a couple of style options
          - [x] styling elements like font weights, etc.
  - [x] Headers
      - [x] Dynamic?
      - [x] Styled? 
      - [x] Change style with sections (e.g. _i, ii, iii_ for frontmatter and 1, 2, 3 for content) 
      - [x] Final Layout 
- **Structure**
  - [x] Set up the style cascades to make customization easier  
example) when the template is applied in word. 
  - [x] Apply the template to a few sample case books to make sure the basics work
[...]

PR #1478, opened 2021-12-16
Export Style Updates

ChefAndy:
Documentation to come (plenty of inline docs, but they need to be updated a little.)

More detailed progress in issue #1427


holtchesley:
Pretty sure all this needs to pass tests is a `docker-compose exec web pytest -k test_export --write-files` (which sets the comparison files we use for regression checks to whatever the code currently produces).

ChefAndy:
Working on docs/reorg and submitting in a subsequent PR.

Issue #1443, opened 2021-10-28
New section reports "this casebook is empty"


rebeccacremona:
Probably just wants s/casebook/section/, if showing a section?

- Add a new section
- Click into it
- Scroll beneath the title/subtitle, headnote fields
![image](https://user-images.githubusercontent.com/11020492/139277203-73c61417-7f2b-4386-8f1c-e6bcc82a2776.png)

lizadaly:
Looks like this was done:

<img width="866" alt="image" src="https://user-images.githubusercontent.com/19571/178070399-e54b46c5-62c3-434b-bc12-c108fdfc9007.png">

Issue #1442, opened 2021-10-26
Empty search results on first try; query string weirdness


rebeccacremona:
1. Visit https://opencasebook.org/search/
2. Enter a casebook title keyword (e.g. torts) or author's surname (e.g lessig)
3. Hit enter/click submit. "No results found", URL includes `school=None` (e.g. `https://opencasebook.org/search/?type=casebooks&q=torts&sort=display_name&author=&school=None` or `https://opencasebook.org/search/?type=casebooks&q=lessig&sort=display_name&author=&school=None`)
4. Hit enter/click submit again, without changing anything. Results found, URL ends at `&school=` (e.g. `https://opencasebook.org/search/?type=casebooks&q=torts&sort=display_name&author=&school=` or `https://opencasebook.org/search/?type=casebooks&q=lessig&sort=display_name&author=&school=`)

dongkyunc:
This is (probably) happening because certain casebooks entries have incorrect properties.

The behavior is likely as follows:
1. The search URL is visited, with no query parameters. No filters are thus provided and so all the casebooks are returned. Some of these do not have an appropriate affiliation fields or no value at all and thus, [`{% if request.GET.school == school %}selected {% endif %}`](https://github.com/harvard-lil/h2o/blob/4e0ad9597fb2860296da060195528e8cd79391e7/web/main/templates/search/filters.html#L24) is firing. The screenshots below confirm that.
2. That filter value is used for the next search, but coerced as the string `"None"`, which would match no casebook entry, thus returns no results.
3. Since now the offending malformed entry is no longer being returned, the `"None"` option can no longer be selected and this query returns as expected.

A quick solution would be to change the linked line to `if request.GET.school and request.GET.school == school`. 

However, there is a deeper issue: The values in the database are occasionally junk. [Three casebooks ](https://opencasebook.org/search/?type=casebooks&q=&sort=display_name&author=&school=,%20) are affiliated with the `,` school and multiple other schools have multiple entries with slight variations in either spelling or formatting. I don't have much visibility into how the affiliations are created so I'm not sure what possible solutions there would look like. 

<img width="535" alt="image" src="https://user-images.githubusercontent.com/8096144/143539965-9ab27a6d-9489-484a-b971-75def6ef6ae7.png">
<img width="416" alt="image" src="https://user-images.githubusercontent.com/8096144/143540210-b963b05a-b581-454a-82ac-c97e6362c503.png">

lizadaly:
[...]

Issue #1439, opened 2021-10-21
Right-aligned text in some footnotes of a CAP case


rebeccacremona:
https://opencasebook.org/documents/5077/

Check out these funny-looking, right-aligned footnotes:

![image](https://user-images.githubusercontent.com/11020492/138330657-4185b437-4d13-4d13-a02b-29564a3c913d.png)

![image](https://user-images.githubusercontent.com/11020492/138330953-590e3367-cf85-4455-921b-60d9430036a8.png)

Example markup (looks fine)
```
<aside id="footnote_1_22" class="footnote" data-label="2">
  <a href="#ref_footnote_1_22">2</a>
  <p id="p-160">
    <em>See</em> <em>Bubolz v. Dane Cty.</em>, <a class="citation" href="https://cite.case.law/wis-2d/159/284/#p294" data-index="127" data-case-ids="8667287">159 Wis. 2d 284</a>, 295-96, <a class="citation" href="https://cite.case.law/wis-2d/159/284/#p294" data-index="128" data-case-ids="8667287">464 N.W.2d 67</a> (Ct. App. 1990) (injunction against violation of restrictive covenant did not prohibit a residents from engaging in business activities that are "incidental to their occupation of the premises as their single family residence."); <em>see also</em> <em>Joyce v. Conway</em>, <a class="citation" href="https://cite.case.law/wis-2d/7/247/#p251" data-index="129" data-case-ids="8677350">7 Wis. 2d 247</a>, 251, <a class="citation" href="https://cite.case.law/citations/?q=96%20N.W.2d%20530" data-index="130">96 N.W.2d 530</a> (1959) (explaining that acquiescence to past violations does not deprive affected property owners of the right to enforce later violations of a restrictive covenant).
  </p>
</aside>
```
(I also note that the last 3 notes on the page don't have matching footnotemarks; this looks to be an issue in [CAP](https://cite.case.law/nw2d/914/643/#footnote_1_23) too, so likely unrelated/not an H2O issue)

dongkyunc:
[Looks like the selector is a bit broad here.](https://github.com/harvard-lil/h2o/blob/develop/web/frontend/styles/cases.scss#L55-L57) Changing the selector from `aside.footnote a` to `aside.footnote > a` so that it only catches the anchor pointing back to the body text and not the links in the footnote text itself should do the trick. 

[...]

PR #1479, opened 2021-12-16
Assorted small improvements

holtchesley:
Addresses #1443, #1442, #1439


rebeccacremona:
Yay! These are delightful fixes.

PR #1480, opened 2021-12-16
Footnote Postprocessing

holtchesley:


docker/pandoc-lambda/function/app.py
```diff
@@ -1,14 +1,133 @@
...
+def promote_case_footnotes(doc):
+    val_att = '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}val'
+    author_footnotes = {}
+
+    for ref in doc.element.xpath("//*[*/w:rStyle[starts-with(@w:val,'FootnoteReference')]]"):
+        mark_text = ref.text
+        style_node = ref.xpath(".//w:rStyle[starts-with(@w:val,'FootnoteReference')]")[0]
+        node_id = style_node.attrib[val_att][18:]
+        mark_id = f"{node_id}-{mark_text}"
+        if mark_id not in author_footnotes:
```
rebeccacremona, line 39: Q: if we encounter `mark_id in author_footnotes`, does that mean there's an infelicity that should be addressed by the author (e.g. choosing a non-duplicative footnote mark), by us (adapting the html code so it doesn't produce duplicative footnote marks), etc. or is this a benign situation? Should we figure out a way to log duplicative notes that are skipped via this branch?

docker/pandoc-lambda/function/app.py
```diff
@@ -1,14 +1,133 @@
...
+            author_footnotes[mark_id] = {'id':mark_id, 'mark': mark_text, 'refs': [], 'texts': []}
+        style_node.attrib[val_att] = 'FootnoteReference'
+        author_footnotes[mark_id]['refs'].append(ref)
+
+    for text_el in doc.element.xpath("//w:p[w:pPr/w:pStyle[starts-with(@w:val,'FootnoteText')]]"):
+        mark_text = text_el.xpath(".//*[*/w:rStyle]/w:t")[0].text
+        node_id = text_el.xpath(".//w:pStyle[starts-with(@w:val, 'FootnoteText')]/@w:val")[0][13:]
+        mark_id = f'{node_id}-{mark_text}'
+        if mark_id not in author_footnotes:
+            author_footnotes[mark_id] = {'id':mark_id, 'mark': mark_text, 'refs': [], 'texts': []}
```
rebeccacremona, line 49: Testing my understanding: is this accurate? `if mark_id not in author_footnotes`+ the following dict entry appear here to handle "orphaned" footnotes that don't have a footnotemark somewhere in the text pointing to them?

docker/pandoc-lambda/function/app.py
```diff
@@ -1,14 +1,133 @@
...
+
+
+    case_footnotes = {}
+
+    for ref in doc.element.xpath("//*[*/w:rStyle[starts-with(@w:val,'CaseFootnoteReference')]]"):
+        mark_text = ref.text
+        node = ref.xpath(".//w:rStyle[starts-with(@w:val,'CaseFootnoteReference')]")[0]
+        node_id = node.attrib[val_att][22:]
+        mark_id = f'{node_id}-{mark_text}'
+        if mark_id not in case_footnotes:
```
rebeccacremona, line 66: (Same question: if `mark_id in author_footnotes` is pathological, maybe we should log when they go by.)

docker/pandoc-lambda/function/app.py
```diff
@@ -1,14 +1,133 @@
...
+        node.attrib[val_att] = "FootnoteReference"
+        parent = ref.getparent()
+        gp = parent.getparent()
+        gp.replace(parent, ref)
+        case_footnotes[mark_id]['refs'].append(ref)
+
+    for footnote_start in doc.element.xpath("//*[*/*[starts-with(@w:val,'CaseFootnoteText')] and .//w:hyperlink]"):
+        mark_text = footnote_start.xpath(".//*[*/w:rStyle]/w:t")[0].text
+        node_id = footnote_start.xpath(".//w:pStyle[starts-with(@w:val, 'CaseFootnoteText')]/@w:val")[0][17:]
+
```
rebeccacremona, line 77: Testing my understanding: is the `current_stack` and corresponding `while` stanza to handle footnotes with multiple paragraphs of text?

I think this stanza might benefit from a comment stating its goal, since it's a little dense to read through.

docker/pandoc-lambda/function/app.py
holtchesley: This could happen in the case that an author has annotated their own text (possibly cloned), and be benign. We still capture+output the note if it's a duplicate. Logging might be a good idea, or I've been thinking about adding in a fab task for bulk analysis when caching is implemented.

docker/pandoc-lambda/function/app.py
holtchesley: Yup, it will have entries in the `texts` key, but not the `refs` key.

docker/pandoc-lambda/function/app.py
holtchesley: This will happen when someone elides a footnote mark in the text, but not the footnote text (which is probably unintended, but expected. Minimizing these would be one of the advantages of a programmatic footnote Ellison option on the UI.

docker/pandoc-lambda/function/app.py
```diff
@@ -52,7 +171,19 @@ def handler(event, context):
                         sig_string = f"unknown signal {-e.returncode}"
                 else:
                     sig_string = f"non-zero exit status {e.returncode}"
-                raise Exception(f"Pandoc command exited with {sig_string}")
+                ss = "Pandoc command exited with " + str(sig_string)
+                print(ss)
+                pandoc_out.seek(0,0)
+                print(response.stderr)
+                raise Exception(ss)
```
rebeccacremona, line 178: Do you still want the prints and seek here? Maybe debugging stuff that got left in by accident?

Do you happen to remember if `print(response.stderr)` ever came in handy here? I would have expected the process to exit before reaching this because of `if response.stderr: raise Exception(`, but I've found this error handling to be somewhat finicky......

docker/pandoc-lambda/function/app.py
```diff
@@ -52,7 +171,19 @@ def handler(event, context):
...
             if not os.path.getsize(pandoc_out.name) > 0:
                 raise Exception(f"Pandoc produced no output.")
+
+            if options.get('word_footnotes', False):
+                doc = Document(pandoc_out)
+                promote_case_footnotes(doc)
+                output = io.BytesIO()
```
rebeccacremona, line 185: I've been interested in whether we want to be working with temp files or with bytes IO objects during this whole project. If we've made it this far in the process and not run out of RAM, then surely we have enough to use a bytes IO here. We get charged the same, regardless of whether we actually USE all the RAM we have allocated, so it's not a win in that way to write to disk. 

Would you be willing to comment if you picked bytes IO for speed, convenience, etc., or if you use it as your go to (over, say, a spooling temp file), or what have you?

docker/pandoc-lambda/function/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with python 3.9
+# This file is autogenerated by pip-compile
```
rebeccacremona, line 2: Hmm weird, how did we manage to run this with different versions of pip/pip-tools...... Oh well.

web/main/utils.py
```diff
@@ -531,8 +543,122 @@ def get_link_title(url):
...
+        for ref,texts in zip(val['refs'], val['texts']):
+            fid += 1
+            lift_footnote(doc, footnote_part, ref, texts, f"{fid}", author=True)
+
+    for val in case_footnotes.values():
+        for ref,texts in zip(val['refs'], val['texts']):
+            fid += 1
+            lift_footnote(doc, footnote_part, ref, texts, f"{fid}", author=False)
+    footnote_part._blob = b'<?xml version="1.0" encoding="UTF-8" standalone="yes"?>' + etree.tostring(footnote_part.element)
+    print(footnote_part._blob.decode("utf-8"))
```
rebeccacremona, line 653: Do you still want this print line here?

docker/pandoc-lambda/function/app.py
holtchesley: You're right, and I'll add some commentary for future perusers.

docker/pandoc-lambda/function/app.py
holtchesley: Yeah, these lines were debugging artifacts and I'll remove them. The print did not produce anything.

docker/pandoc-lambda/function/app.py
rebeccacremona: Ah, gotcha!

docker/pandoc-lambda/function/app.py
holtchesley: My general thought was that pandoc would have just freed up all the RAM it was using, so there'd be plenty of space free for the taking. I suspect it's marginally faster, though probably not enough for anyone to notice. I think this was probably more of a desire to not open up a second temp file, or go through the trouble of erasing the first to overwrite it.

docker/pandoc-lambda/function/app.py
rebeccacremona: > We still capture+output the note if it's a duplicate.
Ah! I was misreading, yes, now I see. Thanks!

Love the idea of using the cache, once implemented, for bulk analysis of varying kinds!

docker/pandoc-lambda/function/app.py
rebeccacremona: 👍 thx!

PR #1482, opened 2021-12-17
Button for Styled Export

holtchesley:


web/main/templates/includes/action_buttons.html
```diff
@@ -48,15 +48,23 @@
...
-      <a class="action one-line export export-experimental export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">Experimental Export</a>
+      <a class="action one-line export export-experimental export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">DocXport (TOC Only)</a>
+
     {% endif %}
-    {% if request.user.is_superuser and not FORCE_AWS_LAMBDA_EXPORT %}
+      {% if request.user.is_superuser%}
+      {% if not FORCE_AWS_LAMBDA_EXPORT %}
       <a class="action one-line export export-lambda export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">AWS Lambda Export</a>
+      {% else %}
+      <a class="action one-line export export-docx-footnotes export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">Styled Export</a>
```
holtchesley, line 63: Styled Export for Andy's nice styles and footnotes.

web/main/templates/includes/action_buttons.html
```diff
@@ -48,15 +48,23 @@
     </form>
       {% endif %}
   {% endif %}
+    
   {% if exportable %}
     <a class="action one-line export export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">Export</a>
+
     {% if not resource and not section and request.user.is_superuser and not FORCE_EXPERIMENTAL_EXPORT %}
-      <a class="action one-line export export-experimental export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">Experimental Export</a>
+      <a class="action one-line export export-experimental export-{{ section.is_annotated|yesno:"has,no" }}-annotations" role="button" rel="nofollow" href="#">DocXport (TOC Only)</a>
```
holtchesley, line 56: Renamed the python-docx export to `DocXport (TOC Only)` since there are a few experiments going.

web/main/views.py
```diff
@@ -2154,15 +2154,15 @@ def export(request, node, file_type='docx'):
     """
     if file_type not in ('docx', 'html', 'json'):
         raise Http404
-
+    experimental, aws_lambda, docx_footnotes = [request.GET.get(var) == 'true' if var in request.GET else default
+                                                for var, default in
+                                                [('experimental', settings.FORCE_EXPERIMENTAL_EXPORT),
+                                                ('aws_lambda', settings.FORCE_AWS_LAMBDA_EXPORT),
+                                                ('docx_footnotes', settings.FORCE_DOCX_FOOTNOTES)]]
```
rebeccacremona, line 2161: nice

rebeccacremona: true, this gives any user access to which ever kind of export they want, if they are clever enough to manipulate the query strings? probably harmless.

web/main/models.py
```diff
@@ -2153,7 +2153,7 @@ def export(self, include_annotations, file_type='docx', export_options=None, is_
         if file_type == 'html':
             return html
 
-        if aws_lambda or settings.FORCE_AWS_LAMBDA_EXPORT:
+        if aws_lambda:
```
rebeccacremona, line 2160: Okay! To me, this change means the setting doesn't actually force aws lambda export, it just toggles the default behavior of the view, but if you prefer it this way, fine by me, especially since we are probably ripping out the non lambda code quite soon!

rebeccacremona: Oh, though it -may- affect which version of the export code is run by various tests; let's see.

rebeccacremona: Yeah, it does.
```
INFO     main.models:models.py:3662 Exporting Casebook 1: serializing to HTML
INFO     main.utils:utils.py:665 Exporting Casebook 1: launching pandoc subprocess
```

holtchesley: Yeah, this could be de-obfuscated. I can switch this to `DEFAULT_TO_AWS_LAMBDA_EXPORT` (or the like). If we want to force the lambda export, I could add a bit of code to also allow for selecting the template docx.

web/main/models.py
```diff
@@ -2112,7 +2112,7 @@ def type(self):
         else:
             return 'resource'
 
-    def export(self, include_annotations, file_type='docx', export_options=None, is_child=False, experimental=False, aws_lambda=False, docx_footnotes=False):
+    def export(self, include_annotations, file_type='docx', export_options=None, is_child=False, experimental=settings.FORCE_EXPERIMENTAL_EXPORT, aws_lambda=settings.FORCE_AWS_LAMBDA_EXPORT, docx_footnotes=settings.FORCE_DOCX_FOOTNOTES):
```
holtchesley: If using them as default arguments this way isn't offensive, I can switch them all from `FORCE` to `DEFAULT`.

rebeccacremona: Interesting! I am a little anxious about it because of https://stackoverflow.com/questions/11238514/why-is-djangos-settings-object-a-lazyobject, but I'm not sure if that's in play here..... 

If it works, totally down with that approach!

holtchesley: Haha! Best to avoid that just in case.

web/main/models.py
```diff
@@ -2124,6 +2124,10 @@ def export(self, include_annotations, file_type='docx', export_options=None, is_
             ...     file_data = full_casebook.export(include_annotations=True)
         """
         # prefetch all child nodes and related data
+        experimental = experimental if experimental is not None else settings.FORCE_EXPERIMENTAL_EXPORT
+        aws_lambda = aws_lambda if aws_lambda is not None else settings.FORCE_AWS_LAMBDA_EXPORT
+        docx_footnotes = docx_footnotes if aws_lambda is not None else settings.FORCE_DOCX_FOOTNOTES
+
```
holtchesley, line 2130: A little bit more safety, then.

PR #1490, opened 2022-01-10
Allow for disabling all exports

holtchesley:


holtchesley:
I have somewhat mixed opinions about throwing up a singleton like this. It does allow us to quickly toggle exporting on and off through the adminterface, and we can add an endpoint if we want automate a shutoff (say for billing issues). Any thoughts?

PR #1491, opened 2022-01-13
S t y l e 

ChefAndy:
New doc output styling & strucutre. It includes updated styles and layouts including odd/even pages, dynamic headers, page numbering with appropriate front matter styling, columned case text and narrower headnote paragraphs for a less fatiguing read, more durable style blocks, updated text choices, odd-page chapters, and more. This involved updating the reference docx, adding a new function in the lambda app to place these things, changing some of our output html, and a few other things.

Just need to resolve some merge conflicts and likely work the tests a bit.


rebeccacremona:
(Just finished reading this diff, and wanted to say before the long weekend: so well-organized and easy to read!)

rebeccacremona:
Exciting to see this getting so close! Since I am an idiot and didn't realize you have to start a review from the "files changes" tab ~this is presently a draft PR, I can't use the actual Github "review" mechanism, so~ please forgive if instead there is a barrage of comments. 

In this comment, general remarks!

1) Quick thing, looks like flake8 doesn't like the indentation of `docker/pandoc-lambda/function/app.py:252-255`; with that fixed up, the tests should run.

2) A bump re: integration with Brennan's enhanced footnote code. Locally, if I spin up and run as-is, and export "Torts!", export fails with:
```
AWS Lambda export failed: x-amz-function-error: None, content-type:text/plain; charset=utf-8, {"errorMessage": "object of type 'NoneType' has no len()", "errorType": "TypeError", "requestId": "4d9cbbbd-3c29-4a99-94c0-8b6906c76481", "stackTrace": ["  File \"/function/app.py\", line 349, in handler\n    promote_case_footnotes(doc)\n", "  File \"/function/app.py\", line 75, in promote_case_footnotes\n    if len(gp) > 0: # 'if gp' triggers a warning and I saw a similar test behave unpredictably\n"]}
```
Brennan has `if gp` there; your diff says, ` if len(gp) > 0: # 'if gp' triggers a warning and I saw a similar test behave unpredictably`. If I switch it back to `if gp` and then `docker-compose restart pandoc-lambda`, export works; I don't see a warning, but I'm not sure where to look for one. What warning did you see, and can you still trigger it? Do you know what test was behaving unpredictably? I note that locally, the tests pass: that means this situation is not well-covered by our test suite. (See https://github.com/harvard-lil/h2o/issues/1493).

This also makes me wonder if perhaps you have been running with `FORCE_DOCX_FOOTNOTES = False` locally? (Which is how things were by default until [recently](https://github.com/harvard-lil/h2o/pull/1485/files#diff-d7c867e0f418251ebe176b2353ce9d465609bce54643a4fc603d175d01194857R260). Have the fancy footnotes gotten a makeover too, or will that be a TODO for a future pass?

FWIW: all the code that make fancy footnotes optional is probably getting removed shortly.

3) It would be kind of nice to deploy this side-by-side with our existing styles, so that we can more easily compare the DOCX produced by the existing code and the DOCX produced by your code, both to test timing and to check for issues, either using prod or on stage. This isn't set up that way presently, but since we have (variously old) dumps of prod locally, we can pull down and test locally before merging.

But I can tell you were thinking about setting it up that way, by adding the `DOCX_SECTIONS = True` setting, which toggles off SOME of your stuff. I also see from the added trailing comma in `def export_via_aws_lambda(obj, html, file_type, docx_footnotes=None,)` that you were thinking about making `docx_sections` an arg, likely in anticipation of adding a view option/button.

So my questions are three:
- Is the output with `DOCX_SECTIONS = False` viable/interesting/acceptable to you? If yes, let's go whole hog and offer it to admins as an export option for now, ala the buttons we were recently using. (Can send code links/more details if desired/useful.) If not, let's lose that setting, the trailing comma, and the export option.

- What's your appetite for wrapping this all behind an option and exposing the legacy-styled export to admins via a button? And a setting to toggle: yours by default, legacy via special button; flip the setting to be legacy by default, yours by special button? Any appetite? Is that too much work / too risk averse / not useful enough? What are your thoughts?

- What's the best way for us to set up a side-by-side comparison? Since I have seen at least one instance of missing content (which I will describe later) and remember seeing a scrambled table go by during your demo on Friday, I think we should cross-check some small number of casebooks. I was thinking of setting up a little google sheet with say 6 books, locally exporting each via legacy and new and recording the timing for each, and uploading the artifacts, and maybe three of us each spot-check two, so no one dies of boredom but we get decent coverage? Does that sound moderate? I'm happy to do that.

4) Already said, but, this is so easy to read and clear! Some questions for documentation:

- Can you introduce us at a high level to your design? What fonts are in use, and where (I think I see century gothic, arial, and baskerville, is that it)? What is the expected structure of a chapter/section/subsection/resource? For instance, `node start -> head separator-> section number -> section title -> head end -> node end -> node start -> head separator -> etc.` What are [`topper_styles`](https://github.com/harvard-lil/h2o/pull/1491/files#diff-73ad048f236dd6af122c17f52eab44f11841f8e0318451c43df3ffd7f9ba25a5R241), and what is their relationship with section breaks (and are those commented out ones there on purpose or by accident)? how much of the final doc is dependent on styles that pandoc applies (if any) versus custom ones?

- The differences between the 7x10 and 8.5x11 and templates. Are both maintained entirely separately, or do they share some files? If you make changes to one, should you make changes to the other? I see a `zsh` script for zipping up the 7x10; shall we make a utility for zipping up both? (I know only one is in use at a time for now, so maybe that feels premature, and we can add that as a component to a project that lets an exporter choose their page size). Is there anything we should know about how the styles differ between the two?

- How did you produce these files? The README still explains how I produced the originals (I had pandoc output its standard reference docx, unzipped that, committed it, and then added sections/ made changes in commented blocks). Did these start from that? 

- What's your recommended process for making further edits and changes? I gather you generally make changes to a document in the Word UI, find the resulting XML, and the port that back here? Am I right about that?

5) I'm wondering about the directory structure. Does this now want to live in the `docker/pandoc-lambda/` directory? If yes, we could make zipping up the source and producing the final reference docx part of the container's build process, so it never has to be done manually. 

I'm also wondering if one more layer of directory structure might be helpful here. That is, all the unzipped XML that gets zipped up to make a DOCX is in its own directory, say, `docker/pandoc-lambda/reference_docx/7x10/src` gets zipped into `docker/pandoc-lambda/reference_docx/7x10/reference.docx`, general README text could be added to `docker/pandoc-lambda/README.md`, other things specific to 7x10 or 8.5x11 (if there are any, like page-size specific versions of this script https://github.com/harvard-lil/h2o/pull/1491/files#diff-fd9700722644776e8851dc53d89cfc47a663e000681ce59408dee8375f9a2428) as siblings of `src` like `reference.docx`? 

Do either or both of those changes sound good to you?

6) Your compilation script includes an `archive` option, and this PR includes some archived DOCXs. a) Do we need those archived ones, or were they mostly useful while you were working? b) Is this mainly for the convenience of not having to use git history to retrieve old versions, or are there other advantages?

## Summary of general remarks
- [ ] flake8 (`docker-compose exec web flake8 ../docker/pandoc-lambda/function/`)
- [ ] `if gp`; footnotes
- [ ] feature flag removal or robustification
- [ ] side-by-side comparisons
- [ ] docs
- [ ] directory structure tweak?
- [ ] comment about or remove archived DOCXs

Coming next: questions about particular lines of code, using that UI!

web/main/templates/export/tbd.html
```diff
@@ -2,7 +2,7 @@
   <div class="SectionNumber" data-custom-style="Section Number" data-toc-idx="{{ index }}">{{ node.ordinal_string }}</div>
   <div class="SectionTitle" data-custom-style="Section Title">{{ node.title }}</div>
```
rebeccacremona: I'm guessing title, subtitle, headnote, etc. all need a `|cut:"&nbsp;"` too? Bummer we have that stuff around in our fields..... Maybe unavoidable with the fancy GUI text editor......

assets/export/docx/7x10/style_structure_reference.md
```diff
@@ -0,0 +1,164 @@
...
+  * Table  
+  * Normal Table  
+    * Plain Table 4 basedOn: TableNormal 
+    * Table Grid basedOn: TableNormal 
+* Numbering
+  * No List  
+  * Current List1  
+  * Current List2  
+  * Current List3  
+  * Current List4
```
rebeccacremona: Do you know what the story is with `Current List1-4`? Is that just a Word way to have 4 lists around?

assets/export/docx/7x10/word/fontTable.xml
```diff
@@ -0,0 +1,52 @@
...
+&#9;&#9;<w:pitch w:val="variable" />
+&#9;&#9;<w:sig w:usb0="00000003" w:usb1="00000000" w:usb2="00000000" w:usb3="00000000" w:csb0="00000001" w:csb1="00000000" />
+&#9;</w:font>
+&#9;<w:font w:name="Arial">
+&#9;&#9;<w:panose1 w:val="020B0604020202020204" />
+&#9;&#9;<w:charset w:val="00" />
+&#9;&#9;<w:family w:val="auto" />
+&#9;&#9;<w:pitch w:val="variable" />
+&#9;&#9;<w:sig w:usb0="00000003" w:usb1="00000000" w:usb2="00000000" w:usb3="00000000" w:csb0="00000001" w:csb1="00000000" />
+&#9;</w:font>
```
rebeccacremona: I'm guessing these `&#9;` shouldn't be here?

assets/export/docx/7x10/word/document.xml
```diff
@@ -0,0 +1,912 @@
...
+xmlns:w10="urn:schemas-microsoft-com:office:word"
+xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main"
+xmlns:pic="http://schemas.openxmlformats.org/drawingml/2006/picture"
+xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing">
+
+  <w:body>
+    <w:p>
+        <w:pPr>
+            <w:pStyle w:val="Head Separator" />
+<!--            <w:sectPr>
```
rebeccacremona: to remove?

assets/export/docx/7x10/word/styles.xml
```diff
@@ -0,0 +1,1952 @@
...
+    <!--
+
+
+
+    Character Styles
+
+
+
+    -->
+    <!-- TODO: Replace DefaultParagraphFont references with references to a deliberate style -->
```
rebeccacremona: Is this TODO important? Can you explain what it will do for us?

assets/export/docx/7x10/word/styles.xml
```diff
@@ -0,0 +1,1952 @@
...
+        <w:uiPriority w:val="2"/>
+        <w:qFormat/>
+        <w:pPr>
+            <w:outlineLvl w:val="8"/>
+        </w:pPr>
+    </w:style>
+
+    <!--  The corresponding subheadings are largely the same size, but de-emphasized through color and text style. I don't
+        think these can be implemented in the h2o interface, but users might find them useful. -->
+    <w:style w:type="paragraph" w:customStyle="1" w:styleId="Subheading1">
```
rebeccacremona: Can you expand on "I don't think these can be implemented in the h2o interface"?

assets/export/docx/8.5x11/word/fontTable.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<w:fonts xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
+&#9;<w:font w:name="Symbol">
+&#9;&#9;<w:panose1 w:val="02000500000000000000" />
```
rebeccacremona: I suspect these `&#9;` are stray as well?

docker/pandoc-lambda/function/app.py
```diff
@@ -123,6 +127,166 @@ def promote_case_footnotes(doc):
...
+            Number of columns in non-header paragraph text. There is no way to specify # of cols in a style
+    """
+
+    doc_w = Twips(doc_w_twips)
+    doc_h = Twips(doc_h_twips)
+    internal_margin = Twips(internal_margin_twips)
+    external_margin = Twips(external_margin_twips)
+    big_margin = Twips(big_margin_twips)
+
+    # See which rIDs pandoc gave our headers and footers. These values come
```
rebeccacremona: Looks like this comment got truncated.

docker/pandoc-lambda/function/app.py
```diff
@@ -123,6 +127,166 @@ def promote_case_footnotes(doc):
...
+            if grafs[i + 1] == 'Case Header':
+                continue
+            else:
+                section_break(doc.paragraphs[i])
+        elif p == 'Case Body':
+            if grafs[i + 1] not in topper_styles:
+                continue
+            else:
+                section_break(doc.paragraphs[i], paragraph=True)
+        #elif p.startswith()
```
rebeccacremona: to be removed?

docker/pandoc-lambda/function/app.py
```diff
@@ -123,6 +127,166 @@ def promote_case_footnotes(doc):
...
+        sec.header_distance = external_margin
+        sec.left_margin = external_margin
+        sec.orientation = WD_ORIENTATION.PORTRAIT
+        sec.page_height = doc_h
+        sec.page_width = doc_w
+        sec.right_margin = external_margin
+        sec.top_margin = big_margin
+
+        # don't need any of this stuff for just a block of paragraphs. There's defintiely a better way to differentiate
+        # who does and doesn't need headers
```
rebeccacremona: Can you elaborate what's undesirable there? You say, "There's defintiely a better way"?

ChefAndy:
> Exciting to see this getting so close! Since I am an idiot and didn't realize you have to start a review from the "files changes" tab ~this is presently a draft PR, I can't use the actual Github "review" mechanism, so~ please forgive if instead there is a barrage of comments.

Tons of great feedback. Thanks a million!


> 1. flake8 doesn't like the indentation of `docker/pandoc-lambda/function/app.py:252-255`; with that fixed up, the tests should run.

Word.
 
>     2. A bump re: integration with Brennan's enhanced footnote code. [...]  AWS Lambda export failed: [...]

Got it.

> Brennan has `if gp` there; your diff says, ` if len(gp) > 0: # 'if gp' triggers a warning [...] What warning did you see, and can you still trigger it? [...]

Running locally, I'd see this in the pandoc-lambda_1 console running under the docker-compose process. I know you run it as a daemon— I run it in the foreground in my shell. Not sure if you see those messages or not? You can recreate the warning in the repl pretty easily:

```
>>> if doc.paragraphs[1]._element.getparent():
...     print("hello")
...
<stdin>:1: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
hello
```

The FutureWarning itself obvioulsy isn't a big deal— I usually ignore them— but it's probably not a bad idea to avoid having some crash or weird behavior if we update that stuff. 


> Do you know what test was behaving unpredictably? I note that locally, the tests pass: that means this situation is not well-covered by our test suite. (See #1493).

By test, I don't mean a pytest test, I mean the `if (condition)` test. In the now replaced code I wrote looping through all of the paragraphs in the python-docx object, I used `if (element):` when checking if a particular parent or child existed. Occasionally, `if element` would return truthy but there wouldn't be an actual element there. I changed my code to if `len(element):` and the test worked as I expected, but that's not exactly conclusive! I reckon avoiding future problems is reason enough to leave the `if len()` version in place but I'm happy to try and recreate that bug to be sure if you're interested.

> This also makes me wonder if perhaps you have been running with `FORCE_DOCX_FOOTNOTES = False` locally? (Which is how things were by default until [recently](https://github.com/harvard-lil/h2o/pull/1485/files#diff-d7c867e0f418251ebe176b2353ce9d465609bce54643a4fc603d175d01194857R260). Have the fancy footnotes gotten a makeover too, or will that be a TODO for a future pass?

Totally. Integrating the changes since I started was the last thing I did and I definitely did not rigorously run it through its paces, which is the primary reason why this is still a draft.

>     3. It would be kind of nice to deploy this side-by-side 

The next push will include config options to swap it out. 

> But I can tell you were thinking about setting it up that way, by adding the `DOCX_SECTIONS = True` setting, which toggles off SOME of your stuff. I also see from the added trailing comma in `def export_via_aws_lambda(obj, html, file_type, docx_footnotes=None,)` that you were thinking about making `docx_sections` an arg, likely in anticipation of adding a view option/button.
> 
> So my questions are three:
> 
>     * Is the output with `DOCX_SECTIONS = False` viable/interesting/acceptable to you? If yes, let's go whole hog and offer it to admins as an export option for now, ala the buttons we were recently using. (Can send code links/more details if desired/useful.) If not, let's lose that setting, the trailing comma, and the export option.
Absolutely.

>     * What's your appetite for wrapping this all behind an option and exposing the legacy-styled export to admins via a button? And a setting to toggle: yours by default, legacy via special button; flip the setting to be legacy by default, yours by special button? Any appetite? Is that too much work / too risk averse / not useful enough? What are your thoughts?
Sure. Going to use that functionality in the future anyway for different export options, and it's minimal code.

>     * What's the best way for us to set up a side-by-side comparison? 
Good question. I think having the different buttons will help.

>     * Can you introduce us at a high level to your design? 

Yep. I changed a few of these at the last minute because I noticed some conlficting terminology— I started using the term header to distinguish from heading without really considering how that would be distinguished from actual headers. Should just take a few quick comments to clear it up.


>     * The differences between the 7x10 and 8.5x11 and templates. 

The 8.5x11 template as it exists is outdated. Since we had previously decided to go with the 7x10, I didn't bother to keep changes up-to-date with the 8.5x11. I'm not quite sure how to go forward with this. Some values in the section break logic would need to change, and that would change the values in styles.xml. It probably makes sens to try and just have some kind of transform logic but I'm loath to spend any more time on this.

>     * How did you produce these files? 
Yes, but they're all pretty heavily modified. My guess is the weird spaces in that fonts file came from a PyCharm reformatting problem. I did use automatic reformatting sometimes if I copied elements over from other Word docs. 

>     * What's your recommended process for making further edits and changes? I gather you generally make changes to a document in the Word UI, find the resulting XML, and the port that back here? Am I right about that?
Depends. I think the most straightforward route is to edit the XML, run an export, and see how it changes. If edits involve big layout changes, you'd probably need to edit specific styles, export those, and then copy them over. Once you save in word, all of your comments and structure are blown away. In the future, I could see having a utility that went through each style of a new reference doc, and wrote them into the current doc.

>     5. I'm wondering about the directory structure. [...] Do either or both of those changes sound good to you?
Sure– those both sound great. I realized too late that the question I posed about this was in the h2o channel and not the h2o dev channel, because I'm an expert communicator. I think that's a pretty reasonable way to do it, especially if we're bound to have mathy transformations rather than distinct reference docs. 

>     6. Your compilation script includes an `archive` option, 
The reason I included those was mainly for design reference. I don't see any need for them to be there, or even in the repo, though. We don't generally commit our design resources to repos. It could easily be changed to a folder on google drive. If you're trying to work through things visually, though, looking at the old numbers in the commit history or having to move back to old commits and build reference docs is a pain.

> Coming next: questions about particular lines of code, using that UI!

Coming next: me reading those :-)

docker/pandoc-lambda/function/app.py
```diff
@@ -123,6 +127,166 @@ def promote_case_footnotes(doc):
...
+        """ Assembles the section break, appends it to the appropriate graf, properly discards the empty wrapper,
+        because it gives a hoot and won't pollute. """
+
+        # The doc itself should have an existing default section attached to body, not a paragraph
+        if destination == 'doc-wide':
+            sec = doc.sections[0]
+        else:
+            sec = doc.add_section()
+            destination._element.xpath('w:pPr')[0].append(sec._sectPr)
+            body_element.remove(doc.paragraphs[-1]._element)
```
rebeccacremona, line 186: I wonder if it would be faster to use a lower-level API to add the section breaks right where you want them, rather than appending to the end of the doc and then removing. Is the current behavior, which clones all the props (https://github.com/python-openxml/python-docx/blob/da75fcf01f7f322e846e2ac3e1936aedd766acc8/docx/oxml/document.py#L35) useful, or would it be more useful if you could just add a bare one where you wanted, if we could work out a nice syntax for that?

ChefAndy: We could do that. I use python-docx to create the section because they're surprisingly fussy to work with and doing them directly strikes me as being more fragile and difficult to maintain. For example, having the sectPr child elements in the wrong order triggers a corrupt document error, or can result in things being ignored silently. However, we are going export a lot more frequently than modify section breaks, I suppose. I was also trying to avoid using protected functions like set_sectPr() anyway, but I probably end up having to do that more in my current method, so whatever.

Creating a string version of a section break using a template (being sure to include w and r namespaces in the xml,)
parsing that, and using the appropriate paragraph's ._element.set_sectPr() function should do the trick.

docker/pandoc-lambda/function/app.py
ChefAndy: Whoops— cleaned up the code but not the comment. What I have in place works well.

docker/pandoc-lambda/function/app.py
ChefAndy: Yeah— this is all different with cols removed.

docker/pandoc-lambda/function/app.py
ChefAndy: Yep. Cleaned up.

assets/export/docx/8.5x11/word/fontTable.xml
ChefAndy: I think these are artifacts from PyCharm's auto-format misbehaving.

assets/export/docx/7x10/word/styles.xml
ChefAndy: I don't see a way to mark text as a subheading in the casebook editor.

assets/export/docx/7x10/word/styles.xml
ChefAndy: Low-level cleanup. The more styles that can revert to the user's built-in styles, the more likely you are to have a mismatched patchwork of styles or break formatting. If one of our styles inherits from a default word style rather than something we explicitly defined, it's going to use whatever the default is on that system and look mismatched. If the user's typeface has weird spacing— like impact or comic sans or that really huge roman-looking one— it could break the layout.

assets/export/docx/7x10/word/document.xml
ChefAndy: I think it should stay (and probably be updated to include its sibling section breaks) but I'll write a comment about it to explain that. It's one of the few things in the reference.docx that pandoc actually uses, so uncommenting it is consequential.

assets/export/docx/7x10/style_structure_reference.md
ChefAndy: All of these styles with multiple numbers (Body Text 2, Body Text 3, etc) are built-in word styles. While it's probably not necessary to define everything, I'm worried about defaults slipping through and breaking the document's look and feel.

web/main/templates/export/tbd.html
ChefAndy: Yeah— kind of a bummer. Makes a night-and-day difference with some of the exported text, though.

assets/export/docx/7x10/style_structure_reference.md
rebeccacremona: Ah, cool, thx

assets/export/docx/7x10/word/styles.xml
rebeccacremona: Oh, gotcha, thx!

rebeccacremona:
> FutureWarning
Aha! Thanks so much for explaining. You know, I have no idea what I am missing out on by always running as a daemon, I should switch it up. 

So it sounds like this wants to be `if gp is not None and len(gp):`, then: cool. Seems a little aggressive for them to object to a person checking if an element is truthy, but if they are changing the API such that lengthless elements are no longer falsey, I guess they don't have a choice....

docker/pandoc-lambda/function/app.py
rebeccacremona: Cool beans: that's convincing. Doesn't sound worth it unless we know it to be way faster, which we don't.

web/main/templates/export/tbd.html
ChefAndy: FYI— I changed this significantly. I now filter out &nbsp; entities with a single .replace operation as it's headed towards the export. The elements I had &nbsp;s in to keep them from getting stripped— section headers and stuff— I put a text placeholder in those and swapped them in a subsequent .replace().

assets/export/docx/7x10/word/document.xml
ChefAndy: I did not get to documenting this better. I figured I just needed to move forward— but it'll be a good 'in-between' project of bite-sized updates.

docker/pandoc-lambda/function/app.py
```diff
@@ -5,7 +5,11 @@
 import subprocess
 import tempfile
 from docx import Document
+from docx.shared import Inches
```
rebeccacremona, line 8: inches instead of twips, hooray!!!

docker/pandoc-lambda/function/app.py
```diff
@@ -68,7 +72,7 @@ def promote_case_footnotes(doc):
         node.attrib[val_att] = "FootnoteReference"
         parent = ref.getparent()
         gp = parent.getparent()
-        if gp:
+        if len(gp) > 0: # 'if gp' triggers a warning and I saw a similar test behave unpredictably
```
rebeccacremona: Per previous discussion, sounds like this needs to be `if gp is not None and len(gp)`, to avoid crashing out when `gp` is None.

docker/pandoc-lambda/function/app.py
```diff
@@ -122,6 +126,134 @@ def promote_case_footnotes(doc):
...
+            pgNumTypeEl.set(qn('w:pgNumType'), pgNumType_value)
+            sec._sectPr.insert_element_before(
+                pgNumTypeEl, 'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
+                'w:textDirection', 'w:bidi', 'w:rtlGutter', 'w:docGrid',
+                'w:printerSettings', 'w:sectPrChange'
+            )
+
+
+    # the number of body styles we use consistently is much smaller than the number of section topper styles, but the
+    # values are more predictable. Body text can get unexpected style names handed down from pandoc
```
rebeccacremona: Sorry, I'm struggling a little here to understand what's taking place. True? The `topper_styles` array exists only so you can post-process it into `topper_style_faked_ids`? I only see the var used in that spot. If that's the case, perhaps just hard-code an array of ids?

Also, is there a quick way to describe the difference between a topper style and a body style, for dummies like me? I want to make sure I understand.

docker/pandoc-lambda/function/app.py
```diff
@@ -180,11 +311,15 @@ def handler(event, context):
             if not os.path.getsize(pandoc_out.name) > 0:
                 raise Exception(f"Pandoc produced no output.")
 
-            if options.get('word_footnotes', False):
+            if options.get('word_footnotes', False) or options.get('docx_sections', True):
```
rebeccacremona: Shall we make this default to False, like in other places?

docker/pandoc-lambda/function/table_of_contents.lua
```diff
@@ -1,4 +1,3 @@
-
```
rebeccacremona, line 1: Cool. Let's make a list of all the things that will change about the default export if this is merged, even with the flag `==False`.

docker/pandoc-lambda/reference_docx/README.md
```diff
@@ -0,0 +1,25 @@
+The contents of this directory is used to customize the output of Pandoc, the software used to produce DOCX exports of casebooks.
```
rebeccacremona, line 1: Cool, we'll want to update this README

rebeccacremona: We can do later this week

docker/pandoc-lambda/reference_docx/[Content_Types].xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Comment (rebeccacremona) [inline docker/pandoc-lambda/reference_docx/[Content_Types].xml]: Nitpick: can you put this, and the other source files that get zipped up to make a complete `reference.docx` into a `reference_docx/src` directory, so that there are fewer files right in `reference_docx`? Might seem ridic, but I think it will really help with repo readability.

docker/pandoc-lambda/reference_docx/word/document.xml
```diff
@@ -0,0 +1,912 @@
...
+xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main"
+xmlns:pic="http://schemas.openxmlformats.org/drawingml/2006/picture"
+xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing">
+
+  <w:body>
+    <w:p>
+        <w:pPr>
+            <w:pStyle w:val="Head Separator" />
+<!--            <w:sectPr>
+                <w:headerReference w:type="even" r:id="rId41"/>
```
rebeccacremona: In previous discussion, I asked about this commented out stuff and asked if it should go. You said no, that it was used by pandoc. I'm sorry, I'm still confused: does this need to be here and commented out? Is it a TODO that needs a comment?

docker/pandoc-lambda/reference_docx/word/fontTable.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<w:fonts xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
+&#9;<w:font w:name="Symbol">
+&#9;&#9;<w:panose1 w:val="02000500000000000000" />
```
rebeccacremona: These `&#9;` are still here.

docker/pandoc-lambda/reference_docx/word/styles.xml
```diff
@@ -0,0 +1,1969 @@
...
+    </w:style>
+
+    <!--
+        -== SECTION TOPPERS ==-
+    -->
+
+    <!--  These are very important and must be treated right to work properly. First, each must be preceeded with a Head
+        Separator (no, not a metal band) element because the styles put text in frames, and the frames will step on each
+        other if they abut directly. Secondly, the order of the elements is also important. Number first, Title second,
+        everything else third. Subtitle, headnotes, and link elements are interchangable. -->
```
rebeccacremona: Thanks so much for this comment!

web/main/models.py
```diff
@@ -2155,11 +2155,11 @@ def export(self, include_annotations, file_type='docx', export_options=None, is_
             'export_options': export_options,
             'export_date': datetime.now().strftime("%Y-%m-%d"),
             'cloned_from': cloned_from,
-        })
+        }).replace('\xa0', '')
```
rebeccacremona: Does this need all the other replacements that you have below? `.replace('&nbsp;', ' ').replace('_h2o_keep_element', '&nbsp;').replace('\x0a', ' ')`. What a bummer that we have BOTH `&nbsp;` AND `\x0a` in this output, which as I understand it, are both non-breaking spaces? I wonder where that's coming in, might be worth opening an issue about, to investigate.....

web/static/images/ui/casebook/fancyexport.svg
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rebeccacremona: sweet

web/main/templates/export/node.html
```diff
@@ -1,63 +1,91 @@
...
+{% if not is_child %}
     {% include 'export/credits.html' %}
 {% endif %}
 
 {% if not is_child %}
 {% include 'export/about.html' %}
-{% endif %}
\ No newline at end of file
+{% endif %}
+<div data-custom-style="Node End">&nbsp;</div>
```
rebeccacremona: I think your last commit had an accidental reversion, switching this and some other things in this file back to &nbsp; from _h2o_keep_element?

rebeccacremona (changes requested):
Sweet! Exciting that this is so close! I asked a bunch of questions and suggested a few nit-picky changes inline.

I have one more question that was not possible to make inline, since the files in question are binaries: this PR contains two `docx` files:
```
docker/pandoc-lambda/function/reference.docx
docker/pandoc-lambda/reference_docx/reference.docx
```
They are very different sizes. Is `docker/pandoc-lambda/function/reference.docx` the real one, and `docker/pandoc-lambda/reference_docx/reference.docx` left over from the 7x10/8x11 two-column setup? Or are both are correctly here?

General remarks: it is good for us to know that when merged, this PR as currently set up-will- affect the default export: the feature flag does not wrap all changes. So, we should discuss whether we want to:
a) change that
b) change our testing strategy
c) continue as planned.

ChefAndy:
# python-docx and docx styles:
* adds sections to docx
* in conjunction with the docx styles, implements alternating headers, chapters on new odd pages, etc.
* added, and subsequently removed two-column text support (at least for now)
* moved reference doc and supporting stuff to the lambda function directory
* fixed table support

## See: 
* docker/pandoc-lambda/function/app.py
* docker/pandoc-lambda/function/reference.docx & family

# Stripping existing styles and formatting:
* removing repeated non-breaking spaces with a .replace filter. They might be useful in a wysiwyg interface, but folks often use them for hacky formatting, and translating that into a different layout with different variable spacing is nigh impossible.
* removing existing div classes and styles if we didn't add them — added to the ToC lua filter loop to save a trip through the doc
* tried to remove inline element styles not applied in our editor, but it was too resource intensive. Maybe normalization at the TinyMCE config level for pasted content without trying to remediate existing content is the way to go.

## See
* web/main/templates/export/casebook.html
* web/main/templates/export/node.html
* web/main/templates/export/tbd.html
* web/main/templates/includes/legal_doc_sources/cap_header.html
* web/main/models.py
* docker/pandoc-lambda/function/table_of_contents.lua


# Button behind feature flag to use both export types
* fancy export button allows exporting new and old casebooks easily
* only available to superuser accounts

## See:
web/config/settings/settings_base.py
web/frontend/legacy/lib/ui/content/export_modal.js
web/frontend/styles/casebook.scss
web/main/views.py
web/main/templates/includes/action_buttons.html
web/static/images/ui/casebook/fancyexport.svg
web/main/models.py
web/main/utils.py

I owe everyone picking through this oversized PR dinner, somehow, someday. I see your struggle. 

_(Never mind that I created it.)_

rebeccacremona:
Love the writeup!

> removing existing div classes and styles if we didn't add them — added to the ToC lua filter loop to save a trip through the doc
> tried to remove inline element styles not applied in our editor, but it was too resource intensive. Maybe normalization at the TinyMCE config level for pasted content without trying to remediate existing content is the way to go.

I know there is a ton of that sort of thing in legacy cases...... I wonder if @holtchesley can think of any places where modern H2O + CAP casebooks use style attributes or classes in user-supplied markup or imported markup.... Maybe just image alignment?

> I owe everyone picking through this oversized PR dinner, somehow, someday. I see your struggle.
> (Never mind that I created it.)
😂

web/main/models.py
ChefAndy: This is extraneous and I removed it. Regarding the larger issue, I think they're both coming in through pastes, and TinyMCE adds \xa0 when a user types multiple spaces, too. I'm a bit surprised by that because I think \xa0 is specific to some encodings.  I think this should be an issue— but relatively low priority.

docker/pandoc-lambda/function/app.py
ChefAndy: if gp is not None and len(gp)

ChefAndy: Yep— explaining in code comments.

web/main/models.py
rebeccacremona: Ah, cool beans.

docker/pandoc-lambda/function/app.py
ChefAndy: Simplified and (heavily) documented code en route.

docker/pandoc-lambda/function/app.py
ChefAndy: Yep

docker/pandoc-lambda/reference_docx/word/fontTable.xml
ChefAndy: Weird. This file must have a different file encoding than the other files so pycharm adds those tab chars but won't display them like that. It clearly reads fine to the machine but it sure looks ugly. I just flattened it.

holtchesley:
> I know there is a ton of that sort of thing in legacy cases...... I wonder if @holtchesley can think of any places where modern H2O + CAP casebooks use style attributes or classes in user-supplied markup or imported markup.... Maybe just image alignment?

There are ~5k unique style tags in text blocks, and ~1k classes (which likely don't do much at all). Image alignment happens with a class for html, and a data-custom-style for export. I don't think any of the style or classes make it to export with the exception of bold and italics, but it's possible there is some other smattering of cases(though I'd be surprised).

ChefAndy:
> > I know there is a ton of that sort of thing in legacy cases...... I wonder if @holtchesley can think of any places where modern H2O + CAP casebooks use style attributes or classes in user-supplied markup or imported markup.... Maybe just image alignment?
> 
> There are ~5k unique style tags in text blocks, and ~1k classes (which likely don't do much at all). Image alignment happens with a class for html, and a data-custom-style for export. I don't think any of the style or classes make it to export with the exception of bold and italics, but it's possible there is some other smattering of cases(though I'd be surprised).

This is 100% on me. I conflated weird styling we were getting in through tables (a pandoc thing,) some whackadoo text structuring with paragraphs just all over the place (I assume from more copied and pasted tables poorly translated before even reaching the editor window, or maybe some CSS layout) and inline styles. You're totally right that this stuff generally doesn't make it through. I've done some style structure cleanup which should nip most of it. I don't think we can automatically clean up the structural problems.

docker/pandoc-lambda/reference_docx/word/document.xml
rebeccacremona: In the spirit of moving forward, I will make an H2O issue asking about this commented-out code instead.

ChefAndy:
@rebeccacremona Oh— I forgot to ask— are you still getting flake8 static? It never barked at me which leads me to believe I've got something different inmy environment I should change.

docker/pandoc-lambda/README.md
```diff
@@ -1,76 +1,13 @@
 ## Testing
 
-### ...locally, against the Lambda Runtime Interface Emulator
+### To test the new sectioning code,
```
rebeccacremona: I think this changes the wrong README, right? Everything in this file, about testing locally against the Lambda Runtime Interface Emulator, against AWS Lambda, updating the image, deploying, still stands.

The README that needs updating with information about new styles is docker/pandoc-lambda/reference_docx/README.md

web/frontend/legacy/lib/ui/content/export_modal.js
```diff
@@ -51,15 +53,17 @@ function downloadFile (includeAnnotations, experimental=false, aws_lambda=false,
...
 }
 
-function showExportModal (e, experimental=false, aws_lambda=false, docx_footnotes=false) {
+function showExportModal (e, experimental=false, aws_lambda=false, docx_footnotes=false,
+                           docx_sections=false) {
   new ExportModal('export-modal', e.target, {
     'click #export-modal': (e) => { if (e.target.id === 'export-modal') this.destroy()},
     'click .export': (e) => {
-      downloadFile(e.target.value, experimental, aws_lambda, docx_footnotes);
+      downloadFile(e.target.value, experimental, aws_lambda, docx_footnotes, docx_footnotes);
```
holtchesley: Should the send `docx_footnotes` be a `docx_sections`?

web/frontend/legacy/lib/ui/content/export_modal.js
```diff
@@ -51,15 +53,17 @@ function downloadFile (includeAnnotations, experimental=false, aws_lambda=false,
       + (includeAnnotations === "true" ? '?annotations=true' : '?annotations=false')
       + (experimental ? '&experimental=true' : '')
       + (aws_lambda ? '&aws_lambda=true' : '')
-      + (docx_footnotes ? '&aws_lambda=false&docx_footnotes=true' : ''));
+      + (docx_footnotes ? '&aws_lambda=false&docx_footnotes=true' : '')
```
holtchesley: The `&aws_lambda=false` should be taken out here, I think the code has been updated to remove the non-lambda pathway.

docker/pandoc-lambda/README.md
rebeccacremona: Let's update that later, with instructions about how to re-compile the packaged up reference docx, introduce people to the directory structure and give hints where to change what, etc. later. We can make an issue about it.

Seems fine to either just revert these changes, or revert these changes and add that info to the top of the PR description? Or append it to the bottom of the other README?

rebeccacremona:
> @rebeccacremona Oh— I forgot to ask— are you still getting flake8 static? It never barked at me which leads me to believe I've got something different inmy environment I should change.

I just saw the `flake8` errors in Circle CI; looks like smooth sailing there now.

docker/pandoc-lambda/function/old_pr1491/table_of_contents.lua
```diff
@@ -0,0 +1,147 @@
+toc_prefix = [[
```
rebeccacremona, line 2: I pulled down the `table_of_contents.lua` that's on develop and ran diff: 
```
diff ~/Downloads/table_of_contents.lua docker/pandoc-lambda/function/old_pr1491/table_of_contents.lua
1d0
< 
54,56d52
<         <w:tabs>
<           <w:tab w:val="right" w:leader="dot" w:pos="9350"/>
<         </w:tabs>
64d59
<             <w:rStyle w:val="Hyperlink"/>
65a61
>             <w:iCs/>
```
https://github.com/harvard-lil/h2o/blob/develop/docker/pandoc-lambda/function/table_of_contents.lua#L54-L56 is present on develop, but not in `old_pr1491/table_of_contents.lua`; there's `<w:iCs/>` in `old_pr1491/table_of_contents.lua` but not on develop. 

Just checking, is that correct? I would have expected these files to be the same. I think that's almost the same diff as ` docker/pandoc-lambda/function/table_of_contents.lua` below.

web/config/settings/settings_base.py
```diff
@@ -258,6 +258,7 @@
 
 MAX_EXPORT_ATTEMPTS = 3
 FORCE_DOCX_FOOTNOTES = True
+FORCE_DOCX_SECTIONS = False
```
rebeccacremona, line 261: Okay, one last question: if `FORCE_DOCX_SECTIONS = False`, then the results of the tests shouldn't change at all, right? 

I think every change to a test file in the repo is an instance of something slipping through.

These 28 byte shrinking of the complete casebook docx'es is probably just that thing I noticed about the "old" lua filter not being identical to the lua filter on develop.
image

The changes to the export of individual sections and resources might need some looking at.

web/frontend/legacy/lib/ui/content/export_modal.js
ChefAndy: Whoops— think that was left in a stash
Edit: oh- just hadn't pushed yet when I responded.

web/frontend/legacy/lib/ui/content/export_modal.js
ChefAndy: Right— will do.

web/config/settings/settings_base.py
ChefAndy: Got it. It was from one of those stupid, hilarious yet infuriating bugs that you cant see even though you're looking right at it— i was running a .replace on the template name like it had side effects rather than re-assigning the string value because I'm a professional.

docker/pandoc-lambda/function/old_pr1491/table_of_contents.lua
ChefAndy: Yep. Forgot I made style updates to this, too.

web/frontend/legacy/lib/ui/content/export_modal.js
ChefAndy: good to go

web/frontend/legacy/lib/ui/content/export_modal.js
ChefAndy: set

docker/pandoc-lambda/README.md
ChefAndy: Derp. Set.

web/config/settings/settings_base.py
rebeccacremona: Oof!!!!

rebeccacremona (approved):

PR #1494, opened 2022-01-25
Changes for development on M1 Macs

bensteinberg:
This also removes the pandoc-on-server dependency.


docker/Dockerfile
```diff
@@ -8,15 +8,13 @@ ENV LANG=C.UTF-8 \
 
 RUN apt-get update && apt-get install -y nano postgresql-client
 
+ARG BUILDARCH
```
rebeccacremona, line 11:
```suggestion
ARG BUILDARCH=amd64
```

bensteinberg: Thanks!

PR #1501, opened 2022-02-08
Preload animated export button

holtchesley:


web/frontend/styles/casebook.scss
```diff
@@ -143,7 +143,7 @@ main > header.casebook {
       color: $light-blue;
     }
     &:focus {
-      @include generic-focus-styles;
+      // @include generic-focus-styles;
```
holtchesley: Get rid of the focus-ring around the action buttons (this shows up longer for exports).

web/frontend/styles/casebook.scss
```diff
@@ -167,10 +167,26 @@ main > header.casebook {
...
 
-    &.export { background-image: url('~static/images/ui/casebook/export.svg'); }
-    &.export.animating { background-image: url('~static/images/ui/casebook/export-anim.svg'); }
-    &.export.fancy { background-image: url('~static/images/ui/casebook/export-fancy.svg'); }
-    &.export.fancy.animating { background-image: url('~static/images/ui/casebook/export-anim-fancy.svg'); }
+    &.export {
+        background-image: url('~static/images/ui/casebook/export.svg');
+        &::after{
+            position:absolute; width:0; height:0; overflow:hidden; z-index:-1;
+            content: url('~static/images/ui/casebook/export-anim.svg');
```
holtchesley, line 174: Preload the animated version of the export animation in an invisible pseudo element so that Safari doesn't make an aborted later attempt to load it.

web/frontend/styles/casebook.scss
rebeccacremona: sorry to butt in: will keyboard users still be able to tell they are focused on the button?

holtchesley: image
Cancel is focused. This may be too subtle.

rebeccacremona: 🤷‍♀️ 👍 at least there is still something.

holtchesley: I rolled it back. It still looks just fine.

PR #1504, opened 2022-02-09
Fix: span tag grabbing event

ChefAndy:
Last minute, I changed an `<i>` tag to a `<span style=` tag without considering that `span` might be the target of the click event while the `I` tag wouldn't. I just took that style information and put it in the parent `div` that needed to be the target.


holtchesley:
Definitely been there.

PR #1505, opened 2022-02-09
Rate limit exports

holtchesley:


web/main/models.py
```diff
@@ -4077,7 +4081,32 @@ class EmailWhitelist(models.Model):
...
+        >>> _ = full_casebook.export(False)
+        >>> _ = resource.export(False)
+        >>> ls.refresh_from_db()
+        >>> assert ls.export_average_rate == prior_count + 2
+        """
+        live_settings = LiveSettings.load()
+        current_time = datetime.now()
+        minute = current_time.hour*60 + current_time.minute
+        elapsed_minutes = (minute-live_settings.export_last_minute_updated) % 1440
+        new_rate = max(live_settings.export_average_rate - (elapsed_minutes * settings.EXPORT_RATE_FALLOFF), 0)
```
rebeccacremona, line 4102: This is a clever technique! I had to step through the math embarrassingly slowly on a note card to see how it works 😄.

web/main/models.py
```diff
@@ -2166,8 +2166,10 @@ def export(self, include_annotations, file_type='docx', export_options=None, is_
 
         if file_type == 'html':
             return html
-
-        return export_via_aws_lambda(self, html, file_type, docx_footnotes=docx_footnotes, docx_sections=docx_sections)
+        if not LiveSettings.export_is_rate_limited():
+            return export_via_aws_lambda(self, html, file_type, docx_footnotes=docx_footnotes, docx_sections=docx_sections)
+        logger.info(f"Exporting Casebook {self.id} prevented due to rate limits")
```
rebeccacremona: Nit: I think `Casebook {self.id}` is not accurate for this export method, am I right? I think this path is only for exporting sections and resources?

web/main/models.py
holtchesley: I drew a little diagram and then verified that mod arithmetic in python works the way I wanted it to (there are some languages where it doesn't, strangely).

web/main/models.py
holtchesley: 100%. Corrected it.

PR #1506, opened 2022-02-10
Fixup footnote edgecase

holtchesley:


web/config/settings/settings_base.py
```diff
@@ -258,7 +258,7 @@
 
 MAX_EXPORT_ATTEMPTS = 3
 FORCE_DOCX_FOOTNOTES = True
-FORCE_DOCX_SECTIONS = False
+FORCE_DOCX_SECTIONS = True
```
holtchesley, line 261: This will make footnotes and sections the default on stage/prod when it goes live.

rebeccacremona (approved):
LGTM!

Curious if you updated the Lambda RIE just for fun or because you saw a message of some sort go by: I haven't been sure how we'd know when an update was called for.

(Also, unrelated to this PR: I spotted in the new test artifacts that sometimes we have `_h2o_keep_element;` and sometimes we have `_h2o_keep_element`, without the semi-colon. I'll file an issue to check into whether that's a stray char somewhere in a template or is correct.)

Issue #1507, opened 2022-02-10
Stray `;`?


rebeccacremona:
I noticed that sometimes we have [`_h2o_keep_element`](https://github.com/harvard-lil/h2o/blob/develop/web/test/files/export/export-casebook-no-annotations.html#L49) and sometimes we have [`_h2o_keep_element;`](https://github.com/harvard-lil/h2o/blob/develop/web/test/files/export/export-casebook-no-annotations.html#L81), with a semi-colon.

When we have a moment, let's see if there's a stray `;` or two in a template (and whether it makes it into export), or whether it serves a purpose.

jcushman:
Ah, this explains why docx2txt.sh showed extra semicolons! They are in fact coming through to the export, as tiny white text. Following select, change text to black, increase size:

<img width="647" alt="image" src="https://user-images.githubusercontent.com/376272/153494420-21cf50cb-484f-4450-8cec-bc59e53ed413.png">

PR #1520, opened 2022-02-11
Semicolonectomy

ChefAndy:
Addresses #1507 

Extracted a stray semicolon left in a structural tag during the non-breaking space-> spacer text replacement. It wasn't noticeable in word because it's very small white text on a white background. I'm pretty sure it was the only one.

### DIFF
--- a/web/main/templates/export/node.html
+++ b/web/main/templates/export/node.html
@@ -97,4 +97,4 @@
 {% if not is_child %}
   {% include 'export/about.html' %}
 {% endif %}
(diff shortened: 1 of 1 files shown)


bensteinberg:
I think I see some others, in test files -- see `rg ';<' --type html .`

ChefAndy:
@bensteinberg Yep— it's the reason why the tests failed. I forgot to update the HTML files which tests compare export output to. Again. 2 second fix that I'll put in once I finish up this PR.

ChefAndy:
Rolling this change into #1524

PR #1525, opened 2022-02-14
Re-enable volume mount for pandoc-lambda on local dev

jcushman:
This adds a docker-compose.override.yml that only runs on local dev and not on CI (I think, assuming I'm adding `-f` to the right CircleCI file, Ben).

This way we can do things like having a `volume` mount for `pandac-lambda` locally without having it on CI. We initially removed the `volume` because it masked an issue on CI with the Dockerfile not copying files over to the image correctly, but it's useful locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ jobs:
       - run:
           name: "Test Django"
           command: |
-            docker-compose up -d
+            docker-compose -f docker-compose.yml up -d       # use -f to suppress docker-compose.override.yml
             docker-compose exec web ./manage.py collectstatic --noinput   # collect static files
             mkdir -p junit/pytest                            # storage for circleci test info
             docker-compose exec web pytest \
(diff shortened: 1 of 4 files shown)


rebeccacremona:
Neat!

Looks like you are also tweaking the procedure for updating packages? Cool. I'm not quite sure I'm getting the order of operations... am I saying this right?:
- against an already running container, run the compile command
- increment the image number
- comment in the "bash" command so that the entrypoint script will be a no-op
- re-run docker-compose up
- comment back out the "bash" command
- restart the container  (or run docker-compose up) again

?

It seems to me you could skip the bash command entirely.

I think the only time you would find yourself needing the `bash` command to start up your container is if you installed new python packages in a running container (e.g. pip sync) and then continued developing, rather than rebuilding the image with the new requirements.... but then got rid of that container.

jcushman:
Yeah, mostly my intent here was, since the volume is being mounted again, no special procedure is needed for updating requirements, just do it normally in the running container the same as for `web`.

But there's an edge case because this container has an entrypoint and `web` doesn't -- if you make some code edits that happen to require an additional package to be installed, but then shut down the container before updating requirements.txt correctly, then the container can't start again so you can run pip-compile, and it's confusing what to do next. (The answer being, either stash your code changes or edit `command` to get the container to start so you can pip-compile.) I threw in a couple of comments to try to smooth that situation, but maybe it's so edge-casey that the comments aren't helpful.

rebeccacremona:
Ah, good, glad my hypothesis was right. My personal dev flow never includes installing packages into a running container, so I've never been in that particular situation; no reason to think other people won't find themselves where you did!

PR #1526, opened 2022-02-14
Pandoc auto-refresh

jcushman:
If #1525 works, then we can auto-refresh the lambda server as well with `watchmedo`. I dunno, seems to work.


rebeccacremona:
(hmm, how to force this diff to correct, since `17ef4f5` is merged already.... if it doesn't fix itself in a couple minutes left to its own devices will mess with it)

jcushman:
I guess cause of the merge commit on the other one it needed a rebase here. I did a `git fetch upstream`, `git rebase upstream/develop`, `git push --force`.

PR #1527, opened 2022-02-15
Style updates

### LINKED ISSUE #1514: [new export styling] Footnote Marks are too loud, need padding
They're huge and orange and probably shouldn't be— this could be a latent styles issue. Time to pay styles.xml a visit. I don’t think it will be difficult and it will be a readability improvement.

reference: https://opencasebook.org/casebooks/2459/

### LINKED ISSUE #1518: [new export styling] TOC Styling
The current tab-based layout likely will remain in place despite its clear shortcomings with multi-line entries— not having a page number on the same line as the entry is a more severe problem than not having a title, subtitle and ordinal on a different page than node contents. The inflexibility of word's vertical spacing between paragraphs of the same style— line spacing which is always present, or space before/space after which is either 100% there or 100% absent— makes styling text that functions as both a headline and a node pretty tricky. 

reference: https://opencasebook.org/casebooks/33-untitled-casebook-176108/layout/

ChefAndy:
# TOC revamp & Tricky Footnote Reference
fixes #1518 
fixes #1514

### docker/pandoc-lambda/function/app.py
* Apparently character styles don't allow any horizontal padding. They do allow you to add a white border, but it's applied to all sides to any significant changes push the line height out of whack. I added spaces to the element's `.text` during footnote processing.

### docker/pandoc-lambda/function/table_of_contents.lua
* Changed TOC output for readability and organization— involved splitting the title run into two

### docker/pandoc-lambda/function/reference.docx, docker/pandoc-lambda/reference_docx/
* 18 new character styles for TOC
* Modified TOC paragraph styles
* Restyled footnote markers

### web/main/templates/export/casebook.html
* passing ordinal and title text separately

plus test files, etc.

#### Update: 

I reckon the separation of the orignal output vs the current implementation should be up-to-date or removed entirely. Added a couple of conditions in app.py so you can get the original output we were shooting for if you disable FORCE_DOCX_SECTIONS.


ChefAndy:
(incoming PR to isolate the changes for non-sections export to keep it consistent if we want to compare)

PR #1530, opened 2022-02-16
Update/move JS deps

holtchesley:


bensteinberg:
I think you only have part of the move for `package-lock.json`...

holtchesley:
> I think you only have part of the move for `package-lock.json`...

If you push the wrong HEAD, is that a breach commit?

PR #1552, opened 2022-05-27
Defend against falsey value, not just None, fixes #1543

### LINKED ISSUE #1543: Admin *IdFilter doesn't handle empty string
We [check for `None`](https://github.com/harvard-lil/h2o/blob/develop/web/main/admin.py#L161), but http://opencasebook.org/django-admin/main/casebook/?collaborator-id=, for example, now results in 
```
ValueError at /django-admin/main/casebook/
invalid literal for int() with base 10: ''
```
at
```
File "./main/admin.py" in queryset
  162.             users = User.objects.filter(id=value)
```

lizadaly:
I _think_ we want to exclude any value here that's not a string (meaning the empty string), rather than just `None`.

Verified that just hitting "enter" in any of these filters just returns the full unfiltered list, as expected, but adding a User ID value (for example) still filters.

### DIFF
--- a/web/main/admin.py
+++ b/web/main/admin.py
@@ -138,7 +138,7 @@ class CasebookIdFilter(InputFilter):
 
     def queryset(self, request, queryset):
         value = self.value()
-        if value is not None:
+        if value:
             return queryset.filter(casebook_id=value)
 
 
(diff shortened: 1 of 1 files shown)


rebeccacremona:
Huh, I wonder why I have an explicit `is not None` check everywhere, here AND in Perma...... I wonder if it was that way on purpose with some original filter, say one that could filter on `True/False` or one that was intended to find objects with some attribute set to the empty string...... and then it got cut and pasted to every subsequent filter.... 

Regardless, I agree, plain falsey check seems right here!! Thx!

rebeccacremona (approved):

PR #1558, opened 2022-06-03
Usage dashboard 

lizadaly:
Start of a usage dashboard to live in the Django admin and provide current usage stats about the product.

Most of the queries are artisanal SQL; at the start of the request session it creates a few temp tables for expensive queries and those get cleaned up by postgres at the end of the session. 

I've omitted two types of metrics requested: counts of books with multimedia (which is expensive to compute because hotlinking means we must introspect the HTML with regexp) and top-N lists by user views, which must be computed from web logs rather than Django db values.

This runs very quickly on my local machine, but will see once merged how this does on staging.

image


lizadaly:
(I'll also want to audit the logic in each query one-by-one because it's very easy to generate numbers that are just total lies.)

PR #1560, opened 2022-06-13
Re-use same HTML structure for single-element "Manage Casebook" link

lizadaly:
It annoyed me that the hover state for the "Manage Casebooks" link didn't change my cursor to a pointer. This turns out to be because the CSS expected the link to exist in the context of a list. Now it does, so it inherits all the same styles including `cursor: pointer`.

image

(It's hard to take a screenshot without modifying the pointer but this just shows the link I mean.)

Ideally these [links should be buttons](http://web-accessibility.carnegiemuseums.org/content/buttons/), I think, but that should probably be done holistically. 

I also considered changing the nesting of the SCSS to move the `a` property out of being nested in the `ul`, but decided this was a change less likely to have side effects.

(Apologies that my editor is aggressive about dropping trailing whitespace—I tried turning off that VSCode feature in this workspace but it seems to be ignoring me.)


matteocargnelutti:
(The PR assignment bot didn't pick this one up so I took a minute to have a look).

Looks good to me 😄 .

Only one question:
- Do we want to commit [web/webpack-stats.json](https://github.com/harvard-lil/h2o/pull/1560/files#diff-9793dc1fd6c8ad509da1b31e725df8e64a3452c06175f6bcff0a9c31a0c9241a) ?

matteocargnelutti (approved):

lizadaly:
The README implied that it should be committed, but it also said I should run `npm run build`, which I didn't, so I will.

lizadaly:
Built resources are probably a no-op but I did generate them so here they are. Also assigning to Becky to review when she's back since this is my first JS PR and we should probably clarify the README a little on the commit process since we were both unsure of what was right!

rebeccacremona (changes requested):
(I personally love the aggressive cleaning up of trailing whitespace.) 

You know what, I think the problem is that these anchor tags (that indeed should be buttons) lack `href`: they can't receive keyboard focus, AND, I think that's enough to stop browsers from switching the cursor style to a pointer on hover as well. In my firefox dev tools, adding the stynken `href="#"` to the `<a>` gets me a pointer without needing to change the structure of the html. Would you be willing to give that a try, maybe add `role="button"` while you are at it, and see if you get the same effect you were looking for?

Oof, this feature needs some attention..... The on-hover effect is also inaccessible... Let's address in a separate PR sometime soon.

I'm totally with you about updating the README.... I honestly wasn't sure whether a person should commit the built assets or not either; I am new to this setup where CI can build and commit. To me "commit if you want to" seems pretty weird.... maybe we adopt a convention of not committing, even if it's harmless to?

lizadaly:
Great point about keyboard focus, I hadn't noticed that absence. I added the same attributes to the other "buttons" because they were also not receiving focus.

rebeccacremona (approved):

PR #1561, opened 2022-06-13
Allow setting an API key from the environment in local dev

lizadaly:
As shipped, local dev won't have external API keys available, but if the developer adds them to their shell this will propagate it through to the container and have it picked up by Django.

```
root@37c7bcfb6a66:/app/web# env | grep API
CAPAPI_API_KEY=XXXXXX
```

and verified on the front-end that I could add a Cap case now:

image


bensteinberg (approved):
Looks good!

Issue #1562, opened 2022-06-16
Usage dashboard counts should match front end


lizadaly:
* Check that  "professors with published casebooks" and "casebooks" (published) is filtering correctly on "currently published" versus historically published. Example frontend reports published casebooks as 337 vs. dashboards 492.

PR #1565, opened 2022-06-21
[#1562] Correct some usage dashboard counts to better align with front-end

lizadaly:
Re-use the logic found in the search "index" SQL to allow counts to match the front end. Replace one usage of the Django ORM for professors with a reporting table to better align the logic with the other tables.

Key updated numbers:

image

Now match values on the front end for Casebooks and Authors:

image


rebeccacremona (approved):
LGTM!

PR #1569, opened 2022-06-22
Create a Casebook Plus Sign

shljessie:
Screen Shot 2022-06-22 at 4 29 25 PM

Adding a New Plus Sign Icon for the Create a Casebook.


matteocargnelutti:
**Note:** Chatted with @shljessie about the opportunity of maybe using `<button>` and an SVG here to help with accessibility and scaling 😄.

lizadaly:
I'm not familiar enough with the code to know for sure—does this button work? I'm guessing so but just checking!

shljessie:
Yes it does! Im using the same line for the original create a casebook button. 
<button type="button" class="create-casebook-plus" data-action="show-casebook-modal"> </button>
don't know if there is a better way though!

shljessie:
```
 <button type="button" class="create-casebook-plus" data-action="show-casebook-modal">
            </button>
```

lizadaly (approved):
Right on then! Once you merge the approved PR, in a minute or so it should end up on our staging environment at https://h2o-stage.lil.tools/, and we can spot-check it there too.

rebeccacremona:
Remark from the peanut gallery: once this is a keyboard-accessible button, it is worth testing whether, if you use the new button to open the modal, if when you press escape, keyboard focus is set to the new button. I -think- the modal JS is smart enough for that to work correctly, but am not positive.

lizadaly:
@rebeccacremona I tested it out on Chrome locally and it both receives keyboard focus when tabbing, then you can hit enter to open the modal, and when the modal is closed with ESC the focus is still on the new button.

lizadaly:
I'm merging because of perms on the main repo.

PR #1570, opened 2022-06-24
Upgrade Django to 3.2

dongkyunc:
Why: Support for 2.2 LTS ended on April 11, 2022.

What changed: 
- `requirements.{in,txt}` 
- Certain kwargs to `on_delete` switched from the string 'ON_DELETE' to the empty function `django.db.models.DO_NOTHING`

Any concerns: I changed a couple of migration files, do additional steps need to be taken re them?


lizadaly:
This looks innocuous to me (good!) but also pulling in Ben to take a look.

bensteinberg (commented):
LGTM, but I am also going to drag @rebeccacremona into this

bensteinberg:
Oh I see @rebeccacremona applied the hooray emoji above, which I think constitutes a review

rebeccacremona:
Oh I didn't mean it as one :-) giving a quick look

bensteinberg:
FWIW here are the release notes: https://docs.djangoproject.com/en/3.2/releases/3.2/

rebeccacremona:
> [X_FRAME_OPTIONS](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-X_FRAME_OPTIONS) now defaults to 'DENY'. In older versions, the [X_FRAME_OPTIONS](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-X_FRAME_OPTIONS) setting defaults to 'SAMEORIGIN'. If your site uses frames of itself, you will need to explicitly set X_FRAME_OPTIONS = 'SAMEORIGIN' for them to continue working.

That is worth knowing about; but I don't think H2O uses any iframes.

> [django.utils.html.escape()](https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.html.escape) now uses [html.escape()](https://docs.python.org/3/library/html.html#html.escape) to escape HTML. This converts ' to &#x27; instead of the previous equivalent decimal code &#39;.

I could imagine this affecting exports, but probably not

> [urlencode()](https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.http.urlencode) now encodes iterable values as they are when doseq=False, rather than iterating them, bringing it into line with the standard library [urllib.parse.urlencode()](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) function.

This would hopefully get caught by tests if we were doing anything fancy there, but is also good to know about.

LGTM!

PR #1572, opened 2022-06-24
Fix support for django debug toolbar

rebeccacremona:
This is nice to have around, especially when debugging DB usage.


lizadaly (approved):
Cool! This is more convenient than remembering to install it manually, which I was doing before.

varun-magesh:
Thank you Rebecca! This works for me too.

PR #1582, opened 2022-07-05
Update Django, handle some deprecations

bensteinberg:
This is a little bit of a mixed bag -- a point version update to Django:

https://docs.djangoproject.com/en/dev/releases/3.2.14/

plus changes to remove warnings after upgrading to 3.2 (closes #1571), and removing our settings for `USE_TZ` (closes #1054).

PR #1586, opened 2022-07-05
Add migrations that were pending since the 3.2 upgrade changes

lizadaly:
I think the changes in https://github.com/harvard-lil/h2o/pull/1582 required a migration (and possibly some others related to just 3.2 itself). These applied fine on local dev, and I don't think they'll have any functional implications as they don't really apply to the underlying DB, just Django.

(I noticed because I was generating migrations for real code changes and saw these mixed in.)


bensteinberg:
Oh whoops I probably should have added this in #1582.

bensteinberg (approved):
LGTM, thanks for catching this.

Issue #1044, opened 2020-01-10
Let's drop fields/tables from pre PR 812


rebeccacremona:
We have not yet completed the "After Deploy" steps from https://github.com/harvard-lil/h2o/pull/812

> There are two temporary sets of data that will need to be factored out after deploy:
>
>- On `Content::Annotation`, `global_start_offset` and `global_end_offset` have been added in order to migrate to the new global offsets while preserving the old offset information such that it can be restored if we find an issue immediately after deploy. Once we've confirmed everything looks good, `start_paragraph`, `end_paragraph`, `start_offset`, and `end_offset` will all be removed and `global_*_offset` will be renamed to `*_offset`.
>- Similarly, `Content::Node` has a new property named `raw_headnote` that preserves text that has been sanitized and placed into `headnote`. I believe this sanitization is reasonable and there's no need to preserve the "raw" form here, and if all agree we can delete this column as well.

We've been hanging on to them because of the annotation-shifting that occurred during the migration, in the event we think of a good way to programmatically improve the situation.

I'm creating this issue to make a place for discussion, and as a reminder that we do want to delete those columns and tables.

PR #1588, opened 2022-07-05
[#1044] Drop legacy annotations fields

lizadaly:
As referenced in #1044 , these fields are no longer used and can be safely dropped.

I verified manually that all annotation features seemed to work, and that the Django admin annotation page behaves as expected.


rebeccacremona (approved):
Huzzah!

PR #1589, opened 2022-07-05
adding create a casebook popup design changes

shljessie:
Screen Shot 2022-07-05 at 2 36 49 PM

Screen Shot 2022-07-05 at 2 40 24 PM

Added svgs for new icons on the popup page, not sure if this is the right way to make css changes though! 😊

### DIFF
--- a/web/frontend/legacy/lib/ui/content/dashboard.js
+++ b/web/frontend/legacy/lib/ui/content/dashboard.js
@@ -23,7 +23,7 @@ class NewCasebookModal extends ModalComponent {
             <div class="modal-button-container">
             <div class="modal-button-box" >
               <a href="/casebooks/new">
-                <svg width="87" height="87" viewBox="0 0 86 86" fill="none" xmlns="http://www.w3.org/2000/svg">
+                <svg width="95" height="95" viewBox="0 0 95 95" fill="none" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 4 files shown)


lizadaly:
This looks so nice!

Two things:

* It looks like most SVGs in the repo are distinct files referenced as background images (see `casebook.scss`). Could you do that here too, so the SVG code doesn't have to be inlined with the template?
* I think to keep it more consistent with the previous behavior we might want keep the word "Search" in the second option. What do you think about labelling it "Search for a Casebook to Customize"? Is that too long?

shljessie:
Screen Shot 2022-07-05 at 3 59 30 PM
Added those changes! 😊

lizadaly:
Much cleaner! Let's check with @rebeccacremona tomorrow to ensure those buttons are maximally accessible. I think for example we want to make an association between the text labels and the buttons themselves.

rebeccacremona (changes requested):
This looks so nice!

Alas, we aren't supposed to nest buttons inside links. Since when you click, you are taken to a new destination, we do in fact want links here (links -> navigation; buttons -> trigger in-page actions), so that's good! But instead of the button inside, I think we just want a `<span>`. For good measure, you can made it a `<span aria-hidden="true">` so that screen readers know it's only an icon for decoration. Then this will be fully accessible.

rebeccacremona (approved):

PR #1590, opened 2022-07-05
Implements by-casebook full-text search for legal documents

### LINKED ISSUE #1568: Add full-text search endpoint to the API
There is no way for students or professors to search through the entire text of a casebook. This is a significant advantage of a digital casebook that is currently under-supported. The first step in enabling this functionality is to create an API endpoint that returns the results of a full-text search through a given casebook.

From @lizadaly, casebook metadata search (which is what you can do through the search box on opencasebook.org) is done by:
1. indexing metadata using this sql (https://github.com/harvard-lil/h2o/blob/develop/web/main/create_search_index.sql)
2. recalling metadata using this Django Model (https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L752-L881)

On first glance, it doesn't seem like the Model needs to change to do the indexing.

Right now, it seems like the steps here are:
- [x] Expand `create_search_index.sql` to index casebook full-text.
- [x] Write a method that will query casebook texts in the SearchIndex for a given query string
- [x] Expose this method through the h2o API

varun-magesh:
This PR is the first-step in enabling full-text search through casebooks. It implements the `/casebooks/<id>/search view`, which displays a list of legal documents associated with casebook `<id>` and displays a search bar that searches the full text of all the listed documents. To implement this view, it full-text indexes all of the cases used by H2O. This functionality has more potential uses that are not fully explored in this PR.

In order to complete full-text search through resources, we need to:
- [ ] index text resources
- [ ] index link resources
- [ ] (optionally) index annotations
- [ ] integrate search functionality into casebook view UI.

This PR is standalone and complete --- it can be merged now, or it can be merged together with other changes that complete the larger project at hand.

Closes #1568; should add a more detailed issue to address the problems directly above.

### DIFF
--- a/web/main/create_search_index.sql
+++ b/web/main/create_search_index.sql
@@ -33,7 +33,6 @@ UNION ALL
                'effective_date', effective_date,
                'effective_date_formatted', TO_CHAR(effective_date, 'Month FMDD, YYYY'),
                'citations', array_to_string(c.citations, ', '),
-               'content', coalesce(content),
                'jurisdiction', jurisdiction
            ) AS metadata,
            'legal_doc_fulltext'::text AS category
--- a/web/main/create_search_index.sql
+++ b/web/main/create_search_index.sql
@@ -21,6 +21,26 @@ CREATE MATERIALIZED VIEW internal_search_view AS
         main_legaldocument c
     GROUP BY c.id
 UNION ALL
+    -- seperate category for full-text search
+    SELECT
+           row_number() OVER (PARTITION BY true) AS id,
+           c.id AS result_id,
+           (
+                setweight(to_tsvector('english',coalesce(c.content, '')), 'D')
+            )  AS document,
+           jsonb_build_object(
+               'display_name', coalesce(c.short_name, c.name),
+               'effective_date', effective_date,
+               'effective_date_formatted', TO_CHAR(effective_date, 'Month FMDD, YYYY'),
+               'citations', array_to_string(c.citations, ', '),
+               'content', coalesce(content),
+               'jurisdiction', jurisdiction
+           ) AS metadata,
+           'legal_doc_fulltext'::text AS category
+    FROM
+        main_legaldocument c
+    GROUP BY c.id
+UNION ALL
     -- via app/models/content/casebook.rb
     SELECT
         row_number() OVER (PARTITION BY true) AS id,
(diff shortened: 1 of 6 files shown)


web/main/models.py
```diff
@@ -880,6 +883,53 @@ def _search(cls, category, query=None, page_size=10, page=1, filters={}, facet_f
...
+        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 2', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'content': 'Dubious legal claim 2'}
+        ...     ],
+        ...     {'legal_doc_fulltext': 1, 'user': 1, 'casebook': 1, 'legal_doc': 1},
+        ...     {}
+        ... )
+        """
+        if isinstance(casebook, int):
+            casebook = Casebook.objects.get(id=casebook)
+        if not isinstance(casebook, Casebook):
+            raise ValueError("param casebook is not an integer id or a Casebook object!")
```
lizadaly: `casebook` should only be one kind of thing here—I'd even recommend considering adding type hints that could enforce that in the future.

I recommend making the function parameter `casebook_id: int` and not supporting the case where the parameter could be a casebook object at all.

web/main/models.py
```diff
@@ -880,6 +883,53 @@ def _search(cls, category, query=None, page_size=10, page=1, filters={}, facet_f
...
+         
+        legal_doc_ids = [node.resource.id for node in casebook.contents.filter(resource_type="LegalDocument").prefetch_resources()]
```
lizadaly: I don't think this needs to change, but I suspect if you only need the ids it would be faster/less memory to implement this query yourself directly using something like https://docs.djangoproject.com/en/3.2/ref/models/querysets/#values-list

web/main/create_search_index.sql
```diff
@@ -20,6 +20,26 @@ CREATE MATERIALIZED VIEW internal_search_view AS
     FROM
         main_legaldocument c
     GROUP BY c.id
+UNION ALL
+    -- seperate category for fts
```
lizadaly:
```suggestion
    -- separate category for full-text search
```

varun-magesh:
Oh no --- I totally mucked this PR by not checking out of the black PR properly. Let me repair this mess :sweat_smile:

varun-magesh:
Okay, this should be fixed now! All of Liza's feeedback has been incorporated.

lizadaly (approved):
I reviewed this branch locally and it works great—the URL is available if you know it but not linked from the UI.

PR #1591, opened 2022-07-05
Add `black` for consistent code formatting

### LINKED ISSUE #1587: Add `black` for consistent code formatting
Add `black` to reduce the number of non-material differences in PRs, and to improve readability through uniformity. This will be especially helpful as we increase the number of contributors.

Suggest starting with no configuration, just defaults.

* Reformat the entire codebase as part of a single PR.
* Pin the current version of `black` in the dependency list (it's important every developer be on the same version)
* Add `black` to the Github Actions linting workflow

varun-magesh:
closes #1587 

- adds `black` to requirements
- runs black on all Python files
- adds black lint checks to gh-actions.


lizadaly:
Awright!

`flake8` and `black` handle different concerns, so I don't think we want to remove `flake8`, just add `black`.

varun-magesh:
@lizadaly  Applied the change and then updated it by merging and re-running black on the current develop branch!

varun-magesh:
Fixed the requirements issue, re-applied black to the latest harvard-lil:develop to avoid merge conflicts.

.github/workflows/black.yml
```diff
@@ -0,0 +1,10 @@
+name: Black
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
```
lizadaly, line 10: This version needs to match too. Docs seem to imply this will work:

```
- uses: psf/black@stable
  with:
    version: "22.6.0"
```

jcushman: Could also consider the approach we use in lints.yml to pull the current version from requirements.txt.

web/main/urls.py
```diff
@@ -6,127 +6,330 @@
...
-register_model_converter(ContentAnnotation, 'annotation')
+register_model_converter(ContentNode, "resource")
+register_model_converter(ContentAnnotation, "annotation")
 register_model_converter(LegalDocument)
-register_model_converter(SavedImage,name='image_uuid', base=UUIDConverter, field='external_id')
+register_model_converter(
+    SavedImage, name="image_uuid", base=UUIDConverter, field="external_id"
+)
 
 # these patterns will have optional format suffixes added, like '.json'
```
rebeccacremona, line 25: Do we want to `fmt: off` for the stanzas declaring paths? I feel like it is significantly easier to read if you can skim the list for the route you are looking for, without being interrupted by other args when things get long, like in the reformatted
```
    path(
        "resources/<resource:resource>/annotations/<annotation:annotation>",
        views.AnnotationDetailView.as_view(),
        name="annotation_detail",
    ),
```
Just a thought.

docker/pandoc-lambda/function/app.py
```diff
@@ -280,67 +412,86 @@ def section_break(destination, frontmatter=False, chapter=False, pre_chap=False)
...
         # get the source html
-        s3 = boto3.resource('s3', **s3_config)
-        s3.Bucket(os.environ['EXPORT_BUCKET']).download_fileobj(input_s3_key, pandoc_in)
+        s3 = boto3.resource("s3", **s3_config)
+        s3.Bucket(os.environ["EXPORT_BUCKET"]).download_fileobj(input_s3_key, pandoc_in)
         pandoc_in.seek(0)
 
         # convert to docx with pandoc
-        with tempfile.NamedTemporaryFile(suffix='.docx') as pandoc_out:
+        with tempfile.NamedTemporaryFile(suffix=".docx") as pandoc_out:
```
rebeccacremona, line 464: I think this (332-337 in the original) might be another good spot for `fmt: off`: IMO the construction of the command is much easier to read with the original formatting.

web/config/context_processors.py
```diff
@@ -21,35 +21,46 @@ def settings(request):
             raise ImproperlyConfigured(m)
     return new_settings
 
+
 _frontend_urls = None
+
+
 def frontend_urls(request):
     def munge_url(urlpat):
```
rebeccacremona, line 29: Same here, maybe? IMO the original replacements dict formatting, in this particular case, while idiosyncratic, is better...

web/main/urls.py
lizadaly: Personally I don't customize `black`'s output any and just let it do what it will, so no opinion here from me!

web/main/migrations/0003_alternate_legal_docs.py
```diff
@@ -12,101 +12,212 @@
...
-        ('main', '0002_auto_20201021_1444'),
+        ("main", "0002_auto_20201021_1444"),
     ]
 
     operations = [
         migrations.CreateModel(
-            name='LegalDocumentSource',
+            name="LegalDocumentSource",
             fields=[
-                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
```
rebeccacremona: Having Black rewrite all the Django migration files also strikes me as kind of weird, and IMO makes the migration files less readable. Maybe we should skip that directory?

web/main/urls.py
rebeccacremona: 🤷‍♀️ If that's the best practice, okey dokey. This may just be something for me to get used to 🙂

matteocargnelutti: Just FYI, here are the `black` settings I've been using recently:

```toml
[tool.black]
line-length = 100
target-version = ['py310']
```

I find that setting the line length to 100 gives a good balance, but that's a personal preference.

.github/workflows/black.yml
varun-magesh: Arg, sorry --- I totally forgot about this, let me make that fix.

web/main/urls.py
rebeccacremona: Would we be willing to give 100 a try at first? Can easily switch subsequently to the default 88 if its grinding people's gears. 

But seriously, don't let me be a pain in the butt here 😁

lizadaly: Totally!

varun-magesh:
I applied the version change, created a config file with 100 line length, and re-applied black to all files. I can also add exceptions wherever needed, although it seems like we're still talking about that.

web/main/migrations/0003_alternate_legal_docs.py
lizadaly: Skipping the whole migration directory makes sense to me!

varun-magesh:
This has been updated with:

- line length=100
- Actions and requirements using the same black version
- exclude migrations directory

lizadaly (approved):

Issue #1363, opened 2021-05-27
Case-Insensitivity for casebook collaboration invitations


holtchesley:
There's some roughness around case sensitive usernames in collaborator invites. Smooth this out, possibly by making it easy to merge accounts.

lizadaly:
Though email addresses today are almost universally case-insensitive throughout, according to the [RFC](https://www.rfc-editor.org/rfc/rfc5321.txt) the username portion of the address is case-sensitive, and Django enforces this.

However, in all practical cases, email addresses are treated as case-insensitive these days and for an app of our scope, I think it's safe to follow common practice rather the original RFC.

There is a complicated "correct" solution involving switching the email address field type to the Postgres-only [CIText](https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/fields/#citext-fields) type, which does case-insensitive retrieval, but that involves enabling extensions on the Postgres side and I think is just not worth the risk and complexity.

Depending on the most common actual use case we can mitigate this a few ways:

* If we think the primary problem is people creating invites in mixed case but the invitee attempting to log in with lower case, we can force lower case all new collaborator invitations. This will make all invitations going forward lowercase-only.

* We could do a case insensitive search for existing accounts in the [Invite form](https://github.com/harvard-lil/h2o/blob/develop/web/main/forms.py#L303) to avoid creating duplicate accounts with different cases. This should be safe to do.

* Newly-invited collaborators hit the reset-password flow first, to set their password. They aren't actually shown which email address was used because normally in this flow the user has already entered their own email. If they're the kind of person who forwards send multiple addresses to the same mailbox, they won't know which address was used without carefully inspecting the invite. This is the same problem as differing case, but probably harder to handle via support. I don't know how common this is.

* If we think it's safe to ignore the spec and assume all email addresses should be case-insensitive, we could do a mass data migration to downcase all addresses, then make all login forms case-insensitive. I think this is probably also overly aggressive.

My recommendation is to do the first two items, which should do little harm, and also generate a small report on how many mixed-case email addresses we have. It may also help to give users following the reset password flow the opportunity to update their email address.

lizadaly:
We're doing a case-insensitive check at invitation time now, but @cath9 and I agree we should just go ahead and force-lower all new incoming email addresses at account creation. If someone tries to log in with mixed case that'll fail unless we also lower-cased all existing legacy addresses first. Given the number of accounts I think it's probably fine to just handle those as needed.

PR #1592, opened 2022-07-05
[#1363] Check case-insensitively for existing collaborators before creating new accounts

lizadaly:
In #1363 we'd like to reduce the number of accidental duplicate collaborator accounts created because people create invites for addresses that technically already exist but in a different mix of upper and lower case.

This PR uses case-insensitive evaluation to check whether a user already exists. Prior to this inviting LIZADALY@gmail.com when lizadaly@gmail.com already existed would create a second user with the all-caps username. This is almost certainly always undesirable; see #1363 for analysis.

This doesn't address duplicates that already exist and it doesn't force-downcase the whole email address, though we might choose to do that coupled with regularizing all the existing addresses. If that does happen some of this code can go away.

### DIFF
--- a/web/main/forms.py
+++ b/web/main/forms.py
@@ -350,27 +350,28 @@ class CollaboratorFormSet(forms.BaseModelFormSet):
 
 
 class InviteCollaboratorForm(forms.Form):
-    casebook = forms.IntegerField(widget=forms.HiddenInput())
-    email = forms.EmailField()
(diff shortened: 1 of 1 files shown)


lizadaly:
Lint error see https://github.com/harvard-lil/h2o/issues/1593 (I used a walrus operator)

matteocargnelutti (approved):
Makes a lot of sense to me 🚀

PR #1596, opened 2022-07-06
Upgrade flake8

### LINKED ISSUE #1593: Upgrade flake8 to match our Python version
Our version of flake8 is currently pinned at `flake8==3.7.8`, which only supports up to Python 3.7 and is from 2019. We're on Python 3.9!

Upgrade to flake8 4.0.1, the current version.

lizadaly:
Fixes https://github.com/harvard-lil/h2o/issues/1593

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -12,5 +12,5 @@ jobs:
       - name: Flake8
         run: |
           pip install `egrep -o 'flake8==\S+' web/requirements.txt`  # install our version of flake8
-          flake8 web/
-          flake8 docker/pandoc-lambda/function/
+          flake8 web/ --config web/setup.cfg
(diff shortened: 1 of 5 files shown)


lizadaly:
Makes sense to leave this until after the `black` PR merges since the new version is so mad about line length.

lizadaly:
This has some unrelated formatting changes in the lambda function because I told my editor to start auto-blacking, then I fixed a linter issue in there. It was probably wrong to exclude that directory (my bad, I didn't realize there was python in there!). I think it's fine to just commit this change otherwise someone else will probably do the same thing eventually. We can optionally also modify the black config in a followup.

bensteinberg (approved):
LGTM. We might wait to merge this until I finish fussing with the failing build from merging #1591.

Issue #1045, opened 2020-01-10
TextBlock legacy fields


rebeccacremona:
- `created_via_import` is never True..... can we drop this field?
- `public` what does this mean? we have no logic related to public/private TextBlocks. There are some 110 TextBlocks w/ public=False
- About 500 old TextBlocks have non-empty descriptions:
```
>>> TextBlock.objects.filter(description__isnull=True).count()
7186
>>> TextBlock.objects.filter(description__isnull=False).count()
2962
>>> TextBlock.objects.filter(description='').count()
2486
```
Do we want to display those, somehow? If not, let's delete this column.

lizadaly:
Re: description:

106 public casebooks contain TextBlocks with description fields:

```sql
select
    id,
    created_at,
    updated_at
from
    main_casebook
where
    id in (
        select casebook_id
        from
            main_textblock
        inner join
            main_contentnode on resource_id = main_textblock.id
        where
            resource_type = 'TextBlock' and nullif(description, '') is not null
    ) and state = 'Public'
order by id;
```

```
  id  |          created_at           |          updated_at           
------+-------------------------------+-------------------------------
   95 | 2018-01-04 09:34:07+00        | 2020-03-31 12:38:48.856416+00
  147 | 2020-03-29 01:11:03.930798+00 | 2020-03-31 12:38:49.788927+00
  163 | 2019-12-16 12:18:55.767769+00 | 2020-03-31 12:38:48.947713+00
  185 | 2018-01-18 15:56:48.64959+00  | 2022-02-08 22:10:51.41532+00
  195 | 2015-01-21 03:46:05+00        | 2022-01-30 13:30:04.041631+00
  237 | 2020-02-08 00:24:42.426861+00 | 2021-09-23 10:45:10.482601+00
  261 | 2013-08-15 15:33:37.478562+00 | 2020-09-26 12:03:49.314824+00
  316 | 2016-02-11 00:34:13.514555+00 | 2020-03-31 12:38:49.086377+00
  343 | 2018-06-19 21:17:04.816859+00 | 2020-10-01 20:51:47.75784+00
  351 | 2018-06-21 18:52:11.109641+00 | 2020-05-22 17:29:01.02942+00
  357 | 2017-12-23 03:20:25.762747+00 | 2020-09-26 12:08:12.909274+00
  360 | 2018-08-28 15:10:33.298621+00 | 2020-03-31 12:38:49.836076+00
[...]

PR #1597, opened 2022-07-06
Drop unused created_via_import field in textblock

lizadaly:
One part of #1045

Confirmed this has no code references and is always false:

```
postgres=# select distinct created_via_import from main_textblock;
 created_via_import 
--------------------
 f
(1 row)
```

Confirmed the admin page for this still loads.


rebeccacremona:
Ah, looks like the test fixture needs tweaking too: https://github.com/harvard-lil/h2o/blob/07fda42d21f6e1294a11554bd3e107b97ccb8f6d/web/conftest.py#L179

lizadaly:
Fun fact: you need to save changes in your editor!

lizadaly:
(I thought I was losing my mind because VSCode's search was showing it as gone, but I guess it doesn't care whether the file was saved.)

rebeccacremona (approved):

PR #1598, opened 2022-07-07
Drop `public` field on textblock

lizadaly:
Another part of [#1045](https://github.com/harvard-lil/h2o/issues/1045)

Drops the `public` field, which is unused.

This field defaults to `true`, but somehow ended up `false` in 341 `TextBlock` items (out of 143,611). Strangely, some of those are recent:

```
postgres=# select id, created_at, updated_at from main_textblock where public is false order by created_at desc;
   id   |         created_at         |         updated_at         
--------+----------------------------+----------------------------
 207127 | 2022-05-16 12:25:32.707802 | 2022-05-16 12:25:32.707807
 207083 | 2022-05-16 12:25:32.706861 | 2022-05-16 12:25:32.706865
 201739 | 2022-04-01 15:39:44.044836 | 2022-04-01 15:39:44.044839
 201738 | 2022-04-01 15:39:44.044817 | 2022-04-01 15:39:44.044821
 201735 | 2022-04-01 15:39:44.044761 | 2022-04-01 15:39:44.044765
[...]
```

Here's the breakdown by year:

```
postgres=# select extract(year from created_at), count(*) from main_textblock where public is false group by extract(year from created_at);
 date_part | count 
-----------+-------
      2012 |     3
      2013 |    10
      2014 |     8
      2015 |     1
      2016 |    19
      2017 |     1
      2018 |    29
      2019 |    33
      2020 |   104
      2021 |   111
      2022 |    22
(11 rows)
```

I looked at one of the Casebooks containing a non-public TextBlock and there's no difference—even using a non-admin, non-owner user I could still see the item, so I'm pretty sure this has no effect, but I can't explain why some items got created with the false value other than using the Django admin, which seems unlikely.


rebeccacremona:
Cool: I think all the recently-created TextBlocks that have `public=False` are clones of old ones. For instance, I just cloned https://opencasebook.org/casebooks/237-an-introduction-to-the-law-of-corporations-cases-and-materials/, resulting in the creation of [217040](https://opencasebook.org/django-admin/main/textblock/217040/change/?_changelist_filters=o%3D2), [217039](https://opencasebook.org/django-admin/main/textblock/217039/change/?_changelist_filters=o%3D2), and [217036](https://opencasebook.org/django-admin/main/textblock/217036/change/?_changelist_filters=o%3D2).

rebeccacremona (approved):

lizadaly:
Oh nice, I should've thought of checking that.

PR #1601, opened 2022-07-07
Make casebook text resources searchable

### LINKED ISSUE #1594: Make casebook Text resources searchable
Full-text search only searches through LegalDocuments at the moment. We would also like to enable search through the text resources of cases.

See also PR #1590 and Issue #1568

- [x] Index text resources in the materialized view
- [x] Write joins to pull up text resources associated with a specific casebook
- [x] Modify `SearchIndex.casebook_fts` to return text results
- [x] Modify templates/search/results.html to show a text resource as a search result
- [x] Modify templates/search/casebook.html to reflect fts for legal docs and text

varun-magesh:
Casebook text resources are now searchable. This closes #1594.

I can feel the text search getting noticeably groggy. If you check out this branch, let me know if you experience it too. If so, this is probably something we need to discuss.

The search index went from ~12,000 records to ~162,000, so a little grogginess would certainly be unsurprising.

### DIFF
--- a/web/main/create_search_index.sql
+++ b/web/main/create_search_index.sql
@@ -40,6 +40,26 @@ UNION ALL
         main_legaldocument c
     GROUP BY c.id
 UNION ALL
+    -- seperate category for full-text search of textblocks
+    SELECT
+           row_number() OVER (PARTITION BY true) AS id,
+           t.id AS result_id,
+           (
+                setweight(to_tsvector('english',coalesce(t.content, '')), 'D') ||
+                setweight(to_tsvector('english',coalesce(t.name, '')), 'A') ||
+                setweight(to_tsvector('english',coalesce(t.description, '')), 'C')
+            )  AS document,
+           jsonb_build_object(
+               'name', t.name,
+               'description', t.description,
+               'ordinals', array_to_string(cn.ordinals, '.')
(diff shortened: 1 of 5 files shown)


varun-magesh:
It looks like the test failure is in `docker-compose up`? I'm not able to reproduce this locally.

lizadaly:
I think this was from @bensteinberg fixing the registry—I re-started the job and it seems to be past that now.

varun-magesh:
thank you!

lizadaly (approved):
Tried this out locally and it looks terrific. 

I'm not sure if we'll need to run the create-index process specifically, or if the process that triggers `REFRESH MATERIALIZED VIEW CONCURRENTLY internal_search_view` every 5 minutes via will do the same thing.

varun-magesh:
The first, legal-document iteration seemed to work correctly on staging without further intervention (as far as I know). 

> I'm not sure if we'll need to run the create-index process specifically, or if the process that triggers `REFRESH MATERIALIZED VIEW CONCURRENTLY internal_search_view` every 5 minutes via will do the same thing.

The first iteration with just legal documents seemed to work correctly on staging without further intervention.

I should also note that creating the index is now also a non-trivial feat that takes several minutes on my old crappy computer.

PR #1604, opened 2022-07-07
Fixed search category persistence issue #1603

### LINKED ISSUE #1603: Search results return to default category on new search
**Current Behavior**
When searching the site or a single casebook, search results are displayed in separate categories. If I search for 'law stuff', hits from Casebooks (or Legal Documents, for full-text) are shown by default, and I can see the counts for other hits and navigate to them by clicking on the top tab. If I navigate to "Users" and then search for "Zittrain," the search screen throws away `type` information and returns to a search for "Zittrain" in casebooks.

**Desired Behavior**
If a search is started from a specific category, then results from that category should be shown by default.

This isn't really that big of a deal in site-wide search, but it's really annoying in full-text search.

varun-magesh:
This closes #1603.

WARNING:
I'm not actually sure if this is prudent. This removes the plurals from `type` in search arguments. Inconsistent use of plurals and a hard coded `s` added to convert between the two in Django's HTML templater is causing this problem. Were the plurals in `type` there for a reason? If so, this PR should be reworked. I wasn't able to find any other references to this in the code, tests pass, and everything seems to be working fine without the plurals.


lizadaly (approved):
I don't have any context into how this got the way it was, but I believe it's a positive change! (The conflicts will have to be addressed before merging.)

varun-magesh:
Fixed the merge --- @rebeccacremona @bensteinberg do either of you have any context on this?

bensteinberg:
I don't have any context on this, so with any luck @rebeccacremona will....

rebeccacremona:
I think "type" was plural in the Rails app. We were running both apps side-by-side and cutting over route by route as things were ready, so, we paid more attention to making URLs perfectly identical than was probably necessary. (E.g. why use "type" instead of "category" as the GET param at all?)

I do not know if https://github.com/harvard-lil/h2o/issues/1603 was always the case, or if something broke at some point.

web/main/views.py
```diff
@@ -2416,7 +2416,7 @@ def search_using(request, source):
     return JsonResponse({"results": results}, status=200)
 
 
-type_param_to_category = {'legal_doc': 'legal_doc', 'casebooks': 'casebook', 'users': 'user', 'legal_doc_fulltext': 'legal_doc_fulltext', 'textblock':'textblock', 'link':'link'}
+type_param_to_category = {'legal_doc': 'legal_doc', 'casebook': 'casebook', 'user': 'user', 'legal_doc_fulltext': 'legal_doc_fulltext', 'textblock':'textblock', 'link':'link'}
```
rebeccacremona: Since now the keys and values are identical, want to remove this mapping?

varun-magesh: I thought there might be a future-reason to want it, but I see now that it makes sense to just chuck it.

rebeccacremona: Ooops, I'm so sorry: I should have said. It also serves to prevent people from being able to supply arbitrary values for category. So if not a mapping, we need at least a list of permitted values.

rebeccacremona (approved):

PR #1609, opened 2022-07-08
Split fulltext and site search indexes

varun-magesh:
This PR splits indexing and searching between site-wide metadata search and fulltext search of casebooks. This is expected to address indexing speed issues on staging.


lizadaly:
I think the test harness needs the index process run at least once, judging from the errors?

varun-magesh:
I think the setup in the doctests explicitly create the search index --- it must be failing for some reason; see e.g. models.py:933
```python
         >>> FullTextSearchIndex().create_search_index()
```

varun-magesh:
Also --- _search is failing but casebook_fts (which depends on search) passes... huh.

varun-magesh:
Oh I'm silly! nothing to do with indexing, just forgot to update test cases :-1:

bensteinberg (approved):
LGTM. I don't think I can comment in detail on the SQL or whether/how to do the inheritance. My proofreading eyes want to change "seperate" to "separate" several places in the SQL comments. Let's discuss elsewhere what to do with the existing cron job on stage before deployment.

varun-magesh:
I just looked up 'seperate' and 'separate' and had no idea they were different --- oops! I probably made that error in a million undergrad papers.

Issue #1574, opened 2022-06-24
Refresh common usage reporting tables on a schedule


lizadaly:
Temp tables are generated as part of the admin-only usage dashboard; they are designed to make it possible to run performant queries like "how many casebooks published by professors contain CAP content between these two dates?" These reporting tables take a few chunky seconds to build and happen on each request.

In #1563 we'd like these counts to be expanded out into full lists, in which case the query needs to be run again but the results materialized into Django objects. 

Pre-building commonly-used reporting tables would have a few benefits:

* The usage dashboard page would load very very fast.
* Some queries that were too expensive to put into the request cycle (like one that searched text content for multimedia links) could now become feasible when generated offline.
* The full result lists pages would now have that table still available to query from and would also be fast.

Evaluate creating materialized views, to be refreshed on a cron schedule of every hour or so. Instead of: 

```sql
create temp table if not exists casebooks_with_multiple_collaborators as
            select casebook_id from main_contentcollaborator
            inner join main_casebook c on main_contentcollaborator.casebook_id = c.id
            where c.created_at >= %s
                and c.created_at <= %s
                and c.state in %s
            group by casebook_id
            having count(user_id) > 1
```

where the table is built pre-filtered by date and publication state, end up with a view like:

```sql
create materialized view  casebooks_with_multiple_collaborators as
            select casebook_id, c.created_at, c.state from main_contentcollaborator
            inner join main_casebook c on main_contentcollaborator.casebook_id = c.id
            group by casebook_id
            having count(user_id) > 1
```

add indexes for `created_at` and `state`, and then delegate the date/state filtering to the individual queries.

PR #1611, opened 2022-07-08
Refresh common usage reporting tables on a schedule 

lizadaly:
Part of [#1574](https://github.com/harvard-lil/h2o/issues/1574)

This switches the reporting tables for the Django admin usage page from "temp tables created on each request to the admin usage page" to "materialized views that are run out of band and repopulated on a schedule." The behavior is identical except that the reporting tables aren't updated in real time (probably fine!) but the admin page now loads much much faster.

I'm expecting the run cadence to be nightly, and triggered by a cron job populated elsewhere.

[Removed Django template behavior here after discussion.]

I think to start, the cron job PR (which I can make in the related repo) can just call `fab create_reporting_views` every time, as the views build from scratch quickly enough and it only needs to be run once a day. If we want it to run more often or start adding tables that take longer to build, we could switch to the refresh task and run the rebuild on-demand.


web/main/reporting/create_reporting_views.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+
+def create() -> None:
+    """Create a little data warehouse of prefiltered reporting-ready tables for reuse in this session.
+    This logic should match what's used in `create_search_index.sql` unless otherwise annotated."""
+
+    with connection.cursor() as cursor:
+
+        sql = render_to_string(
```
matteocargnelutti: **Question:** Does `render_to_string()` offer any protection against SQL injections in this context _(what kind of escaping does it perform)_? 

If not, my _slight_ concern here would be that future, more complex iterations of this reporting tool could use more   variables to customize `reporting.sql.j2`, which would slowly become more vulnerable?

matteocargnelutti (commented):
This makes a lot of sense to me and I can see how this is a nice improvement 😄 .

LGTM, but I think that'd be nice if either @bensteinberg or @rebeccacremona had a second look, as they are much more familiar with H2O's codebase and history than I am.

web/main/reporting/create_reporting_views.py
lizadaly: It does not! https://github.com/sripathikrishnan/jinjasql  does. On the other hand, this kind of async job cannot take user parameters since there's no entry point. On the third hand, it does set up a potential pattern where that could be true.

lizadaly: I can definitely swap it out with just a plain text file if that would feel safer.

matteocargnelutti: > On the other hand, this kind of async job cannot take user parameters since there's no entry point. On the third hand, it does set up a potential pattern where that could be true.

Yes, I think we're on the same page 😄 

---

> I can definitely swap it out with just a plain text file if that would feel safer.

Alternatively, here's a suggestion:
- We could add a comment around `sql = render_to_string(...)`to signal that there's a slight risk here _(so future us / somebody else is mindful of what is added and how)_
- We could also add a quick format check in  `{% for view in views %}` to make sure `view` is a string matching `"this_format"` ?

lizadaly: Eh, I'd rather not introduce complexity and readability challenges just to save a couple slightly redundant lines. I'll switch this to plain text and if we want to introduce more complex SQL templating we can do that with a purpose-built library. Thanks for raising the issue!

matteocargnelutti (approved):

PR #1613, opened 2022-07-08
Stores `casebook_id` in search metadata instead of join-ing

varun-magesh:
TextBlock and Link resources form the vast, vast majority of our search index. Unlike LegalDocuments, they have a one-to-one relationship with casebooks. In order to filter down to the TextBlocks present in a single casebook, we were previously using the same strategy we used for LegalDocuments: get a list of PKs from Casebook.contents, and then do an `id__in` filter against this list of primary keys.

This is necessary for LegalDocuments, which have a one-to-many relationship with casebooks. However, for other resources, we can pack casebook_ids directly into their search metadata and save us this join step. This speeds search up considerably, because we previously had to perform expensive IN lookups.

### DIFF
--- a/web/main/create_fts_index.sql
+++ b/web/main/create_fts_index.sql
@@ -32,7 +32,8 @@ UNION ALL
            jsonb_build_object(
                'name', t.name,
                'description', t.description,
-               'ordinals', array_to_string(cn.ordinals, '.')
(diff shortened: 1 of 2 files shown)


bensteinberg (approved):
Great!

PR #1614, opened 2022-07-08
Modifies indexing process to build a new index before chucking the old one

### LINKED ISSUE #1610: Full-text search indexing takes too long
**Problem**
Search indexing is currently done by storing `ts_vectors` of searchable information (content, names, etc.) in a Postgres Materialized View. The as I understand it, a materialized view is essentially a persisted query, as per https://www.postgresql.org/docs/current/rules-materializedviews.html. Its benefit is that it can store the result of aggregations and joins so that they can be accessed again without hitting the underlying tables later, which brings significant performance benefits. It does so without actually replicating any of the data from the db into more tables, which is a good thing to avoid so you don't end up with data mismatch issues and the like.

Materialized views cannot be updated; they can only be created or dropped.

If a materialized view is not updated with some frequency, then it may no longer reflect the underlying content and become stale.

We used to update indexes every five minutes, but updating a full-text index takes nearly an hour, and possibly hangs searches during update, which is a problem.

**Solutions**
I can see two solutions for this:
1. Solve the index-hanging problem by using an old index until the new one is built, and just update the search index less frequently. 
        **PROS** less developer time
        **CON** still takes an hour on staging, no search for a day after casebook update.
2. Without the `ts_vector` creation, updating the index takes seconds. We could cache `ts_vectors` in a table and build the materialized view by drawing from these cached `ts_vectors`. The `ts_vectors` are then updated as necessary when content is updated.
         **PRO** fast, simple, search live soon after content change.
         **CON** need to create content-update hooks or similar, possible mismatch/updating issues

varun-magesh:
This PR modifies the indexing process to build a new index, delete the old one, and then rename the new to hotswap in for the old.  This process makes sure that search does not hang even if the index needs to be recreated for whatever reason. This also creates two new `fab` commands, `create_fts_index` and `refresh_fts_index` that should help manage the indexing process.

NOTE: if we actually want to suspend searches for some reason (e.g., migration makes search incompatible), then we need to manually drop the index view.

This closes #1610.

### DIFF
--- a/web/fabfile.py
+++ b/web/fabfile.py
@@ -66,6 +66,19 @@ def refresh_search_index():
     from main.models import SearchIndex
     SearchIndex.refresh_search_index()
 
+@task
+@setup_django
+def create_fts_index():
+    """ Create (or recreate) the search_view materialized view """
+    from main.models import FullTextSearchIndex
+    FullTextSearchIndex.create_search_index()
+
+@task
+@setup_django
+def refresh_fts_index():
+    """ Update an existing search_view materialized view; will create if create_search_index hasn't been run once """
+    from main.models import FullTextSearchIndex
+    FullTextSearchIndex.refresh_search_index()
 
 @task
 @setup_django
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
:+1:

PR #1616, opened 2022-07-11
Speeds up full-text search by removing metadata `casebook_id` filters and removing count displays

varun-magesh:
This PR reduces full-text search speeds from ~1s to ~.2s. This is achieved by not filtering on metadata terms (surprisingly expensive!) and by removing count displays, which force the query and an aggregation function to be run against all categories, doing much more work than immediately necessary.

This PR was made possible by the Django debug toolbar! Thanks @rebeccacremona :)

2022-07-11-113211_1692x553_scrot

2022-07-11-131206_1687x383_scrot
The slowest part is currently headline generation. We could improve on this by delivering results first and then filling in headlines as they load, or perhaps by caching content in the search index metadata ---- although this would come as a memory cost.


rebeccacremona (approved):
LGTM!

PR #1621, opened 2022-07-12
Calculate and add read time estimate to UI

varun-magesh:
This pull requests estimates the read time for a casebook resource. This read time is added to the UI in (what I think is) a non-obtrusive way.

A future PR will build on this work to display estimates for whole sections or casebooks in the interest of helping professors gauge the volume of their casebook.

2022-07-12-163245_850x254_scrot


lizadaly:
This rules! I think it looks nice too.

I have a few testing suggestions:

Run through a few (5 or so?) different casebooks + resource types and eyeball the estimates. Do they look reasonable?

Write a little testing program using your local database as a datasource and calculate the reading times for every content node, outputting it to a text file (like a csv). Something like this:

```
for node in ContentNode.objects.all():
  print (f"{node.id),{node.reading_time()}")
```

and get a sense of the distribution of values—are a lot of them `None`? Does any node throw an exception?

This wouldn't be a test you'd commit to the repo, just something to validate against live data since we happen to have it easily accessible.

varun-magesh:
sans_html

In general, things look pretty reasonable --- this is a distribution of 1000 nodes.

One drawback to our current approach is that it requires  HTML tags to be removed from every node's content, which is expensive. I tested out a version which didn't bother stripping HTML (but did apply elisions) so that it could be run with a couple of SQL queries, but this version ended up being very inaccurate quite frequently, so we will probably need to cache these results somewhere or be okay with slightly slower load times on casebook pages.

lizadaly:
Yay graphs!

I think ultimately we'll want them precomputed for a lot of reasons, not the least of which would be to display a total reading length for the whole work. But I think this is a terrific feature and unless you're notably concerned with request load time, we can start with on-demand and then maybe add a database field that gets updated any time the content is saved.

lizadaly (approved):

varun-magesh:
Ultimately, this PR shows an estimated reading time + num links for sections and individual resources. It does not make casebook-level estimates in an effort to trial the feature before deciding on how to expand it best. However, it does provide sufficient backend work to make any display pretty easy.

varun-magesh:
2022-07-14-113128_1415x868_scrot


note: I, too, was skeptical of an entire section of this book only taking 20 minutes to read. However, it turns out that most of the cases within are heavily elided, down to just a paragraph or two, so this number makes sense.

web/main/templates/includes/breadcrumbs.html
```diff
@@ -11,5 +12,13 @@
         <span class="separator">.</span>
       {% endif %}
     {% endfor %}
+    {% if content.reading_time %}
+      <div class="reading-time">
+        {{ content.reading_time | humanize_minutes }}
+        {% if content.num_links %}
+          + {{ content.num_links }} links
```
lizadaly: What if it's one link? :)

lizadaly:
I like this as an incremental approach and it's nice that you've done enough discovery already to know where to take it next, and have that be easy to implement!

web/main/templates/includes/breadcrumbs.html
varun-magesh: good catch! let me push a fix

varun-magesh:
I'd also appreciate thoughts on the DB persistence approach --- I don't have much experience here, I know my solution works but I'm not sure that it's the paradigmatically correct one.

relevant changes should be in: https://github.com/harvard-lil/h2o/pull/1621/commits/d9eb7d0a6a04d9002f4b7c2882b69060bc3f5730

and https://github.com/harvard-lil/h2o/pull/1621/commits/77c4728f9ee67604d55218d30f40416b953a5f23

web/main/templates/includes/breadcrumbs.html
varun-magesh: should be done!

varun-magesh:
No, the reading time won't be updated on edit --- but it is updated whenever it is needed. It's usually not too expensive --- maybe less than.25s or so on my machine for even the longest content nodes. I will certainly make the change, but I'm not sure I fully understand the benefits of that approach.

varun-magesh:
Hmm, just realized that updating reading time works correctly for text blocks but not eliding legal docs. ;(

lizadaly:
I see that you null out the `reading_length` value on save, I get it now!

lizadaly (approved):

web/main/templates/includes/breadcrumbs.html
```diff
@@ -11,5 +12,15 @@
         <span class="separator">.</span>
       {% endif %}
     {% endfor %}
+    {% if content.reading_time %}
+      <div class="reading-time {% if editing %}reading-time-tooltip{% endif %}" title="Read time estimate is visible to readers">
+        {{ content.reading_time | humanize_minutes }}
+        {% if content.num_links %}
+        + {{ content.num_links }} link{% if content.num_links > 1 %}s{% endif %}
```
lizadaly: Django has a built-in for this!

https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#pluralize

(Though honestly it's barely fewer characters.)

varun-magesh:
@lizadaly pushed changes from our conversation yesterday --- this should be ready to go if everything looks good to you!

lizadaly:
Huzzah!

PR #1629, opened 2022-07-14
Add migration

rebeccacremona:
This goes with the changes from https://github.com/harvard-lil/h2o/pull/1609 (and possibly from subsequent followups).


varun-magesh:
Ah, thank you! I think it worked without the migrations since the materialized views are made with raw SQL anyway, but this is probably good.

lizadaly (approved):

Issue #1579, opened 2022-07-05
Allow adding, replacing, and remove a book cover image


lizadaly:
To differentiate titles, provide a sense of ownership, and support professors in promotion and branding, allow uploading a book cover image.

* Review any previous discussions on book cover images. Is there a reason this feature was not previously implemented?
* Consider copyright implications.
* Allow previewing the book cover during the creation process. 
* Allow admins to remove cover images.

rebeccacremona:
Update on the status of this feature:

I added limited support for adding cover images in #1630 and #1635; Seonghee added support for displaying them in #1664. 

But, the feature is still disabled in production because as [Liza noticed](https://github.com/harvard-lil/h2o/pull/1664#issuecomment-1201652016), I didn't consider drafts/revisions or clones:

- Cover image needs to be added to the [revision merging process](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3894)

- We might want to reconsider the paths at which I'm saving the images, which right now is by casebook ID... if not, we will need to change the image filename during the merging process, so that it has the ID of the published book, not the id of the temporary clone

- I didn't include any logic for clones. I think that means that, if the parent book's cover image gets updated, so does the cover image of all the children. We need to arrange for the file to be duplicated when cloning, so that each book points to its very own file.

PR #1630, opened 2022-07-14
Cover image field

rebeccacremona:
This PR is to support ongoing work on https://github.com/harvard-lil/h2o/issues/1579.

It adds a new field, `CoverImage`, to the Casebook model and adds that field to the Django admin. It reuses the existing minio/S3 storage that is used to store images that are uploaded to text blocks by verified professors, but saves them to a new `cover_image` directory, with the filename set to the id of the casebook + extension.

Currently, no validation is performed on the file before saving: the user-facing upload feature should use `main.utils.validate_image`, probably in a Django form validation step.

### Updated development setup

All developers should update their local `/etc/hosts` as per the README. With that change and this PR's adjustments to Django settings and `docker-compose.yml`, uploaded images and the minio interface can be accessed at `http://opencasebook.minio.test:9000` both from the host/your laptop's web browser and from within our Docker containers. 

(With the status quo, minio is at http://localhost:9000 on your laptop, but is at http://minio:9000 from within a container; all uploaded images' links point to the internal address and thus are broken if viewed from your browser.) 

Developers may now also work with their local instance at `http://opencasebook.test:8000` instead of `http://localhost:8000`, if they prefer.


matteocargnelutti (approved):
Makes a lot of sense to me 🚀 ! Implementation is solid and scope / limitations clearly explained ⭐ .

PR #1633, opened 2022-07-15
Adds search bar tab to casebook view

### LINKED ISSUE #1619: Full-text search casebook link takes you to the legaldocument page
Currently, clicking a case in the full-text search view takes you to the page of the relevant legal document. It should take you to the page of the casebook resource instead, as this maintains the casebook header.

varun-magesh:
2022-07-15-095255_826x456_scrot

Not in love with this look, but opening a draft to solicit opinions.

closes #1619


varun-magesh:
Minimally, I think I would change the headlines to be more like a bluebook citation with the case name bolded, e.g. **RAV v. St. Paul**, 505 U.S. 237 (1976). I might make this bigger, drop the line separators, and make the font of the preview text a little grey?

varun-magesh:
Also I might show the ordinals of the relevant content on the right hand side, and sort it by order in the casebook? Sorting by casebook order might also let us mix link, text, and legal doc. results together here instead of having tabs, which I don't like.

shljessie:
> bluebook citation

I also think it would be nice to highlight( or bold a little more ) the part of the preview text where the search keyword was found!

varun-magesh:
2022-07-15-124832_1251x900_scrot

varun-magesh:
This PR also includes changes in PR #1634 --- the easiest things would be to close #1634 and just merge this when the time comes.

web/main/models.py
```diff
@@ -1060,15 +1057,15 @@ def casebook_fts(
...
-        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 2', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'headlines': ['<b>Dubious</b> legal claim']}
+        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 0', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'headlines': ['<b>Dubious</b> legal claim'], 'ordinals': ''},
+        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 1', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'headlines': ['<b>Dubious</b> legal claim'], 'ordinals': ''},
+        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 2', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'headlines': ['<b>Dubious</b> legal claim'], 'ordinals': ''}
```
rebeccacremona: The smallest of nitpicks: I find it a little confusing that these test docs are nearly identical to each other. Is that helpful for simplicity, making it clear which fields do and do not matter, etc., or would these tests be just as effective if it were more obvious that the documents can be arbitrarily different from each other?

web/main/models.py
```diff
@@ -1060,15 +1057,15 @@ def casebook_fts(
...
         ...     ],
         ... )
 
         >>> assert dump_search_results([FullTextSearchIndex().casebook_fts(casebooks[0].id, 'legal_doc_fulltext', query_str='2'),]) == (
         ...     [
-        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 2', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'headlines': ['Dubious legal claim <b>2</b>']}
+        ...         {'citations': 'Adventures in criminality, 1 Fake 1, (2001)', 'display_name': 'Legal Doc 2', 'jurisdiction': None, 'effective_date': '1900-01-01T00:00:00+00:00', 'effective_date_formatted': 'January   1, 1900', 'headlines': ['Dubious legal claim <b>2</b>'], 'ordinals': ''}
```
rebeccacremona: Similarly: is it legal, with our data model, for an empty ordinal string? `Does 'ordinals': ''` here suggest that a test fixture needs to be updated to have `ordinals: [1]`, such that some resource is the first in a book? Or am I confused?

web/main/templates/includes/casebook_search.html
```diff
@@ -0,0 +1,50 @@
+{% load current_query_string %}
+
+<section class="casebook-search">
+  <header class="advanced-search view-searches-show">
+    <form novalidate="novalidate" class="simple_form search" accept-charset="UTF-8">
+      {# Hidden field to preserve active tab's category during form submission #}
+      {# Note that 'category' is singular, but 'type' is plural #}
```
rebeccacremona: This category/type comment is no longer accurate, right, since your recent PR?

rebeccacremona (approved):
This looks super snazzy!! Love it! Have you taken a look to see how it does w/ mobile screen sizes? Might be worth considering a shorter snippet at smaller sizes, or similar... but even if needed, could be in a later PR. 

LGTM, with the tiniest of suggestions about comments/fixtures.

varun-magesh:
> This looks super snazzy!! Love it! Have you taken a look to see how it does w/ mobile screen sizes? Might be worth considering a shorter snippet at smaller sizes, or similar... but even if needed, could be in a later PR.
> 
> LGTM, with the tiniest of suggestions about comments/fixtures.


Ouch, good reminder:
2022-07-15-134002_336x603_scrot

The results hold up okay, I think, but the tabs at the top get squished.

In an unrelated note, I think this looks good (if I may say so myself), but I'm actually reconsidering it. I think it breaks our model of what a resource "looks like", which is something like this:
2022-07-15-134930_740x72_scrot

what do you think about the results appearing in these little boxes instead?

web/main/models.py
varun-magesh: The appearance of the docs is a consequence of the factory. Honestly, I think these fields don't matter at all --- only the number of results and the highlighting. They've also been a total pain to keep updated as the output of the function changes in tiny tiny ways. I think it makes a lot of sense to check the function output once (all of the fields are there), and then just make sure we get the right number of results in the other cases.

Thanks for pointing this out, let me tweak these tests.

web/main/templates/includes/casebook_search.html
varun-magesh: Ah yes, good catch, this is a holdover.

web/main/models.py
varun-magesh: I believe empty ordinals are permitted; see `SELECT * FROM main_contentnode WHERE main_contentnode.ordinals='{}'`.

varun-magesh:
@rebeccacremona --- I made all of the changes that you suggested. I also scoped out the change I wanted to explore for the results look, but it is all in Vue which I am not too familiar with it yet, so it will be a bigger lift than expected. Perhaps the move is to merge as is (pending your approval), discuss whether it would be improved with the box approach, and address it then?

rebeccacremona:
@varun-iyer Huzzah! LGTM! Personally totally comfortable with this getting merged as is, and if you want to redesign to look more like the way Resources/Sections look in the TOC, to add that later. (Or maybe even the other way around lol.) I'll wait to hit merge until you tell me :-)

web/main/models.py
lizadaly: I'll make a pitch here for a heuristic that if the test is more than just a few lines long, it's probably easier to implement as a standalone `pytest` function rather than a doctest.

lizadaly: (I know you were just following the existing patterns but I find them a little long for doctests, personally!)

varun-magesh:
Merging sounds good to me! Design is probably going to be a bigger conversation so those changes can happen as they get figured out.

web/main/models.py
varun-magesh: yes, I changed them to just ensure that the right number of search results popped up, rather than the full content of the searches!

varun-magesh: That also makes sense, some of these doctests are so long that they interfere with trying to read the function itself (in my opinion)

web/main/models.py
rebeccacremona: FWIW: I think these are exclusively the old `ContentNode`s that used to be the root of casebooks. From my local DB:
```
>>> from main.models import Casebook, ContentNode
>>> 
>>> Casebook.objects.filter(old_casebook_id__isnull=False).count()
823
>>> ContentNode.objects.filter(ordinals=[]).count()
819
```
(dunno what the 4 weird ones are, but oh well)

Doesn't matter for this PR/your project, but good to know.

web/main/models.py
rebeccacremona: Yeah, TBD with @jcushman: I believe this style was intended to be somewhat of an experiment in [literate programming](https://en.wikipedia.org/wiki/Literate_programming), and an experiment in "developers will read the tests more and write more tests if everything is all together in one file"

jcushman: When we set this up we were definitely playing with the upper boundary on doctests -- in particular asking, can the doctests be written compactly enough that they serve the purpose of a docstring but better, convincing you that a function does what it says it does and teaching you how to call it, without being _too_ long? I still think there's some appeal to that, especially with functions that take and return complex data structures where it's hard to write documentation without showing examples anyway.

But! This is totally a matter of taste and I won't be working on the codebase, so I don't want to have an oar in as far as where to draw the line on length and move to a separate pytest function. Carry on!

PR #1635, opened 2022-07-15
Add cover image to casebook edit form, for admins and verified profs.

rebeccacremona:
This PR builds on https://github.com/harvard-lil/h2o/pull/1630, in continued support of work on https://github.com/harvard-lil/h2o/issues/1579.

If the feature flag `COVER_IMAGES` is toggled `True` (default `False`), verified professors and admins may use the `edit_casebook` form to upload a cover image to a casebook. Right now, the built-in Django file field widget is used, without any effort at further styling.

Without an image:
image

With an image:
image

Uploading a cover image otherwise has no effect on the book.

Before an uploaded file is saved, it must pass the same validation step as the images that are uploaded to TextResources or headnotes, `main.utils.validate_image`. If it fails, the form does not save and displays an error message. (Again, I made no attempt to adjust the form's built-in styling for validation errors, or to ensure the validation flow is accessible.)
image

For simplicity, `COVER_IMAGES=True` in settings.pytest.

With this merged, there will hopefully be enough back-end support to start iterating on the design of cover images in earnest!


rebeccacremona:
(Tests failing pending work w/registry)

lizadaly (approved):
Nice!

PR #1636, opened 2022-07-15
Add usage reporting detail pages; fixes #1563

### LINKED ISSUE #1563: Expand counts in usage dashboard to full result lists
image

Will probably need to materialize this as Django objects at the point of retrieving the result set rather than try to recreate the SQL logic in the ORM.

lizadaly:
This enables all of the reporting items on the usage dashboard to be viewed as admin lists that respect the filters applied in the UI to the subsequent list pages.

image

This ends up being complicated because the Django admin is intimately tied to a given database model, but the reporting tools are generated out of views. The recommended way to handle having multiple representations of the same model in the admin is to use Proxy models, which this does. Because of the amount of code involved I split the reporting work into a new Django app, which also conveniently groups the proxy model tables together in the UI:

image

You can use the reporting tables in the admin directly—for example, clicking on `Professors` here will only show you users who are verified professors without needing to subsequently filter. There will certainly be edge cases where specific filters from that point on don't quite work as expected because not all the fields of the original model are replicated in the view. If necessary we can independently exclude some or all of the reporting models from the home page of the admin; a few are already excluded.

There weren't tests for the reporting work before but there are now—the tests don't go deep into the nuances of the SQL but do ensure that the views are all doing generally what's expected.


bensteinberg (approved):
Wow, this is great. I think materialized views do not require migrations?

lizadaly:
They don't, but the proxy models do!

lizadaly:
Which I omitted by accident!

Issue #1615, opened 2022-07-11
Add checking of Python type hints via mypy in the lints Github Action


lizadaly:
Type annotations in Python can be added progressively to a codebase using `mypy` or other tools. Doing so reduces the likelihood of introducing certain kinds of bugs, and adding types to existing code can reveal subtle bugs that already exist. 

It can be introduced using very generous settings to reduce the number of initial type tasks that need to be performed. For example:

```
mypy . --ignore-missing-imports --allow-untyped-globals
```

plus installing some type stubs returns a fairly small number of errors:

```
$ mypy . --ignore-missing-imports --allow-untyped-globals

config/settings/settings_dev.py:45: error: Unsupported target for indexed assignment ("object")
config/settings/settings_pytest.py:18: error: Value of type "object" is not indexable

main/models.py:977: error: Syntax error in type annotation
main/models.py:977: note: Suggestion: Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)
main/models.py:977: error: Name "legal_doc_fulltext" is not defined
main/models.py:977: error: Name "textblock" is not defined
main/models.py:1041: error: "Type[object]" has no attribute "objects"

# Might consider just skipping checking the admin at first
main/admin.py:295: error: "Callable[[CasebookAdmin, Any], Any]" has no attribute "short_description"
main/admin.py:315: error: "Callable[[SectionAdmin, Any], Any]" has no attribute "short_description"
main/admin.py:336: error: "Callable[[ContentNodeAdmin, Any], Any]" has no attribute "short_description"
main/admin.py:359: error: "Callable[[ResourceAdmin, Any], Any]" has no attribute "short_description"
main/admin.py:375: error: "Callable[[AnnotationsAdmin, Any], Any]" has no attribute "admin_order_field"
main/admin.py:379: error: "Callable[[AnnotationsAdmin, Any], Any]" has no attribute "admin_order_field"
main/admin.py:417: error: "Callable[[TextBlockAdmin, Any], Any]" has no attribute "short_description"
main/admin.py:472: error: "Callable[[UserAdmin, Any], Any]" has no attribute "admin_order_field"
main/admin.py:584: error: "Callable[[LegalDocumentAdmin, Any], Any]" has no attribute "short_description"
```

PR #1639, opened 2022-07-18
Add mypy

dongkyunc:
# What

Adds mypy and django-stubs

# Why

Static type checking is nice

# How

- Add package to `requirements.{in,txt}
- Add settings to `web/main/config.
- Add `.github/workflows/mypy.yml`
- Change some lines that were making mypy mad

# Notes to Reviewers

- `web/main/admin.py` is ignored, along with any imports that may affect
- Originally, `check_untyped_defs` was passed in as an option, there were errors I did not know how to fix due to an unfamiliarity with the codebase
- `typing.Sequence` is unused, but it's making mypy mad if it isn't there for some reason.
image


lizadaly:
Thanks for taking this on!

When I looked at this in https://github.com/harvard-lil/h2o/issues/1615 it seemed valuable to install the type stubs for some of the imported packages too.

I recommend deferring the mypy Github Action for now. We probably want to combine all three of pyflakes, black, and mypy into a single workflow to avoid too much overhead in spinning up containers. (There's a Github Action quota for private repositories but that wouldn't seem to apply here.)

web/config/wsgi.py
```diff
@@ -29,6 +29,6 @@ def retrying_send(message, *args, **kwargs):
         return _orig_send(message, *args, **kwargs)
 
 
-EmailMessage.send = retrying_send
+EmailMessage.send = retrying_send  # type: ignore  # Where is this being used?
```
rebeccacremona: This is a monkey patch of Django's base email sending method: whenever Django tries to send email, it gets our function, `retrying_send`, instead of the original method. We are patching it because sometimes emails fail for spurious network reasons, and waiting and trying again in a second succeeds. The patch is all the way here, in wsgi.py, so that we can get in our change before the app is initialized.

Do we need an explicit ignore?

web/main/models.py
```diff
@@ -1071,20 +1076,34 @@ def casebook_fts(
                 result_id__in=textblock_ids
             )
 
+        query_vector: Union[SearchQuery, str]
```
rebeccacremona, line 1076: Let's stick with the original
```
query_vector = SearchQuery(query_str, config="english") if query_str else None
```
and use the type `Optional[SearchQuery]` instead.

Unless an empty string instead of `None` has some other benefit I'm not thinking of?

web/main/models.py
```diff
@@ -1071,20 +1076,34 @@ def casebook_fts(
...
+        # Can replace w/ match statement when upgraded to 3.10
+        query_class: ResourceType
+        if category == "legal_doc_fulltext":
+            query_class = LegalDocument
+        elif category == "textblock":
+            query_class = TextBlock
+        elif category == "link":
+            query_class = Link
+
         content_name = "description" if category == "link" else "content"
```
rebeccacremona, line 1102: Are [these changes](ttps://github.com/harvard-lil/h2o/pull/1639/files#diff-0088ea3fd6dd38dc8969fda2a1559c6de8bab04df5fe137f09208738e9490e97L1080-L1086) related to adding mypy and a few type hints to this file, or is this an unrelated little refactor you decided to do while you were in the file? Do you find this if/else switch more readable than the dict lookup that was there?

web/main/models.py
```diff
@@ -1071,20 +1076,34 @@ def casebook_fts(
...
-        ids_headlines = (
+        ids_headlines_query = (
```
rebeccacremona, line 1103: Is this variable renamed with the appendix `_query` because it's assigned to  a `queryset`? If yes, I'm not sure that's a convention we need to adopt, since we have so many variables assigned to querysets across the codebase, for instance, yours `ids_ordinals_nodes` below).

web/main/models.py
```diff
@@ -1093,16 +1112,17 @@ def casebook_fts(
...
-            ids_ordinals = (
+            ids_ordinals_nodes = (
                 casebook.contents.filter(resource_type="LegalDocument")
-                .filter(resource_id__in=[r.result_id for r in results])
+                .filter(resource_id__in=[r.result_id for r in results_chunk])
                 .values_list("resource_id", "ordinals")
             )
-            ids_ordinals = {i: [str(n) for n in h] for i, h in ids_ordinals}
-        for r in results:
+            ids_ordinals = {i: [str(n) for n in h] for i, h in ids_ordinals_nodes}  # type: ignore
```
rebeccacremona: I'm a little confused about what's happening [here](https://github.com/harvard-lil/h2o/pull/1639/files#diff-0088ea3fd6dd38dc8969fda2a1559c6de8bab04df5fe137f09208738e9490e97L1096-L1124): the new logic doesn't seem the same to me as the original. Am I reading this wrong, or is this in fact different?

`ids_ordinals` is no longer initialized with `None`; it is now defined inside the `if category == "legal_doc_fulltext":` block. Does this code still work without `ids_ordinals = None`, when ` category != "legal_doc_fulltext"`?

Why is `for r in results` no longer necessary? I see that `results` has been renamed `results_chunk`: does this need to be `for r in results_chunk:` instead of deleted?

Added line 1117 adds a type annotation for `ids_ordinals`, but added line 1124 is marked up with ` # type: ignore`. Can you help me understand why that's the case?

web/main/models.py
```diff
@@ -1292,14 +1312,18 @@ def has_bad_footnotes(self):
...
-class ContentAnnotationQueryset(models.QuerySet):
+class ContentAnnotationQuerySet(models.QuerySet):
     def valid(self):
         """
         Return annotations excluding those that were marked invalid when shifting.
         """
         return self.exclude(global_start_offset=-1, global_end_offset=-1)
 
 
+# (2022-07-19) django type-stubs workaround
```
rebeccacremona, line 1326: Interesting. Do you know of any documentation/github issue/etc. explaining why this workaround is necessary?

rebeccacremona (changes requested):
Hooray for adding support for type annotation!

I have a bunch of little questions.

web/main/models.py
dongkyunc: Oops, forgot to add: https://github.com/typeddjango/django-stubs#my-queryset-methods-are-returning-any-rather-than-my-model

Will add to the code

web/config/wsgi.py
dongkyunc: Currently, not ignoring it will run into this error:
```
config/wsgi.py:32: error: Cannot assign to a method
config/wsgi.py:32: error: Incompatible types in assignment (expression has type "Callable[[Any, VarArg(Any), KwArg(Any)], Any]", variable has type "Callable[[EmailMessage, bool], int]")
```

web/main/models.py
dongkyunc: `SearchRank` a few lines down expects type `Union[SearchQuery, Union[str, Combinable, SearchQueryCombinable]]` for the second argument. 

In general, I tend to lean towards avoiding null / None if I can help it, and in this case, would it be appropriate to raise an exception if `query_vector` is `None`? I'm not sure what behavior to expect if you search for nothing.

web/main/models.py
dongkyunc:
```
query_class: ResourceType = ({"legal_doc_fulltext": LegalDocument, "textblock": TextBlock, "link": Link})[
            category
]
```

was starting to look a little unwieldy, but also valid. This was mostly a small refactor for readability.

web/main/models.py
dongkyunc: > `ids_ordinals` is no longer initialized with `None`; it is now defined inside the if `category == "legal_doc_fulltext":` block. Does this code still work without `ids_ordinals = None`, when `category != "legal_doc_fulltext"`?

Within the for loop, since `id_ordinals` is never touched outside the if block, this code should remain the same.

If `ids_ordinals` was referenced, the following error would be raised: `UnboundLocalError: local variable 'ids_ordinals' referenced before assignment` (rather than a `NameError: name 'ids_ordinals' is not defined`

> I see that `results` has been renamed `results_chunk`:

N.B. `results` is initialized as type `QuerySet` due to `results = base_query.filter(category=category)`. Thus, attempting to assign `result` to `Paginator(results, page_size).get_page(page)` a few lines down will result in a type error since `.get_page` return a `Page`, rather than a `QuerySet` object. So the `results` remains, but only as a `QuerySet` object.

>  does this need to be for `r in results_chunk:` instead of deleted?

Er, sorry, not sure what you mean, `for r in results` was [replaced ](https://github.com/harvard-lil/h2o/pull/1639/files#diff-0088ea3fd6dd38dc8969fda2a1559c6de8bab04df5fe137f09208738e9490e97L1105-R1125)with `for r in results_chunk`. 

image

> Added line 1117 adds a type annotation for `ids_ordinals`, but added line 1124 is marked up with `# type: ignore`. Can you help me understand why that's the case?

Since `ContentNode.resource_id` is a field with attribute `null=True`, so even though all `ContentNode`'s of `resource_type` `LegalDocument` all have `resource_id`, `mypy` isn't aware of that, and will throw an error. However, the comment is an opaque fix and there is probably a better way to indicate what's going on here.

web/main/models.py
dongkyunc: > Is this variable renamed with the appendix `_query` because it's assigned to a `queryset`? 

Yeah, mypy really does not like variables changing types.

web/main/models.py
rebeccacremona: Alrighty: I've been poking at this:

This evidently exposes something interesting about type hinting with Django/Django stubs.

A person can initialize a `SearchRank` object with `query` set to `None` or an `int`, `str`, or `float`... and Django will happily convert it to a `SearchQuery` for you. (At least with this version of Django; I haven't checked others.) E.g.

```
(Pdb) SearchRank(F("document"), None)
SearchRank(F(document), SearchQuery(Value(None)))
(Pdb) SearchRank(F("document"), "")
SearchRank(F(document), SearchQuery(Value()))
(Pdb) SearchRank(F("document"), 3)
SearchRank(F(document), SearchQuery(Value(3)))
(Pdb) SearchRank(F("document"), 3.1415)
SearchRank(F(document), SearchQuery(Value(3.1415))
```

That doesn't seem to me to be a bug: I think that's an intentional feature of the class... but it doesn't match the signature as per Django Stubs, [`SearchRank.query`](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/postgres/search.pyi#L78), which is the union of `SearchQuery`, `str`, `Combinable`, and `SearchQueryCombinable`. I think the signature is not perfectly correct.... but whatever.

If it wants the empty string instead of `None` here, and it can have it! 😄 

No need to raise an exception if query_vector is falsey; we want `https://opencasebook.org/casebooks/2566-torts/search/?type=legal_doc_fulltext&q=` to continue to return the full result set, lilke it does now :-)

web/main/models.py
rebeccacremona: Okay, now I see how this works: `ids_ordinals` can be defined exclusively `if category == "legal_doc_fulltext"` because it is only used `if category == "legal_doc_fulltext"`....Something about the way the diff was displaying in Github was throwing me, don't know why. And similarly, sorry about "does this need to be for r in results_chunk: instead of deleted?": that too was me misreading the diff.

rebeccacremona: Re: `ContentNode.resource_id` and the error...
```
main/models.py:1122: error: Key expression in dictionary comprehension has incompatible type "Optional[int]"; expected type "int"
```
... got it. 

If you want the annotation to be [`ids_ordinals: dict[int, list[str]]`](https://github.com/harvard-lil/h2o/pull/1639/files#diff-0088ea3fd6dd38dc8969fda2a1559c6de8bab04df5fe137f09208738e9490e97R1117) instead of `ids_ordinals: dict[Optional[int], list[str]]`, you can explicitly check that `i` is truthy in the dict comprehension:
```
ids_ordinals = {i: [str(n) for n in h] for i, h in ids_ordinals_nodes if i}
```

That seems less desirable to me than using the signature `ids_ordinals: dict[Optional[int], list[str]]`, or not annotating `ids_ordinals` at all.

I think it would also be better to annotate `ids_ordinals` where is defined, rather than above the `if`.

I'm going to make bold and push a commit that switches to `Optional[int]` and puts the annotation inside the `if`; please feel free to lobby for something else.

web/config/wsgi.py
rebeccacremona: Got it: mypy doesn't like monkey patching https://github.com/python/mypy/issues/2427

I'm going to add a comment to that effect.

rebeccacremona (approved):

rebeccacremona:
Cool, thanks for doing this, and sorry to be so slow with approving the changes!!

I'm going to take the liberty of pushing a few commits.

## Stubs

I pulled this code down and ran `docker compose exec web mypy .` and got the following errors:

```
rcremona@HLSNHJGWY2X2J h2o % docker compose exec web mypy .
main/sanitize.py:1: error: Library stubs not installed for "bleach" (or incompatible with Python 3.9)
main/sanitize.py:1: note: Hint: "python3 -m pip install types-bleach"
main/utils.py:15: error: Library stubs not installed for "requests" (or incompatible with Python 3.9)
main/models.py:8: error: Library stubs not installed for "dateutil.parser" (or incompatible with Python 3.9)
main/models.py:8: note: Hint: "python3 -m pip install types-python-dateutil"
main/models.py:8: note: (or run "mypy --install-types" to install all missing stub packages)
main/models.py:8: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
main/models.py:8: error: Library stubs not installed for "dateutil" (or incompatible with Python 3.9)
main/models.py:13: error: Library stubs not installed for "requests" (or incompatible with Python 3.9)
main/models.py:13: note: Hint: "python3 -m pip install types-requests"
reporting/admin/views.py:5: error: Library stubs not installed for "dateutil.relativedelta" (or incompatible with Python 3.9)
reporting/admin/usage_dashboard.py:10: error: Library stubs not installed for "dateutil.relativedelta" (or incompatible with Python 3.9)
Found 7 errors in 5 files (checked 88 source files)
```

I bet you ran `mypy --install-types` in your environment? I'm going to add all those stubs to `requirements.[in, txt]` so that they will be installed in the Docker image. Then, subsequent to `pip-compile` and `pip-sync`:

```
docker compose exec web mypy .
Success: no issues found in 88 source files
```

## Regarding `typing.Sequence`

This turns out to be a known bug with the handling of custom model managers. As per [the docs](https://github.com/typeddjango/django-stubs#my-queryset-methods-are-returning-any-rather-than-my-model), `QuerySet.as_manager()` is not yet supported. But the workaround they suggest, `Manager.from_queryset`, as of the latest release, runs into django-stubs#1022... which is fixed in the merged, but as-yet-unreleased django-stubs#1028.

I installed from the [merged commit](https://github.com/typeddjango/django-stubs/commit/84eff755661c3c498c36e66b7ff0b6cc8fc3d4bb):
```
docker compose exec web bash -c "yes | pip uninstall django-stubs"
docker compose exec web pip install git+https://github.com/typeddjango/django-stubs.git@84eff755661c3c498c36e66b7ff0b6cc8fc3d4bb
```

And behold:
```
docker compose exec web mypy main/models.py
Success: no issues found in 1 source file
```

Version 1.12.0 was released on [June 17, 2022](https://pypi.org/project/django-stubs/#history); hopefully that means we won't have to wait long for the next release!

I added a comment to that effect.

rebeccacremona:
Now let's see if I'm smart enough to resolve the merge conflict using the Github UI 😅

lizadaly:
@rebeccacremona I can do this in a bit if you want since I'm more familiar with the code.

rebeccacremona:
> @rebeccacremona I can do this in a bit if you want since I'm more familiar with the code.

whoops, thanks, didn't see this!

PR #1640, opened 2022-07-18
Page-Navigation Buttons

shljessie:
Starter for creating page navigation buttons on the casebook.

- [x] Getting next node ordinals  : @varun-iyer @dongkyunc
- [x] Getting URLS through ordinals
- [x] Page Button Design
- [x] Go to next page on click


web/main/models.py
```diff
@@ -2060,6 +2060,15 @@ def contents(self):
         res = ContentNode.objects.filter(**filter_map).exclude(id=self.id)
         return res
 
+    @property
+    # gets the next node ordinals
+    def next_node(self) -> tuple(int, ...):
```
dongkyunc: MB, I gave the wrong type. 
```suggestion
    def next_node(self) -> Optional[Tuple(int, ...)]:
```
Also need `from typing import Optional, Tuple` at the top too

shljessie: hmm seems to be still failing.. 🥲 what does the `-> Optional[Tuple(int, ...)]:` part do??

dongkyunc: oh i'm clowning, it should be

`    def next_node(self) -> Optional[Tuple[int, ...]]:`

It's a type hint that lets mypy know what type of object to expect when next_node is returned

web/main/models.py
```diff
@@ -2060,6 +2061,15 @@ def contents(self):
...
 
+    @property
+    # gets the next node ordinals
+    def next_node(self) -> Optional[Tuple(int, ...)]:
+        ordinals = [_[0] for _ in ContentNode.objects.filter(casebook_id=self.casebook_id).order_by("ordinals").values_list("ordinals")]
+        idx = ordinals.index(self.ordinals)
+        if idx + 1 >= len(self.ordinals):
+            return None
+        return ordinals[idx + 1]
+
```
dongkyunc, line 2116: Probably also want a `prev_node` function?

```
    @property
    def prev_node(self) -> Optional[Tuple[int, ...]]:
        ordinals = [_[0] for _ in ContentNode.objects.filter(casebook_id=self.casebook_id).order_by("ordinals").values_list("ordinals")]
        idx = ordinals.index(self.ordinals)

        if idx > 0:
            return ordinals[idx - 1] 
        else:
            return None
```

web/main/models.py
shljessie: gotcha! Thanks!

web/main/models.py
shljessie: added!

dongkyunc:
LGTM!

lizadaly:
Good start! I think this would benefit from some unit tests to demonstrate that it's doing the right thing.

shljessie:
This is the design of the page buttons! 
I still have some stuff to workout like keeping it fixed on the screen, focus effects..etc 
but I wanted to do a design check before I move further 

The buttons automatically disappear when there is no content before or after it!

@lizadaly @rebeccacremona  @cath9 

Main Function


https://user-images.githubusercontent.com/59305253/180040682-2f1d7561-e7c7-49f2-a2bc-041d5d408db4.mov

shljessie:
The page buttons also show up on the sections and I was wondering if we should remove this or keep it? It would move on to the next section or subsection that is right before or after it.
Screen Shot 2022-07-20 at 11 57 52 AM

cath9:
this is looking really good! 

how would the navigation work when a reader comes to the last resource in a section? if they click the next arrow, would they just go right to the first resource in the next section? I ask because one thing we should be mindful of is that we allow authors to put headnotes all over the place. theoretically every section and subsection could have a headnote that readers may miss if we're allowing them to go right to the next resource. What if instead when a reader has reached the last resource in a section, the arrow changes somehow to indicate that they'll now navigate to the next section, they can click and they'll land on the section page, and then they can use the arrow to begin reading that section. Too complicated? 

I'm not sure we should allow them to just click from section to section. I think that might make the arrow functionality too confusing. In my head the arrows are most intuitive if they essentially mimic the experience of flipping pages in a book. 

Another thing we could do is figure out how many people are using headnotes in sections and potentially remove the capability. 

None of these options are mutually exclusive :)

shljessie:
@cath9
Sorry I didn't make this clear 😅

 
1. (END of BOOK) At last subsection in book (aka. last page of book)
> the next button will disappear and only the prev button will be available. 

2. (MIDDLE of BOOK) Last resource in a section 
> the next button leads them to the next section page with the section headnotes and lists of subsections. So it basically treats the sections pages (with all the subsections) like normal pages!

https://user-images.githubusercontent.com/59305253/180045778-fbfceb17-b46f-441f-b001-93d8c3dfb0b3.mov

cath9:
@shljessie awesome! looks like we're on the same page I think this is perfect.

rebeccacremona:
Huzzah!

Worth considering where in the tab order these appear. Do you tab to them after the rest of the page's content, but before the footer? After the section/resource metadata but before the headnote, or after the headnote but before the content? Etc.

I don't have a strong opinion.

rebeccacremona:
I can also suggest some markup, in a bit, for marking these buttons up as a navigation tool for screen reader users.

shljessie:
@rebeccacremona 

Yup! So I am working on fixing them to that location even if the user scrolls down! 

By markup do you mean text under the buttons like 'Next Page' 'Previous Page' ??

rebeccacremona:
@hljessie Oh, I mean the tab order for keyboard users, which is to say, when you tab tab tab through the page, when do you arrive at these buttons: the css will not affect that.

And by "markup" here I mean "HTML", which is to say, stuff we can add behind the scenes, like `<nav aria-label="previous/next">....` or similar: something that won't be visible, but will be of assistance to screen reader users 😄

rebeccacremona:
Note for future us: let's squash these commits when merging. 🙂

shljessie:
@rebeccacremona
Yup right now the it goes after the main block of content(if it is a url)--> next button --> prev button
Added the aria-labels!

shljessie:
Got the page buttons to fix now! thanks for your help :) 
@rebeccacremona

https://user-images.githubusercontent.com/59305253/180467428-d1fa4b6a-f4f7-48f3-b318-08ca1874b2c0.mov

rebeccacremona (changes requested):
This is looking great!!! !

I have a couple tiny requests, and one question.

Thanks for grouping the buttons in a labelled `<nav>` landmark! I think a little punctuation or a conjunction might help in the label, maybe "Previous/Next Page" or "Previous and Next Page", to make it grammatical.

And thanks for labeling the previous and next page links, so that they have an accessible description to supplement the visual styling. Can you remove the word "button" from the labels? Since semantically, links navigate to new content and buttons trigger in-page actions (like opening a drop down menu), these are in fact correctly links (`<a>`) and not buttons `<button>`, just like you have them marked up. With the current labels, screen reader users will hear, "Link: previous page button" and "Link: next page button", which is confusing. "Link: previous page" and "Link: next page" would be better; you don't need to communicate that you added CSS so that these links visually look like buttons.

I love the hover style, where the buttons get bigger when you mouse-over! Can you add that same style change for when the links receive keyboard focus? I think that would be much better than falling back to the browser's default link focus outline, which looks like this in my Firefox
image

and like this in my Chrome
image

Finally, a question: it looks to me like the "previous" link is a lot further away from the table of contents than the "next" button is:
image

Is that an intentional part of the design?

web/main/templates/includes/page_buttons.html
```diff
@@ -0,0 +1,19 @@
+<nav aria-label="Previous Next Page">
+{% load call_method %}
```
rebeccacremona: Are you using `call_method`, or is this left over from copying/pasting?

rebeccacremona:
Sorry, one more tiny request. Can you review the indentation of your changes, and fix things up so that each level of nesting in `.html` files increases by 2 spaces, and then closing tags outdent 2 spaces?

 That is to say:
 ```
<div>
  {% if ... %}
    <span>
      Content.....
    </span>
  {% endif %}
</div>
```

File: web/main/templates/casebook_page.html [hunk 4]
```diff
@@ -23,6 +23,11 @@
 
 <section class="casebook {{casebook_color_class}}">
     <div class="content">
+        <div>
+        {% if section %}
+            {% include 'includes/page_buttons.html'  %}   
+        {% endif %}
+        </div>
```
rebeccacremona: Could this wrapping `<div>` go inside `includes/page_buttons.html` so that there isn't an empty div, when `if section` is false? Or do we need it here all the time?

Sorry to include more questions/etc. outside of the review!

web/main/templates/includes/page_buttons.html
shljessie: leftover! removed!😁

web/main/templates/casebook_page.html
shljessie: No problems thanks for all the comments! added div inside the wrapping!

shljessie:
Changes implemented!  Increase size on hover and fixed naming issues!  @rebeccacremona 

For the spacing issue, I did 80vw and 20vw (https://www.quirksmode.org/css/units-values/viewport.html)  in the (right: 80vw, right: 20vw) but that made the spacing even worse.. 🤷‍♀️ so I customized the spacing for now! 


https://user-images.githubusercontent.com/59305253/180861932-c9932f1b-5389-468b-b3e7-45d7a7c2cac7.mov

rebeccacremona:
This is looking so great! Thanks for those updates! One very last thing, and then we'll merge.

I noticed that the "next"  link and the "previous" link don't both get equally bigger: `transform: scale(1.15);` vs `transform: scale(1.2);`; I assume that's not on purpose? I also noticed that, if you are using a mouse, they scale once when you hover, and then again when you click (because when you click, the focus style is applied). I'm guessing that is by accident too?

I think you can fix both at the same time by making the css "DRY", which is to say: organize things so that the traits that are in common are applied once to both links, rather than separately to each, and so that the traits that are applied on hover are applied on focus at the same time. E.g., selectors like:
```
.previous_button_container, .next_button_container {
  position: fixed;
  bottom: 40vh;

  a {
    display:inline-block;
  }  
  a:focus, a:hover {
      transform: scale(1.15); 
      outline: none;
   }
}
```

Thanks for all your hard work on this! Very nearly there!

rebeccacremona (approved):

rebeccacremona:
Thanks for making those changes!!

I took the liberty of pushing a few more small commits to the PR:

- I noticed that we were hitting the database and building the casebook's ordinal list twice, once for the "previous" link and once for the "next" link... so I combined the calculation into one.

- I added a few tests to prove that the new method `get_previous_and_next_node_urls` works as expected

- I noticed that the buttons don't currently look good on smaller screens, so temporarily hid them except on desktop
image
image

- I fixed what I think was a bug: the buttons were showing up on unpublished casebooks' "edit" tab, as well as their "preview" tab, even though when clicked, they always take you to the "preview" tab.

- I tweaked the css selectors the teeny tiniest bit, so that the cool styling is applied to the `<a>` instead of the `<div>`, so that you can reuse the same transition/animation on both hover and focus.

Since I made changes, I'm going to ask another team member to give this a quick look, but, after I fix the merge conflict, from my perspective this is good to go!

rebeccacremona:
@lizadaly would you mind giving this a glance, since I pushed some commits?

I think when we merge, we should consider squashing+merging

lizadaly (approved):
Thanks to many people for their work on this!

PR #1643, opened 2022-07-19
Allow adding explicit Description to casebooks

### LINKED ISSUE #1580: Allow adding an optional explicit Description to Casebooks
Distinct from the introductory headnote, allow adding a description field to a book.

* The description should be attached at the casebook level.
* The description should be ~150 words or fewer.
* The description should be included in the site search index.
* The description should appear in search results (possibly truncated for length).
* The description should appear on the landing page of the casebook.
* The description should be editable in the Django admin.

varun-magesh:
I don't know how to test the Django admin interface, but I expect that it will work. Everything else listed in #1580 is complete, so this closes #1580.

2022-07-19-132917_1033x571_scrot


lizadaly:
(You can generally assume that the admin "just works" in most cases, unless you're really modifying its behavior.)

varun-magesh:
Tested it to be sure and it works!

forgot to update all the doctests though (and there might be actual failures too), let me convert to draft and come back to this.

varun-magesh:
2022-07-19-150524_1500x107_scrot

matteocargnelutti (changes requested):
Nice, @varun-iyer ! Looks good to me.
The only thing missing is the migration file, since you've made edits to the models 😄 . Mind having a look?

Thank you.

varun-magesh:
> Nice, @varun-iyer ! Looks good to me. The only thing missing is the migration file, since you've made edits to the models smile . Mind having a look?
> 
> Thank you.

Whoops! Thanks for catching that, I did a `git add -u` and forgot to add it explicitly. Should be good to go now!

matteocargnelutti (approved):
Aaaah awesome, thanks @varun-iyer 🚀

PR #1644, opened 2022-07-19
Removing focus box for mouse only

shljessie:
Small thing but it has been bothering me...
Getting rid of the white focus box when we click drop downs in the casebook and only having it for keyboard users.

The white box will appear for keyboard users still!

Before 
Screen Shot 2022-07-19 at 4 56 54 PM

After
Screen Shot 2022-07-19 at 4 57 00 PM

### DIFF
--- a/web/frontend/components/TheTableOfContents.vue
+++ b/web/frontend/components/TheTableOfContents.vue
@@ -439,7 +439,7 @@ export default {
                 color: $white;
                 border-color: $white;
             }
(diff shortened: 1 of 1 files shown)


shljessie:
@sabzo 
Ready to go! The white focus box only appears for keyboard-only (keyboard shortcut users)!

sabzo:
Just seeing this! @shljessie looks good to me!

Issue #1448, opened 2021-11-02
Refactor to remove Section and Resource proxy models


rebeccacremona:
(placeholder issue)

In the meantime, potential gotchas (update, all these links now point to the wrong lines, whoops):
- ContentNode's [from_db](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1660) doesn't cast new ContentNodes to Resource or Section, only content nodes associated with an old_casebook_id, but otherwise, just returns ContentNode
- `Section` querysets are [not limited to sections](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3856); Section.objects.all() is synonymous with ContentNode.objects.all(). That's why, for instance, you can add a resource a book, and then in the URL, change /resources/ to /sections/ and still edit its content.
- the [`clone_to`](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L2479) method will not clone section contents
- when deleting a section, no HistoricalContentNode record is created for the object itself, only its contents
- The asserts for [this claim](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1145) in the docstring have [been removed](https://github.com/harvard-lil/h2o/commit/1f41a655a411c677e51bbbd8359d6c60a7f23b68#diff-0088ea3fd6dd38dc8969fda2a1559c6de8bab04df5fe137f09208738e9490e97L450-L460), so we might need to take another look at whether no. of queries is still linear
- The resource_type field, described as [only being for Resources](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1646), now sometimes [`self.resource_type == 'Section`](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1684)'
- The casebook.sections method will [not return those sections](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3039), or sections with `resource_type == ''`, which also [appears sometimes](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L2738) (though at least in my db `ContentNode.objects.filter(resource_type="").count() == 0`)
- `as_proxy`, which is used in hydrating from URL params, still [endeavors to cast to section](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1683)
- the content node [sub_sections method](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3051) could return resources
[...]

PR #1647, opened 2022-07-20
Speeds up table of contents load for casebook and section pages

varun-magesh:
Speeds up load by:
1. prefetching resources
2. selecting annotations in advance
3. using preloaded children tree


web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+    ...     n.save()
+    >>> for i, (n, d) in enumerate(zip(nodes[3:], docs[3:])):
+    ...     n.ordinals = (1, 2, i+1)
+    ...     n.resource_id = d.id
+    ...     n.edit_url = ""
+    ...     n.casebook_id = casebook.id
+    ...     n.save()
+    >>> nodes[6].ordinals = (1,)
+    >>> nodes[6].resource_id = docs[6].id
+    >>> nodes[6].save()
```
rebeccacremona: I think [this setup](https://github.com/harvard-lil/h2o/pull/1647/files#diff-520b40584c9ab08368ee82905be5a0fae5c23cf54c94c00516356f701ddf006dR759-R777) wants to become its own test fixture defined in [conftest.py](https://github.com/harvard-lil/h2o/blob/develop/web/conftest.py) and then summoned here by name. See [full_casebook_parts_factory](https://github.com/harvard-lil/h2o/blob/develop/web/conftest.py#L462) and [full_casebook](https://github.com/harvard-lil/h2o/blob/develop/web/conftest.py#L532).... or, does the existing `full_casebook` fixture do what you need? If not, I suggest when naming/making your fixture, including some comments that explain what kind of casebook structure you are looking for here.

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+    >>> serialized = manually_serialize_content_query(casebook.contents)
+
+        One top level section, as set up
```
rebeccacremona: I think some stray indentation.

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+    from .models import ContentAnnotation, LegalDocument
+    from .serializers import ContentNodeSerializer
+
+    toc = list(
+        content_query.prefetch_resources(
+            legal_doc_query=LegalDocument.objects.defer("content").all()
+        )
+        .order_by("ordinals")
+        .annotate(
+            has_annotation=Exists(ContentAnnotation.objects.filter(resource_id=OuterRef("pk")))
```
rebeccacremona: Awesome. For extra clarity, that this is something special, maybe consider `has_annotation`->`_has_annotation`? Just a thought.

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+    )
+    # optimize expensive call to is_transmutable
+    for t, t1 in zip(toc, toc[1:]):
+        if not t.resource_type or t.resource_type == "Section" or t.resource_type == "":
+            # if t1 is a child of t, then t is not transmutable
+            t.transmutable = t.ordinals != t1.ordinals[: len(t.ordinals)]
+    serialized = {tuple(c.ordinals): ContentNodeSerializer(c).data for c in toc}
+
+    for cns in serialized.values():
+        if cns["resource_type"] == "Section":
```
rebeccacremona: Does this need to be if `t.resource_type or t.resource_type == "Section" or t.resource_type == ""`, like above? I think yes: I think not all sections have `resource_type='Section'`, especially old ones. https://github.com/harvard-lil/h2o/issues/1448

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+        .all()
+    )
+    # optimize expensive call to is_transmutable
+    for t, t1 in zip(toc, toc[1:]):
+        if not t.resource_type or t.resource_type == "Section" or t.resource_type == "":
+            # if t1 is a child of t, then t is not transmutable
+            t.transmutable = t.ordinals != t1.ordinals[: len(t.ordinals)]
+    serialized = {tuple(c.ordinals): ContentNodeSerializer(c).data for c in toc}
+
+    for cns in serialized.values():
```
rebeccacremona: The nitpickiest of nitpicks: here and below, I keep getting confused by the variable name `cns` here, thinking it means `content nodes`, or `content node serializer`, when in fact it is a single dict/already-serialized node. Maybe `node` or `node_dict` or `serialized_node`, or something else would be clearer?

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+        try:
+            parent = serialized[ordinals[:-1]]
+        except KeyError:
+            # no parent, append to root
+            parent = serialized[()]
+        try:
+            parent["children"].append(cns)
+        except KeyError:
+            raise ValueError("Trying to append children to non-Section!")
+    return serialized[()]["children"]
```
rebeccacremona: I'll admit I'm struggling a little with how this bit works. Would using a fresh dict as the root, like the below, also work, or is it important to do the work on a falsey key in `serialized`?

```
tmp_root = {"children": []}
for ordinals, serialized_node in serialized.items():
    try:
        parent = serialized[ordinals[:-1]]
    except KeyError:
        # this node is at the top-level of the data being serialized
        parent = tmp_root
    try:
        parent["children"].append(serialized_node)
    except KeyError:
        raise ValueError("Trying to append children to non-Section!")
    return tmp_root["children"]
```
Is that the same?

web/main/serializers.py
```diff
@@ -57,6 +57,42 @@ def to_representation(self, value):
         return serializer.data
 
 
+class ContentNodeSerializer(serializers.ModelSerializer):
```
rebeccacremona, line 68: True? This is identical to `SectionOutlineSerializer` except for `children = RecursiveField(many=True, allow_null=True, default=[])` -> `children = []` (and therefore no `children` in the `Meta`, and the presence of `ordinals` as well as `ordinal_string`?

If True, can `SectionOutlineSerializer` be removed entirely?

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+    >>> nodes[6].ordinals = (1,)
+    >>> nodes[6].resource_id = docs[6].id
+    >>> nodes[6].save()
+    >>> serialized = manually_serialize_content_query(casebook.contents)
+
+        One top level section, as set up
+    >>> assert len(serialized) == 1
+
+    Serialized data has all expected keys
+        >>> assert all([
```
rebeccacremona: I think some stray indentation.

web/main/utils.py
```diff
@@ -741,3 +742,109 @@ def validate_image(file, formats=None):
...
+            has_annotation=Exists(ContentAnnotation.objects.filter(resource_id=OuterRef("pk")))
+        )
+        .select_related("casebook")
+        .all()
+    )
+    # optimize expensive call to is_transmutable
+    for t, t1 in zip(toc, toc[1:]):
+        if not t.resource_type or t.resource_type == "Section" or t.resource_type == "":
+            # if t1 is a child of t, then t is not transmutable
+            t.transmutable = t.ordinals != t1.ordinals[: len(t.ordinals)]
```
rebeccacremona: Same here: for extra clarity, maybe this special property should be `_transmutable`?

web/main/views.py
```diff
@@ -568,8 +566,10 @@ class SectionTOCView(APIView):
     @method_decorator(user_has_perm("casebook", "viewable_by"))
     @method_decorator(user_has_perm("section", "viewable_by"))
     def get(self, request, casebook, section, format=None):
-        section.content_tree__load()
-        return Response(SectionOutlineSerializer(section).data)
+        mscq = manually_serialize_content_query(
+            ContentNode.objects.filter(id=section.id) | section.contents
```
rebeccacremona, line 574: I don't quite understand this line; I think I'm stumbling over the `|`. Can you add a comment?

web/main/views.py
```diff
@@ -568,8 +566,10 @@ class SectionTOCView(APIView):
...
+        )[0]
```
rebeccacremona: I have a syntax recommendation! To make it clear that you expect `manually_serialize_content_query(...)` to return a list of length one and extract that item at the same time, you can:
```
[mscq] = manually_serialize_content_query(....)
```
I think that would be a bit clearer than `[0]`, which could make a person wonder why we aren't returning the [non-existent) rest of the list too.

web/main/views.py
```diff
@@ -2973,7 +2973,10 @@ def add_sections_and_resources(parent_section, nodes):
     add_sections_and_resources(parent_section, nodes)
     parent_section.content_tree__repair()
     if section:
-        return JsonResponse(SectionOutlineSerializer(section).data, status=200)
+        mscq = manually_serialize_content_query(
+            ContentNode.objects.filter(id=section.id) | section.contents
+        )[0]
```
rebeccacremona: (same as above)

rebeccacremona (changes requested):
This is SO FANTASTIC 🎉 Thanks so much for noticing and tackling this!

I have a bunch of little notes/questions made inline.

When we were chatting about this code and you mentioned not being sure where to put it, I mentioned `utils.py` as a catch-all: I think that actually, it would be extremely natural for this code to live in `serializers.py`... and I think that means you wouldn't have to do anything fancy with the imports.

I think you can just put the method in there, but now I'm musing whether it might be elegant to write it as class that looks more like a DRF serializer, something like:
```
class CustomFooSerializer:

  __init__(content_query: QuerySet):
    # logic of `manually_serialize_content_query` except...
    self.data = {...........}
```
It's probably too crazy to use https://www.django-rest-framework.org/api-guide/serializers/#baseserializer..... 

Maybe either class-based idea is goofy....

Again, thanks for this optimization!

rebeccacremona:
Oh! I had one more thought: I think this deserves an `assert_num_queries` in the tests, to brag about how efficient it is.

web/main/utils.py
varun-magesh: you mentioned this and I totally forgot --- will do!

web/main/utils.py
varun-magesh: good to know!

web/main/utils.py
varun-magesh: Yes, this is definitely the same --- the story here is that I originally wrote this to work for casebooks only, where all top-level sections have `len(ordinals)==1`. I didn't need a try catch there because the slice alone would take the ordinals to an empty tuple, and this solution felt rather natural. I came back later to modify it for sections, which required the try catch, and required me to give up the nice brevity of the original solution while keeping its extra confusing-ness.

varun-magesh:
> It's probably too crazy to use https://www.django-rest-framework.org/api-guide/serializers/#baseserializer.....

I considered writing it as a DRF Serializer, but the problem I ran into was that the thing we are serializing  is not a Django model or object --- it's a QuerySet of a series of objects that we know how to put together. Even the previous implementation did some steps outside the serializer (converting a section's tree into a dict) and used the built-in serializer to serializer individual nodes in the tree. Basically (and I might be wrong because I haven't worked with serializers or django much), I think that what we are trying to do is different from what Serializers are made for (mainly classes and Django models).

So I think that using the serializer interface (either by copying it with `.data` or by inheriting explicitly) is misleading to someone who is familiar with serializers since this code does something pretty different than what serializers are expected to do --- one major example, it cannot be used to deserialize trees.

It seems like the best option might be to include this as a seperate method in serializers.py.

That being said, I'm not particularly familiar with Serializers or DRF; I wanted to explain my thought process here but if it makes more sense to use serializer-type objects here I'm happy to implement it that way.

rebeccacremona:
> So I think that using the serializer interface (either by copying it with `.data` or by inheriting explicitly) is misleading to someone who is familiar with serializers since this code does something pretty different than what serializers are expected to do --- one major example, it cannot be used to deserialize trees.

👍 Fair enough

web/main/serializers.py
varun-magesh: True and True!

web/main/utils.py
varun-magesh: full_casebook_factory is exactly what I needed, thankyou!

varun-magesh:
Thanks so much for your feedback! Everything should be addressed now!

varun-magesh:
Hmm --- before we merge this:

When deleting a section or a resource in casebook edit mode, it seems like this PR introduces a 1-2 second lag between hitting the "Delete" button and the section actually disappearing from view, which didn't use to happen. I'm not sure how this might be, I'd like to investigate a bit.

rebeccacremona (approved):
This looks great! Thanks so much for all the work!

We won't merge until told to, pending [your investigation about delete time/queries](https://github.com/harvard-lil/h2o/pull/1647#issuecomment-1194371492)

varun-magesh:
Update: this bug is, unfortunately, very real. The requests actually look a little different before and after speedup. My code, however, is a drop-in replacement for the serialization that was happening before, and didn't touch requests, and in theory, the serialized data returned by views should be identical. Banged my head on this for too long, gonna take some air and go at this with a fresh pair of eyes tomorrow.

varun-magesh:
Yay this is finally ready! The bug has been fixed. Thanks @rebeccacremona !

Issue #1581, opened 2022-07-05
Allow authors to add additional personal information to their public page


lizadaly:
From the Edit Profile page, allow authors to add additional fields about themselves.

We may want to make this a key/value relationship that's open-ended, but pre-populate a set of fields to prompt them.

Examples:

* Professional title
* Pronouns
* Institution (is this distinct from the existing "Affiliation" field?)

cf. https://github.com/harvard-lil/h2o/issues/1390

PR #1652, opened 2022-07-21
Add pronouns to User model and expose to /dashboard

dongkyunc:
# What

Add pronouns to User model and expose to /dashboard

# Why

Pro-pronoun gang

# How

- Add a migration to User
- Removed a comment from Rails

# Notes to reviewers

- Addresses #1581 
- Taking suggestions on how to style this, I have TUIs-are-perfectably-acceptable-UX linuxbrain. Current conception

image


shljessie:
This looks really good! 
I think this could be a good starter in creating the information necessary for allowing more collaboration btw authors.@cath9  In addition to this, having a profile photo of the author, stuff like main topics taught, and info on the number of clones the author received on their books would be nice! 
Additionally, maybe a contact this author button, that leads to an email page? 

@rebeccacremona is there a way of knowing how many clones a book received?

rebeccacremona:
> is there a way of knowing how many clones a book received?

Yes, using https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1918, though keep in mind you don't have to clone a whole book, you can clone any little piece of it separately.

If wanting a per-author stat something like this, it probably needs to be cached on the model to avoid an expensive series of database queries.

lizadaly:
This looks good! Just one admin thing—there's been another migration since, so you'll need to regenerate that. The most recent one is `0034_add_instructional_content_field`.

dongkyunc:
`makemigrations` also attempted to add the following:
```
    - Alter field display_ordinals on contentnode
    - Alter field does_display_ordinals on contentnode
    - Alter field is_instructional_material on contentnode
    - Alter field ordinals on contentnode
    - Alter field display_ordinals on historicalcontentnode
    - Alter field does_display_ordinals on historicalcontentnode
    - Alter field is_instructional_material on historicalcontentnode
    - Alter field ordinals on historicalcontentnode
```

I've removed them, but not if they should be migrated.

lizadaly:
Those changes are what was in the 0034_add_instructional_content_field migration, so I think maybe that just hadn't been applied yet when you saw that message?

cath9:
Looks good! Is this info that can be populated from their profile? If somebody leaves a field blank, does that entire line get removed from this public facing page?

dongkyunc:
> Is this info that can be populated from their profile? 

Yup, these are all fields in `/user_edit`

> If somebody leaves a field blank, does that entire line get removed from this public facing page?

Yes

lizadaly (approved):

PR #1653, opened 2022-07-21
Starting off the models.py refactor

dongkyunc:
# What

Organizing models in a package

# Why

I'm being annoying about this once again after confusing `ContentNode` and `Casebook `methods with each other for the nth time while reading code. 

More seriously, this coincides well with Monday's dev discussion about slowly getting around to refactoring the models. This will preserve the git history of the file while letting us gradually refactor by moving code out of `models.py` and into subpackages inside `models/`.

# [How](https://docs.djangoproject.com/en/4.0/topics/db/models/#organizing-models-in-a-package)


rebeccacremona:
Thanks for weighing in on readability/ergonomics! 

Let's chat in person as a group at the dev meeting on Monday about whether we think this is worth the tradeoffs. Converting to a draft in the meantime.

rebeccacremona:
Discussed in today's team meeting! We decided collectively that, as we are refactoring models going forward, we will endeavor to start migrating logic out of one big file and into separate modules: either one-per-file, or into logical groupings, as seems fitting. We'll aim to do so keeping an eye out for circular imports and inter-dependencies; we will aim for maximum readability and discoverability; we'll keep in mind that editors like PyCharm and VS Code can do a lot for us and will approach this from an architectural/engineering perspective.

PR #1654, opened 2022-07-21
Move github workflows into a single file

dongkyunc:
# What

Some CI cleanup.

# Why

Addresses this [comment](https://github.com/harvard-lil/h2o/pull/1639#issuecomment-1187847187). (Avoiding overhead in spinning up multiple containers).

# How

- Moved `black.yml` and `lints.yml` to steps of `tests.yaml`

# Note to Reviewers

- Not quite sure how to test workflows w/o committing it to master.
- I'll add `mypy` once #1639 gets merged.


lizadaly:
Oh sorry, I should've been more clear—I specifically meant just combining the "linting" steps (mypy, black, flake8) in a single workflow, but keeping that distinct from the test suite. 

I'd be interested to hear others' opinions too but I find it useful to find out early if some code has broken an assumption, but I don't typically care about the style checks until the functionality is logically correct.

jcushman:
> I'd be interested to hear others' opinions too but I find it useful to find out early if some code has broken an assumption, but I don't typically care about the style checks until the functionality is logically correct.

This might be what you're saying, but I think the ideal is to provide failure information in parallel as fast as possible, for any check that's allowed to block a PR. One common case is to make a change that is obviously logically correct, and then have to revise once or twice because of style lints. (The CI lints fail for some reason that doesn't match dev, or you forgot them on dev or whatever.) In that case you want the style check to fail quickly.

So like the hypothetically best thing is for *everything* to run in parallel. But if some things take 1x seconds and some take 100x, it's close to as good to just run the 1x tests first?

lizadaly:
Yeah that's right. In practice Github won't allocate infinite capacity for action runs, so they can end up being serial after a point—the actual runs can be fast, but spinning up the containers is not, plus you factor in wait time. 

But I'm also a little sensitive to reducing the number of runs overall because they're a paid service in private repos and I didn't realize until recently that doesn't apply here.

rebeccacremona:
I too am a fan of having linting and tests run separately; no opinion on whether grouping them together or having them all be separate actions is nicer.....

Purely from a github UI perspective, I think I'd rather be looking for two green check marks instead of lots of green check marks, but that is SO individual and trivial.....

lizadaly (changes requested):
Thanks everyone! Looks like the consensus here is to move the `black` and `mypy` workflows into `lints.yml`.

.github/workflows/tests.yml
```diff
@@ -11,6 +11,17 @@ jobs:
...
+        run: |
+          pip install `egrep -o 'flake8==\S+' web/requirements.txt`  # install our version of flake8
+          flake8 web/ --config web/setup.cfg
+          flake8 docker/pandoc-lambda/function/ --config web/setup.cfg
+
+      - name: black
+        run: |
+          pip install `egrep -o 'black==\S+' web/requirements.txt`  # install our version of black
+          black --check --diff . # Uses pyproject.toml
+
```
rebeccacremona: True? With this setup, lints will run both in the `lints` workflow and again in the `tests` workflow? Does [.github/workflows/tests.yml](https://github.com/harvard-lil/h2o/pull/1654/commits/b1a66eeddfb5a21dc9f9c913fefd9bd9e2fd083f#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957f) want to be reverted?

dongkyunc: Ah oops, thanks for the catch

lizadaly:
I think this looks right now!

lizadaly (approved):

PR #1660, opened 2022-07-21
Add missing test for invitation form

lizadaly:
This test was once part of the invitation flow; you can even see it in the [code coverage report for the original PR](https://github.com/harvard-lil/h2o/pull/1592#issuecomment-1175536455). It didn't make it to merging, probably because that got held up in the long chain from black -> flake8 -> merge. Somewhere along the way that typo got introduced too. :(

This puts the test back (I confirmed the test would've caught the typo. Tests work, assuming you add them.)

### DIFF
--- a/web/main/test/test_forms.py
+++ b/web/main/test/test_forms.py
@@ -1,0 +1,28 @@
+from unittest.mock import Mock
+from main.forms import InviteCollaboratorForm
+from main.models import User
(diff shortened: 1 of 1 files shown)


matteocargnelutti (approved):
LGTM! Glad this test found its way back 🤖 !

PR #1662, opened 2022-07-22
Add integration in usage dashboard with Matomo API; fixes #1584

### LINKED ISSUE #1584: Import web metrics from Matomo into admin usage dashboard
In order to answer questions like, "How many views did this casebook get?" web analytics data needs to be available. While this can be done manually, it would better inform decision-making to have it side-by-side with usage data from the application itself. Web metrics could also be used to populate a professor-facing dashboard e.g. #1250 

Matomo has an [API](https://developer.matomo.org/api-reference/reporting-api). Investigate whether it is possible to pull down sufficient data to answer usage questions. The data can be stored in local reporting tables following the pattern used in #1574 

Initial questions to answer:

Top 10ish most visited casebooks
Top 10ish most frequently used cases in H2O
Top 10ish most frequently used sections of code in H2O
Top 10ish authors in terms of # of published casebooks

Should respect the date filters in the page.

lizadaly:
Incorporate usage data from Matomo (like Google Analytics but open source and self-hosted) into the usage dashboard.

It looks like this:

image

The usage data will respect the filters provided at the top (date range and whether the casebook is published), except that the possible date range is capped at 6 months max or else Matomo will crap out.

**Note for deployment**

This needs to be paired with a salt change as two new Django settings/secrets have been added: `MATOMO_API_KEY` (a real secret; we should generate a key specifically for prod) and `MATOMO_SITE_URL` (not really a secret but specific to our environment, which should be parametrized.) Could do the same for `MATOMO_SITE_ID`, which corresponds to H2O in our Matomo install.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -37,3 +37,5 @@ services:
     environment:
       - CAPAPI_API_KEY
       - GPO_API_KEY
+      - MATOMO_API_KEY
+      - MATOMO_SITE_URL
--- a/web/config/settings/settings_base.py
+++ b/web/config/settings/settings_base.py
@@ -127,6 +127,9 @@ USE_L10N = True
 
 # LIL's analytics JS
 USE_ANALYTICS = False
+MATOMO_SITE_URL = ""
+MATOMO_API_KEY = ""
+MATOMO_SITE_ID = "3"
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/2.2/howto/static-files/
@@ -153,6 +156,8 @@ PROFESSOR_VERIFIER_EMAILS = ["info@opencasebook.org"]
 # e.g. <a href="mailto:{{ CONTACT_EMAIL }}">Contact Us</a>
 TEMPLATE_VISIBLE_SETTINGS = (
     "USE_ANALYTICS",
+    "MATOMO_SITE_ID",
+    "MATOMO_SITE_URL",
     "CONTACT_EMAIL",
     "GUIDE_URL",
     "BLOG_URL",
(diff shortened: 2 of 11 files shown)


lizadaly:
Moving to draft to sort out the settings issue.

bensteinberg (approved):
This looks great. I just set up a user, API key, and the settings interpolation for deployment. I think the web_usage view will be broken on stage, as it throws a `NotImplementedError` in the absence of the settings. Should its presence be conditioned on `USE_ANALYTICS`?

lizadaly:
> This looks great. I just set up a user, API key, and the settings interpolation for deployment. I think the web_usage view will be broken on stage, as it throws a `NotImplementedError` in the absence of the settings. Should its presence be conditioned on `USE_ANALYTICS`?

I think it'd be nice if it worked on stage, actually. It's set up to be tolerant of a mismatch between the casebooks available on its local database and the prod reporting, because this is what happens in local dev. That's what this test asserts:

https://github.com/harvard-lil/h2o/pull/1662/files#diff-52891a1f4a70b70f9e50154b1ddbc863c14c8cec88b6e04fb9289dc75f68fc26R36-R42

Would it be OK to include the API host/key settings in stage but keep `USE_ANALYTICS` as false, so it can read reports but won't issue metrics?

bensteinberg:
Oh neat! Sure, I've set it up that way now.

lizadaly:
(It's also intended to be resilient to the case where we've hard-deleted a casebook for some reason, perhaps because the content was inappropriate, but the reporting data is still extant.)

lizadaly:
Thanks so much for the infrastructure changes! Let's see if this works on staging...

PR #1664, opened 2022-07-25
Book Covers --> Fixing Layout of Casebook Dashboard

shljessie:
Display casebook covers on H2O! 😊

Now!
Screen Shot 2022-08-01 at 12 21 18 PM

https://user-images.githubusercontent.com/59305253/182197655-3e3ccf0d-3e54-4093-a41e-dbb51b3a40b7.mov

Before!
Screen Shot 2022-08-01 at 12 25 05 PM


shljessie:
Uploaded book covers can be displayed in H2O. Before I move further, I think we have some design decisions to discuss!
@cath9  @rebeccacremona  

Currently it looks like this .. 
Screen Shot 2022-07-25 at 12 25 25 PM

Should we have casebooks with covers - with just covers? 
Screen Shot 2022-07-25 at 12 30 38 PM

Or should we have their descriptions below? 
Screen Shot 2022-07-25 at 12 33 56 PM

rebeccacremona:
> Before I move further, I think we have some design decisions to discuss!

Let's make this a draft until you think it is ready to merge.

shljessie:
New Designs 
image
image

shljessie:
Talked with Catherine and it seems that we will be going with the first design above! Open to other suggestions too!

shljessie:
Frontend final touches are not finished but here is the progress so far! @cath9 

https://user-images.githubusercontent.com/59305253/181360390-5c7ec4ed-f53a-4571-990c-c6fa0f3e851d.mov

shljessie:
https://user-images.githubusercontent.com/59305253/182195845-750951aa-f0d3-48f7-9d32-62f23c80af40.mov

shljessie:
- Keyboard Focus added 
- Manage casebooks select (checkboxes for selecting books) styling changed
- Hover effects 
- CSS edits to Create a Casebook

https://user-images.githubusercontent.com/59305253/182194976-7ab84201-6d11-482c-9503-26671809557f.mov


Keyboard Focus ! For Books, For Selecting Books

https://user-images.githubusercontent.com/59305253/182195037-a19bec06-089d-4fe9-9c4c-3e7baea43aa2.mov

https://user-images.githubusercontent.com/59305253/182195057-eeaec4ab-40e7-45a3-8629-651a2b46503e.mov



Select Styling
Screen Shot 2022-08-01 at 12 14 20 PM

lizadaly:
This functionality looks and works great!

I think there's a problem with the actual book cover uploading functionality though, that might make us want to hold this (or ship it but keep the `BOOK_COVER` flag off, if that's a compatible option).

I'm having a hard time cleanly reproducing but I think it might have to do with how revision works. I think the simplest reproduction case is:

* Add a cover to a draft casebook
* Change or clear the book cover
* Observe that the book cover is changed or removed, as expected
* Publish the book
* Revert it back to draft
* Change or clear the book cover 
* Publish again
* Observe on the dashboard page that the old book cover is still there

Maybe this has something to do with how revisions are temporary clones of the book?

rebeccacremona:
> Maybe this has something to do with how revisions are temporary clones of the book?

Oh, I didn't think to look at drafts/revisions at all when adding this field... Thanks for thinking of that!!

So, looks like cover image needs to be added to the merging process here: https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3894. And, we might want to reconsider the paths at which I'm saving images... right now, it's by casebook ID, but that will get weird w/merged drafts, I think.

I also didn't think to check how cloning works... probably need to add logic to duplicate the image on clone....

rebeccacremona:
That's enough work that maybe this PR wants to be considered separately, and then me-or-somebody follows up with the additional back end logic before we toggle the feature flag?

shljessie:
That sounds good! should I convert this to a draft till it is done or just make the `COVER_IMAGES = False` in settings_base.py??

rebeccacremona:
@shljessie How do things look/work with this PR's code and `COVER_IMAGES = False`? If all is well, yes: let's do that. If not, let's tweak this so that `COVER_IMAGES` can be False while this is deployed and we keep working on it: I'm not sure if one of us will get to working the additional back end changes this week.

I'll pull it down and check too, but you are most qualified to judge :-)

rebeccacremona:
Quick look: I think this is fine w/ `COVER_IMAGES = False`: people won't be able to upload their own cover images; if you did already locally, or if you add one in prod via the Django admin, the dashboard will still display them... That's probably fine, we just shouldn't add any images in prod.

The more robust way would be to pass the `COVER_IMAGES` setting to the frontend JS like we do [with these other settings](https://github.com/harvard-lil/h2o/blob/develop/web/main/templates/base.html#L10), and then check the setting before [rendering the image](https://github.com/harvard-lil/h2o/pull/1664/files#diff-c2c0c751886605797cf7e74830014d12dd991261e5eb25ec1e97397392eab3c0R23). Probably overkill?

rebeccacremona:
Probably worth fixing: each dashboard item is an `<a>`
image

`<button>` is not allowed inside of `<a>`. (Also since buttons are inherently focusable elements, no need for `tabindex=-1` on a button element, which makes an element programmatically focusable without inserting it into the tab order.)

Sounds like this "button" is a purely visual thing, a fake button to indicate visually that if you click anywhere in the big link, you'll get to see the book.

I think therefore it would more appropriately be a `<span>` element, styled with `.btn` classes.

shljessie:
I've commented out the sub -info section of the casebooks and just leaving this PR with the cleaned layout of casebooks, focus effects and fixes to the add a casebook button like this!

** The button is the view button in the bottom sub-info section so I think removing it should solve the `<button>` inside an `<a>` problem!

Let me know if you think commenting out is not a good idea!

Screen Shot 2022-08-04 at 12 02 03 PM

lizadaly:
I think that’s a great idea!

lizadaly (approved):
This looks very clean! Thank you for this work and I'm looking forward to getting the covers in too!

lizadaly:
Ran this locally and the hover effects are so nice.

PR #1665, opened 2022-07-25
Add CSV export option to reporting admin; fixes #1627

### LINKED ISSUE #1627: Add export-list-as-csv to reporting dashboard
For any reporting page in the reporting dashboard, add an export step to return a CSV version of the list rather than a Django list page.

It might be simpler to just hand-define an export CSV template for casebooks and professors, since the fields we might care about for outreach may be different from what we want to see in the UI.

lizadaly:
Adds the ability to export any of the reporting views as a CSV in addition to a Django admin changelist view.

image

This hooks in to the changelist view and checks for a `csv=True` query param; if so it returns a CSV view that auto-downloads; otherwise it returns an admin list view.

The list of fields in the CSV is not the same as the corresponding admin view; they have to be coerced into strings anyway, and I assume what is useful in a CSV export (like for sending a mailing) is different from other use cases, like editing a record. Those are defined in a mix-in class as `field_list` and in some cases reference new properties on the reporting models. This lets us customize the reporting output in a reusable way without cluttering up the main models.  I made a guess at fields that might be useful in a CSV report; Catherine can give me additional feedback later about changes or additions.

Since a CSV view can't be paginated there's a danger this view could take too long to render or exhaust all the RAM available to a request. That's especially plausible if the user unchecks "published casebooks only", where the number of casebooks can exceed 5,000. I set an arbitrary cap of 1,000 records on CSV results; we can adjust that downward if that's too generous on prod.

### DIFF
--- a/web/main/templates/admin/reporting/index.html
+++ b/web/main/templates/admin/reporting/index.html
@@ -111,6 +111,8 @@
         <th>Verified professors
             <small>
                 <a href="{% url "admin:reporting_professor_changelist" %}?{{ date_query }}">View as list »</a>
+                <a href="{% url "admin:reporting_professor_changelist" %}?{{ date_query }}&_csv=True">Export as csv »</a>
+
             </small>
 
         </th>
@@ -120,6 +122,8 @@
         <th>Professors with published casebooks
             <small>
                 <a href="{% url "admin:reporting_professorwithcasebooks_changelist" %}?{{ query }}">View as list »</a>
+                <a href="{% url "admin:reporting_professorwithcasebooks_changelist" %}?{{ query }}&_csv=True">Export as csv »</a>
+
             </small>
 
 
@@ -129,8 +133,11 @@
     <tr>
         <th>Casebooks
             <small>That have been modified within this period</small>
+
             <small>
                 <a href="{% url "admin:reporting_reportingcasebook_changelist" %}?{{ query }}">View as list »</a>
+                <a href="{% url "admin:reporting_reportingcasebook_changelist" %}?{{ query }}&_csv=True">Export as csv »</a>
+
             </small>
 
         </th>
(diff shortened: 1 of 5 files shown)


lizadaly:
Moving back to draft to resolve a changed behavior after merging from upstream.

web/main/templates/admin/reporting/index.html
```diff
@@ -111,6 +111,8 @@ <h1>Usage dashboard</h1>
         <th>Verified professors
             <small>
                 <a href="{% url "admin:reporting_professor_changelist" %}?{{ date_query }}">View as list »</a>
+                <a href="{% url "admin:reporting_professor_changelist" %}?{{ date_query }}&amp;_csv=True">Export as csv »</a>
```
rebeccacremona: I don't know if this matters: I'm pretty sure you can use `&` directly here, rather than `&amp;`

rebeccacremona: Just to satisfy curiosity: why `_csv` rather than `csv`? Is it to make it especially clear this is custom logic, not part of the Django admin's own functionality?

rebeccacremona (approved):
LGTM!

web/main/templates/admin/reporting/index.html
lizadaly: My editor was grumpy about `&` but I agree it's fine so I'll change it to the more readable thing!

When I was trying to figure out how to do this weird delegation thing I saw that Django internally used a pattern like this to determine whether to pop up confirmation dialogues or otherwise dispatch differently during the change list request cycle:

https://github.com/django/django/blob/main/django/contrib/admin/options.py#L1370-L1372

so I just went with that as yeah "this is a special param" rather than a query param that filters the result list.

lizadaly:
I updated this to account for a change to the logic in #1666 where the professor usage report requires the user to have logged in to generate a record. It turns out the Django test client doesn't fire off the related signal, which our code depends on to add the timestamp, so the test just adds the timestamp manually. 

This is the only material change since it was originally reviewed:

https://github.com/harvard-lil/h2o/pull/1665/files#diff-e6c990a97284443b3a5fbb96078b0668e7cff9f783557f357ee475e0ca6fc882R200-R202

PR #1666, opened 2022-07-25
Change date filters in reporting to apply more to recent usage than creation time

### LINKED ISSUE #1628: Have reporting date filters on users and professors apply to last-logged-in times, not creation times
It's more useful from a reporting and usage perspective to look at when users were last active on the site than when they created their accounts. Switch the reporting filters that apply to humans to look at `last_login_at` rather than `created_at`.

lizadaly:
It's more useful in outreach activity to sort and filter by active use of the product, than when an account or casebook was first created.

This switches the date filters in usage reporting to reflect usage rather than blunt instance creation time. The specific date varies by the type of object being reported on:

* For users (e.g. professors), the date filters apply to the time they last logged in to the product, regardless of when their account was created.
* For casebooks, the date filters apply to either the last-modified date of the `Casebook` instance, or the most recent modified date based on the edit log, whichever is greatest.

Introduces the [freezegun](https://github.com/spulec/freezegun) dependency for testing, which I find extremely useful for any date-based testing.

Fixes #1628

This generally makes the reported numbers go up. For example, with casebooks:

image

All of the ones with created-by dates prior to 2022 would not have shown up in the prior report, but we want to know about them!

### DIFF
--- a/web/main/templates/admin/reporting/index.html
+++ b/web/main/templates/admin/reporting/index.html
@@ -34,7 +34,7 @@
         }
         #toolbar th, #toolbar td { vertical-align: middle; }
         #toolbar {  margin-bottom: 1em;}
-        #dashboard th small { font-weight: normal; display: block; }
+        #dashboard th small { font-weight: normal; display: block; padding-bottom: .5rem; }
         aside { background-color: #eee; padding: 1em; margin-top: 2em;}
         #dashboard {
             display: grid;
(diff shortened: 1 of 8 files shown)


bensteinberg (approved):
LGTM! I think this does not need a migration?

lizadaly:
Correct, it does not need a migration. Thanks for the review!

Issue #1669, opened 2022-07-26
Remove remaining references to FORCE_DOCX_* settings and related code?


lizadaly:
It looks like in https://github.com/harvard-lil/h2o/pull/1485 a global setting was enforced that turns on a previously-experimental pathway to provide a richer export in docx and HTML format.

There are still a number of places in the code where this value is checked even though I think it's now enabled in all cases:

* For admins, an extra purple export button appears that uses a query param to force the setting. I believe this is redundant and does the same thing as the normal export button. The only difference in the request it generates is that it appends `docx_sections=true`, but I think that is true in all cases. New team members are confused by the extra button (it's not obvious that it's admin-only either). Can this be removed?
* Several places in the code check the request for the query param (see https://github.com/harvard-lil/h2o/search?q=docx_sections). If I'm reading it right you could still force it to be off by setting the query param to something other than `true`, but I don't think it's being set anywhere like that on the front end.
* It seems that this setting affects HTML output and chooses a different HTML template if _disabled_: https://github.com/harvard-lil/h2o/blob/b8a2641027d1b98b1330cf2daf7027c12673f83b/web/main/models.py#L4363-L4365. That template looks like something we probably don't want to be using anymore!

Is there a reason to keep the old pathway around?

rebeccacremona:
@lizadaly Definitely no! We can for sure remove all of this!

PR #1670, opened 2022-07-27
Remove experimental export button; related to #1669

lizadaly:
Part of cleanup in #1669 : remove the admin-only purple export button which does the same thing as the regular export button now.

Was: 

image

Now admins see the same thing as everyone else:

image

Non-admin users should see nothing new.


matteocargnelutti (approved):
LGTM 👍

PR #1671, opened 2022-07-28
Correct limit filter on casebook date filtering

lizadaly:
Goofy mistake in generating the SQL for casebook reporting with new date logic.

There's already a test for the date logic itself and the previous wrong query happened to satisfy it. I don't think a specific test to catch this mistake is necessary since it's extremely obvious (the whole casebook reporting view had only one item in it).

When this is deployed it would be ideal to run `fab create_reporting_views` again so the fix is applied. (That would safe to run on every deploy and then we don't need to worry about it—it's fast.)


bensteinberg:
Shall I add `fab create_reporting_views` to the deployment orchestration?

sabzo:
Looks good to me @lizadaly

PR #1672, opened 2022-07-29
Improve casebook date filtering

lizadaly:
This changes the behavior of the casebook reporting views to use the dates of major activity associated with the History tab rather than favoring the most-recent modification date. Partly this is to better reflect what "recent activity" really means—we're concerned with edits and publication status changes more than minor metadata changes. It also make the reporting more resilient to immaterial changes like bulk modification date updates, as has occurred in the past.

The practical outcome of this fewer casebooks are reported as having been recently changed, but the reports should better reflect the kinds of changes we care about.

This PR includes some other small fixes to reporting: 

 * Handling the case where Matomo returns a query validation message but not does strictly return an error response (because you sent it nonsensical date ranges)
 * Modifies the date-modified in the Casebook CSV result to be specific to the history table, so the CSV export better reflects the input query.
 * Switches the assertion in the casebook date test to reflect the new desired behavior favoring the edit log.

### DIFF
--- a/web/main/templates/admin/reporting/index.html
+++ b/web/main/templates/admin/reporting/index.html
@@ -254,9 +254,16 @@
                 <em>revising</em> state, matching the front end search.
             </p>
 
+            <h3>When filtering by date:</h3>
             <p>
-                Date fields apply to the <em>modification date</em> of casebook content,
-                but to the <em>date of last login</em> for professors.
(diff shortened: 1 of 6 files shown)


rebeccacremona (approved):
LGTM!

I love how nuanced this reporting is getting 🎉

PR #1674, opened 2022-08-01
Add a pathway to generate an HTML rendition of the complete casebook in the browser

lizadaly:
For now only available to admins and requires a `LiveSetting` to be enabled, as this work is incomplete:

Creates a new endpoint, `as-printable-html`, that renders the entire casebook in the browser applying styling from Paged.js to generate a printable view. 

*Contents of this PR*

* The view that generates the page—this just pulls the casebook and all its child nodes and passes it to a new set of templates.
* Templates in `export/as_printable_html`—these are derived from the intermediate HTML templates in the pandoc pipeline, but use more semantic HTML and readable class names. 
* A new Vue bundle of SCSS that applies only to the printable HTML view. There is more customization to be done here, especially for CAP-derived and non-HTML resources. `pagedjs.scss` is directly from Coko; they want you to vendor it locally.
* Some tests that largely just validate that the views are protected as expected
* A migration to add the new field to `LiveSettings`.

At the moment the Paged javascript is imported in the HTML head; in the future we'll probably want to include that directly in the build but for the moment I'm using a beta version of the library. 

There's no pagination or optimization applied here; for a full casebook the content will be far too long to render in someone's browser, especially over a network. We'll certainly want to chunk this and probably limit it to a single top-level section at a time. 

I've committed the build artifacts here because the view tests need the new `casebook_html` bundle available to work.

**The LiveSetting will need to be toggled in the admin to enable this in a given environment, even locally**

image

My expectation is that this will remain off in production for the time being. If this fails to ship for whatever reason, the feature and setting should be removed.


lizadaly:
Moving back to draft to update the perms test to better fit the expected pattern.

lizadaly:
Ready for review!

bensteinberg (commented):
- I like the TBD.
- You don't need to commit webpack-stats.json, though it should be a no-op.
- Loading fonts from Google made me look back -- apparently we've been doing this in the base template since 2019, as well as loading CSS from fontawesome, unlike in our other projects. I don't know if this is the moment to revisit that.

lizadaly:
The `tbd.html` template came from the existing pipeline! I'm not sure what triggers it.

My understanding is that hot-linking from Google is no longer as beneficial as it used to be and overall it's better to self-host assuming you're behind a CDN of some kind. In this particular case it's more a placeholder since people rightfully have lots of opinions about fonts so I dunno what we'll end up using for the print view, in the end.

lizadaly:
For the purposes of this PR, maybe it's best if I just drop the font declaration and import for now?

lizadaly:
@bensteinberg This is ready to re-review!

bensteinberg (approved):
👍

Issue #1626, opened 2022-07-14
Improve publishing experience


cath9:
Make publication more of an event, and encourage certain next steps. 

New language and UI along the lines of:

"Congratulations! Your Casebook is live! Use this link to share your book with your students [insert link]. 

We also encourage you to share your work with your colleagues or on social media! Your book is an important contribution to open education for law and we hope you'll help others find it and use it in their own classrooms. Because you are a verified professor, your book will automatically be surfaced on H2O's search page and will appear in web searches."

Internal notification to me and anyone else interested would also be great so I can follow up. 

Someday would be great to also prompt a pre-publication checklist (Want to add a book cover? Apply tags? Add a description?) but there's obviously more background work that would need to happen first.

PR #1675, opened 2022-08-01
Add publish notice

dongkyunc:
# What

image

# How

After publishing, the redirect will append a `published` to the URL querystring. Django will check for that and display the notice if it is there.

# Why

Addressing https://github.com/harvard-lil/h2o/issues/1626

# Note the reviewers

This is triggerable by someone else just manually appending `?published` to the casebook URL, but I don't think that should be too much of an issue?


web/main/views.py
```diff
@@ -1519,6 +1520,8 @@ def get(self, request, casebook):
                 "tabs": casebook.tabs_for_user(request.user),
                 "casebook_color_class": casebook.casebook_color_indicator,
                 "contents": contents,
+                # .get is a dirty test hack
+                "abs_url": request.META.get("HTTP_HOST", "") + casebook.get_absolute_url(),
```
dongkyunc: Is there a way to add this to the mock?

web/main/templates/casebook_page.html
```diff
@@ -27,6 +27,35 @@
...
+                            <button onclick="navigator.clipboard.writeText('{{ abs_url}}')">
+                            Click me to copy
+                            </button>
+                        </div>
+
+                        We also encourage you to share your work with your
+                        colleagues or on social media! Your book is an important
+                        contribution to open education for law and we hope you'll
+                        help others find it and use it in their own classrooms.
+                        Because you are a verified professor, your book will
```
lizadaly: This text says "because you're a verified professor", so the entirety of the message needs to be behind a check for whether that's true.

lizadaly (changes requested):
Thanks for this!

* This text is intended, I believe, to only apply to verified professors, so for sure it'll need to be behind a check for that state.
* Django natively supports showing a response after an event, using the [Django messages framework](https://docs.djangoproject.com/en/3.2/ref/contrib/messages/). We do make use of this in the code ([template](https://github.com/harvard-lil/h2o/blob/f3d28884a58f41853adabb81116b8130d54e7f02/web/main/templates/base.html#L50-L57), [view](https://github.com/harvard-lil/h2o/blob/6fe82d13f85446a36c40c558e0a3083450c728a4/web/main/views.py#L1040-L1046)] already, so I think rather than rolling a new thing, we should leverage the existing feature.

PR #1677, opened 2022-08-02
Add new desired local hostname

lizadaly:
When accessing the site using `opencasebook.test` per the README, Chrome was showing an infinite loop of "Invalid Host/Origin" in the console. 

This adds that hostname following the pattern done to CAP in https://github.com/harvard-lil/capstone/commit/e8fa291ca3d3b698ec9c4ef99e9a62005ac74233

Also deletes some apparently unneeded config.


matteocargnelutti (approved):
Good call! LGTM 😄

PR #1683, opened 2022-08-02
Remove python-docx from main app

lizadaly:
This library is in use in the python-lambda part of the application, but that has its own `requirements.txt` and  gets installed as part of the function instantiation. The library is not in use in the main part of the application and can be removed.


rebeccacremona (approved):
Thanks!

PR #1690, opened 2022-08-05
Drop archived description field from textblock; fixes #1045

### LINKED ISSUE #1045: TextBlock legacy fields

lizadaly:
Dropping this old field after archiving its data as a spreadsheet.


lizadaly:
It turns out this field is actually used in the in-casebook search, which is maybe not that useful if it's underpopulated, but it's a bigger change than this PR, so closing for now.

PR #1692, opened 2022-08-08
Fixing Sort by Dropdown in Main Search

shljessie:
Fixing the sort by features in the main search tab

1. Sort by Date 
sort by date from **most recent- least recent** 

Before
Screen Shot 2022-08-08 at 3 25 55 PM

After
Screen Shot 2022-08-08 at 3 27 36 PM

2. Remove sort by Relevance
Sort by relevance and sort by title features look the same in the search results.
The code uses the '-rank' to sort by relevance. but 'rank' does not seem to be defined.
So I commented out the relevance for the meantime. 

Screen Shot 2022-08-08 at 3 28 54 PM

### DIFF
--- a/web/main/templates/search/filters.html
+++ b/web/main/templates/search/filters.html
@@ -21,9 +21,7 @@
       <select class="form-control select required" name="school" aria-label="Filter by School" id="search_school">
         <option {% if not request.GET.school %}selected {% endif %}value="">All Schools</option>
         {% for school in facets.affiliation %}
-            {% if school is not None %}
-              <option {% if request.GET.school and request.GET.school == school %}selected {% endif %}value="{{ school }}">{{school}}</option>
-            {% endif %} 
+          <option {% if request.GET.school and request.GET.school == school %}selected {% endif %}value="{{ school }}">{{school}}</option>
         {% endfor %}
       </select>
     </div>
--- a/web/main/templates/search/filters.html
+++ b/web/main/templates/search/filters.html
@@ -3,8 +3,8 @@
     <div class="form-group select required search_sort">
       <select class="form-control select required" name="sort" aria-label="Sort Results" id="search_sort">
         <option {% if request.GET.sort == 'display_name'%}selected {% endif %}value="display_name">Sort by Title</option>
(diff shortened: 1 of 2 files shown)


web/main/templates/search/filters.html
```diff
@@ -4,7 +4,7 @@
       <select class="form-control select required" name="sort" aria-label="Sort Results" id="search_sort">
         <option {% if request.GET.sort == 'display_name'%}selected {% endif %}value="display_name">Sort by Title</option>
         <option {% if request.GET.sort == 'decision_date'%}selected {% endif %}value="decision_date">Sort by Date</option>
```
lizadaly: I think this line contains the actual problem to be fixed—

`decision_date` applies to cases imported from CAP. I assume it's the date of the actual decision in the case itself.

Casebooks have `created_at` dates—when the book itself was first created. 

I think the issue is that the casebook "sort by date" field needs to be `created_at`, not `decision_date`. But keep your change in `models.py` since that's what allows searching by legal document to sort by date correctly for cases.

lizadaly (changes requested):
I'm not sure that 'relevance' doesn't work (or at least, that it does nothing). Compare these two results:

Search for "school" sorted by relevance: https://opencasebook.org/search/?type=casebook&q=school&sort=score&author=&school=

Search for "school" sorted by title: https://opencasebook.org/search/?type=casebook&q=school&sort=display_name&author=&school=

`rank` is coming from the Postgres search code, I believe: https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/search/#searchrank

So I think the date search is definitely broken, but should be fixed in a different place, and I think relevance is probably working, but sorting by relevance is just very opaque and sometimes doesn't provide useful ordering.

web/main/templates/search/filters.html
shljessie: Great! Do you suggest I update filters.html with `created_at`? I have uncommented the relevance search for now!

lizadaly: Yeah!

lizadaly (approved):
Good with me assuming you checked that this also fixes the date sorting! Thank you!

shljessie:
Yup it does! :)

Issue #1691, opened 2022-08-08
Add admin page for editing of Series (CommonTitle) lists


lizadaly:
There's no way for us to make changes on behalf of users on Series because that model isn't currently reflected in the admin.

We think there's a bug in the Series implementation right now so it's somewhat pressing to be able to go in and remove extraneous items while we spend a little time triaging the problem.

Create an admin page that lists all Series and allows adding and removing titles from it.

PR #1693, opened 2022-08-09
[#1691] Admin page for editing casebooks in a series

lizadaly:
There's currently no way for admins to edit a Series; this allows a minimal amount of functionality to clean up Series with erroneous books added, which seems to have occurred for as-yet unknown reasons.

This adds a page to the admin for Series (the database model here is called `CommonTitle`) and allows admins to remove casebooks from them, as well as see how many books are in each series.

The list page looks like this:

image

Clicking on an individual Series will show all the books in it:

image

**About the Series functionality**

The data model for `CommonTitle` benefits from some explication:

* Each `CommonTitle` has a foreign key to exactly one Casebook, the `current` casebook. This corresponds to the most-recent edition.
* In addition, any casebook instance may have its `common_title` field populated with a foreign key to `CommonTitle`, in which case it's considered to belong to that Series.

There's no database-level coupling between `CommonTitle.current` and `Casebook.common_title`. The relationship is set up by the (two!) `CommonTitle` serializers when a series is grouped in the user's Dashboard. I'm not sure what would happen if the `current` Casebook had its `common_title` field nulled out—I think the feature would be in an inconsistent state.

As a result this admin is pretty custom.

(If we re-implemented this from scratch, I think we'd probably want to consider a [many-to-many through model](https://docs.djangoproject.com/en/3.2/topics/db/models/#extra-fields-on-many-to-many-relationships), and use one of the extra fields to designate the "current" title. Then I think we could delete most of this code because Django would natively render an admin interface that allowed for full editing.)
 

**Testing**

I used an admin inline to render the list of titles with a custom queryset method and an overridden delete method. Normally a model inline wants to _edit_ the represented model, but we really really don't want to delete casebooks, just the relationship between a casebook and a series, so I tried hard here to ensure that the delete feature is non-destructive.


web/main/test/test_admin.py
```diff
@@ -0,0 +1,49 @@
...
+
+    data = {
+        "casebooks-0-DELETE": "on",
+        "casebooks-0-id": casebook1_in_series.id,
+        "casebooks-0-common_title": series.id,
+        "casebooks-1-id": casebook2_in_series.id,
+        "casebooks-1-common_title": series.id,
+        "casebooks-TOTAL_FORMS": 2,  # Fields required for formset validation
+        "casebooks-INITIAL_FORMS": 2,
+    }
```
lizadaly, line 37: This looks gross but I swear this is the recommended way to test inlines/formsets.

lizadaly:
I'd love an actual hands-on test on this just to have another person verify that casebooks themselves can't be deleted using this tool.

web/main/templates/admin/main/casebook/inline_series.html
```diff
@@ -0,0 +1,68 @@
+{% load i18n admin_urls static admin_modify %}
+{# Based on https://raw.githubusercontent.com/django/django/stable/3.2.x/django/contrib/admin/templates/admin/edit_inline/tabular.html #}
+
```
rebeccacremona: Tiniest of nits: indentation seems a little tangled in this template.

web/main/admin.py
```diff
@@ -843,6 +844,55 @@ def live_annotations_count(self, obj):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.queryset = self.model.objects.filter(common_title_id=self.instance.pk).exclude(
+            pk=self.instance.current.pk
+        )
+
+    def delete_existing(self, obj: Casebook, **kwargs):
+        """Drop this item from the Series, but don't actually delete it"""
```
rebeccacremona, line 860: What would you think of dropping a link to https://github.com/django/django/blob/e03cdf76e78ea992763df4d3e16217d298929301/django/forms/models.py#L762-L765 here?

rebeccacremona (commented):
This seems fine to me! I confirm that I cannot use this to delete casebooks.

Part of me wonders whether it might be safer to like... add a custom field to the formset for "remove from series", rather than override the built-in delete field to serve this purpose, and then hooking into the save method, but I haven't thought through the details: does that sound more difficult/too difficult/less clean to you?

lizadaly:
> Part of me wonders whether it might be safer to like... add a custom field to the formset for "remove from series", rather than override the built-in delete field to serve this purpose, and then hooking into the save method, but I haven't thought through the details: does that sound more difficult/too difficult/less clean to you?

Yeah let me try that. I wasn't sure how to go about that when I first started the work but now that I've been in the depths of the inlines I think I can make it work, then switch the inline to `can_delete = False`, which does what it sounds like.

lizadaly:
There may not be a better way to do this given the current data model:

* You can't add a custom field that is also editable to an Inline form; Django immediately checks the underlying model and complains if the field doesn't exist. You _can_ add a custom field that's in the `readonly` property, but then you have to hack the model's `clean` methods to allow it to process a field that's otherwise supposed to be uneditable.
* I tried being clever and just adding `current_title` as an editable property of the form, since hey that's the field I want to edit! But Django knows it's the PK of the parent `CommonTitle` model implicitly and just ignores it even if it's present in `fields`.  
* I could replace the delete checkbox with a totally custom JS-based widget that drops the relationship with an `onClick` handler that pings the frontend's API. I think that's only worth doing if we really want to avoid the current solution.

rebeccacremona:
> There may not be a better way to do this given the current data model

Cool cool. This seems safe enough as is. I suppose the danger is, this breaks during a Django upgrade, but the test should catch it! Seems okay to merge to me!

PR #1694, opened 2022-08-09
Main Search Casebooks Styling Change

shljessie:
Adding Hover Effects to Search Page Styling, Fixing up Search Button, Tab Hover Effects 
All Hover Effects are added to Focus too!

https://user-images.githubusercontent.com/59305253/183671932-8b9b8181-93b0-4186-87dd-9ebe227d84df.mov


lizadaly:
This looks nice! I like that the header gives us a place to add other calls-to-action or links to help material if we want. 

Can you double-check that everything looks reasonable at a mobile screen width? I know the site isn't especially optimized for mobile but just want to make sure the new layout doesn't break it on narrow viewports.

shljessie:
Added Responsiveness! 
Also wondering if we could change the search naming to Search Casebooks. 
As I'm doing the User Study, people are confusing `Search` with `Search Inside` and tend to click on the `Search` to  search for cases inside a casebook.

https://user-images.githubusercontent.com/59305253/183699447-430933a8-61a6-412f-9c21-34f48bbb5573.mov

shljessie:
Screen Shot 2022-08-09 at 11 53 02 AM

shljessie:
Before and After views! 

Before
Screen Shot 2022-08-09 at 11 53 59 AM
Screen Shot 2022-08-09 at 11 54 06 AM
Screen Shot 2022-08-09 at 11 54 24 AM


After
Screen Shot 2022-08-09 at 11 54 49 AM
Screen Shot 2022-08-09 at 11 54 37 AM

lizadaly:
Wow that's so nice!! I think you might need to resolve that conflict which I think comes from your other PR, but after that happy to approve this.

PR #1695, opened 2022-08-09
Removing Button Focus Outline on Mouse Click

shljessie:
Removing button focus gray outline on mouse click and only enabling it for keyboard focus
`focus-visible` will only create focus effects for keyboard focus

Screen Shot 2022-08-09 at 10 34 01 AM
Screen Shot 2022-08-09 at 10 34 21 AM

Before 

https://user-images.githubusercontent.com/59305253/183676372-2809a020-7695-428e-b922-4b725d9244e7.mov

After

https://user-images.githubusercontent.com/59305253/183676363-17c28c4b-959b-4596-be6f-abd4d72c8351.mov

### DIFF
--- a/web/frontend/styles/casebook.scss
+++ b/web/frontend/styles/casebook.scss
@@ -65,7 +65,7 @@ main > header.casebook {
       &.disabled {
         background-color: $light-gray;
       }
-      &:focus {
+      &:focus-visible {
         @include generic-focus-styles;
         z-index: 999;
       }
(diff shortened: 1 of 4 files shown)


matteocargnelutti (approved):
Looks good to me @shljessie! Appreciate the attention given to keyboard navigation 😄 .

**Something we might want to look into (later?):** saying _"Always hide outline for `:focus`, but not for `:focus-visible`"_ early on in the stylesheet, let the cascade do its thing, and only override as needed, as suggested by experts such as Lea Verou:
https://twitter.com/leaverou/status/1045768279753666562

Anyway, good to go. Thanks for this 🚀 .

PR #1700, opened 2022-08-10
Add publish notice

dongkyunc:
# What

image

# How

Adding a small check and additional text to the pre-publish modal.

# Why

Addressing https://github.com/harvard-lil/h2o/issues/1626

# Note the reviewers

- I'm not sure this is the proper way to pass data all the way from Django to the Vue component.
- This adds FontAwesome to our dependencies
- This is a mild copy of  #1675 but I borked git in an odd way.


web/main/views.py
```diff
@@ -1702,13 +1702,13 @@ def create_draft(request, casebook):
...
 @hydrate_params
-def edit_casebook(request, casebook):
+def edit_casebook(request, casebook: Casebook):
     """
     Given:
     >>> private, with_draft, for_verified_prof, client = [getfixture(f) for f in ['full_private_casebook', 'full_casebook_with_draft', 'full_private_casebook_for_verified_prof', 'client']]
     >>> draft = with_draft.draft
 
-    Users can edit their unpublished and draft casebooks:
+    Users can edit their un'apublished and draft casebooks:
```
lizadaly:
```suggestion
    Users can edit their unpublished and draft casebooks:
```

lizadaly (changes requested):
There's a feature flag in settings, `COVER_IMAGES`, that determines whether the cover image functionality is enabled. Right now it's off—can you modify this to check it and if it's not enabled, don't display the cover image note?

web/frontend/styles/casebook.scss
```diff
@@ -279,6 +279,45 @@ form.edit_content_resource, form.edit_content_section, form.edit_content_caseboo
...
+    }
+
+    div.link {
+      display: block;
+      text-align: center;
+      margin: 1em 0 1em 0;
+
+      a { 
+        background-color: #F7F6EE;
+        // color: #333333;
```
lizadaly: Should this just be deleted?

web/main/views.py
dongkyunc: Thanks for the catch, errant vim keystrokes 😅 

Frontend now checks for the feature flag

Addressed in [d48ded4](https://github.com/harvard-lil/h2o/pull/1700/commits/d48ded4bbefb3f94624f711299741197a3078d08).

lizadaly (approved):
Thank you! I want @cath9 to weigh in on the specifics of the copy, and if you get a chance, finding a way to reduce the duplication in the `publish_check` JSON object that (I think?) is the same in all those cases would be a nice addition, but I don't think that needs to block this!

web/frontend/styles/casebook.scss
dongkyunc: Addressed in [a3fb036](https://github.com/harvard-lil/h2o/pull/1700/commits/a3fb036b8c96410f3d3b8af97c96802fc2c12bd9)

cath9:
I think ideally this would be a two step process. See below. 

Pre-Publication Pop-Up:

"You're almost ready to publish your book! The following elements are optional, but can make your book more discoverable by students and colleagues:

XDescription (is there a link to where they could add this)
XCover Photo (is this enabled yet?)"

[Buttons]

Post-Publication Pop-Up:

"Your book is live on H2O! Because you are a verified professor, your book will automatically be surfaced in H2O's collection and will appear in web searches. 

Use this link to share it with your students: [link]

We also encourage you to share your book with colleagues and on social media! You've made an important contribution to open education for law, and we hope you'll help others discover your work and build from it."

dongkyunc:
The two-step was closer to the original goal, however, some technical issues arose (https://hlslil.slack.com/archives/C4FE9FL56/p1659986418755369) it seems we're mostly limited to the pre-publication pop-up.

lizadaly:
Chatted with Catherine; this is okay to go once the json conflict is resolved!

lizadaly (approved):

PR #1701, opened 2022-08-10
Add mypy to lints

dongkyunc:


lizadaly:
Closing this but going to break it up into a few different PRs—bringing `mypy` up to the most recent version, correcting types, and then adding the GH action!

Issue #1318, opened 2021-03-08
Enable adding sections that are only available to other verified professors


cath9:
Professors have asked me fairly regularly about ways to add notes to their books that would only be visible to other professors. This would include any information, guidance, notes about pedagogy, etc. that helps a professor adopt another's open casebook.

lizadaly:
* The section should be clearly marked differently from other sections
* Visiting the section as a non-verified-professor user should (probably?) return a 404.
* To avoid hassles with gaps in numbering, the sections should probably only be available at the front or back of the casebook. [edit: this is fine, they should marked as `does_display_ordinals=False`]

cath9:
Consider including attachments. Some folks have mentioned including a syllabus, exams, course proposal for the Dean, etc.

lizadaly:
Collaborators should probably be able to add and see this material, especially since that's the kind of thing an RA might be in charge of entirely.

PR #1702, opened 2022-08-11
[#1318] Add a field for instructional material available only to professors

lizadaly:
Adds a field to the `ContentNode` table to indicate that this node should only be available to professors and collaborators. In practice this will always be a `TextBlock` custom content resource because that's where we intend to add the user-facing option, as this is intended to support teaching notes, quizzes, and other educator material.

This necessitates a range of changes throughout the code because prior to this we had no node-level permissioning. All permissions were at the Casebook level, so while there were stub `viewable_by()` methods on content nodes, they always delegated to the parent casebook. Now they have an additional check, which in some cases required passing through the `request` to get at the current user.

The intention is that non-educators should be unaware that instructional materials are present in the book. This means ensuring that any method that exposes the shape or content of the book needs to incorporate a check on whether a given node should be viewable by that user. I think at a minimum this means the following sections need to be aware of the role of the current user and their relationship to the book:

- [x] The website TOC 
- [x] The TOC API 
- [x] The new next/previous navigation needs to skip over hidden nodes
- [x] Any direct URL to the node needs to either 404 or return Forbidden (I went with the latter; thoughts?)
- [x] The DOCX export view 
- [x] The printable HTML view

I tried to implement the following authorization tiers:

* Any verified professor can always see instructional content 
* Any collaborator on the current casebook can see instructional content 
* Any logged in user who is not a current collaborator or a professor *cannot* see instructional content 
* Any anonymous, non-logged in user *cannot* see instructional content 

**To think through and discuss**

What should happen if you clone one of these books? 

If the user is a verified professor, there's no problem; they had the right to see the content anyway.

If not, the user will (I think!) implicitly be a collaborator on the cloned casebook and therefore be able to see the instructional material in their own copy. If you're a student, congrats, you can now see the quiz answers! (I have not yet tested cloning behavior and would prefer to do that in a follow-up PR based on what we decide here.)

1. This can be prevented by restricting the feature to verified professors in all cases, but since many professors have help creating their casebooks, this will severely restrict the utility of the feature.

2. We could not allow teaching notes to be cloned, which also seems very undesirable since the point of the feature is to foster a community of practice among instructors.

3. We could only allow teaching notes to be cloned when the user who initiated the clone is a verified professor, otherwise strip them. In this scenario we might opt to note, when the user initiates a clone action, that material is on the book that is only available to verified professors and suggest they reach out to us to get verified.

Any other ideas? 

**Current status**

The goal of this PR is to get the field on the model and in the admin and allow us to assess what other places need to be modified to include the permission check before adding the feature to the front end. 

If this PR is successful, it should be a complete no-op on the current site. The instructional content flag will only be available to users with access to the admin (which we can use to test sample content at our leisure.)


web/main/models.py
```diff
@@ -2077,41 +2098,61 @@ def contents(self):
         res = ContentNode.objects.filter(**filter_map).exclude(id=self.id)
         return res
 
-    def get_previous_and_next_node_urls(self) -> tuple[Optional[str], Optional[str]]:
+    def get_previous_and_next_nodes(
```
lizadaly, line 2101: When I was adding role awareness to the next/previous arrows I thought that this feature was useful enough that it should more generally return the next and previous _nodes_ , and then turn those into URLs in a separate method, so that's mostly what happened here.

web/main/models.py
```diff
@@ -2403,8 +2444,22 @@ def _delete_related_links_and_text_blocks(self):
     def get_slug(self):
         return slugify(self.title)
 
-    def viewable_by(self, user):
-        return self.casebook.viewable_by(user)
+    def viewable_by(self, user: User) -> bool:
+        """A node may belong to a viewable casebook, but itself not be viewable because of professor-only content"""
+
+        if not self.casebook:  # An orphaned node isn't viewable, presumably
```
lizadaly, line 2450: This is mostly a sop to the type checker. A `ContentNode` does not actually need to belong to a `Casebook` according to the data model (it's `null=True`), though in practice I'm not sure if that happens. Maybe it's for cloning by node?

web/main/models.py
```diff
@@ -2425,20 +2480,6 @@ def is_legacy_casebook_node(self):
     def is_resource(self):
         return self.resource_id is not None
 
-    @property
-    def sub_sections(self):
```
lizadaly, line 2429: This didn't seem to be called anywhere.

web/main/views.py
```diff
@@ -115,8 +115,8 @@
 ### helpers ###
 
 
-def login_required_response(request):
-    if request.user.is_authenticated:
+def login_required_response(request: HttpRequest, always_raise=False):
+    if request.user.is_authenticated or always_raise:
```
lizadaly, line 119: Without this, attempts to hit a professor-only resource by an anonymous user would prompt to log in rather than return Forbidden. For consistency with the logged-in but not authorized experience, I went with forcing a 403 rather than a login prompt here.

web/reporting/models.py
```diff
@@ -15,10 +15,10 @@ def most_recent_casebook_title(self) -> str:
         return ""
 
     @property
-    def most_recent_casebook_modified(self) -> str:
+    def most_recent_casebook_modified(self) -> Optional[datetime]:
```
lizadaly, line 18: An unrelated type fix.

web/main/models.py
```diff
@@ -2021,12 +2024,33 @@ class ContentNode(
...
+
     tracked_fields = ["headnote"]
 
     # Stores the number of ‘read’ characters in a content
     # Length of content, less elided text and html tags.
     reading_length = models.IntegerField(null=True)
 
+    is_instructional_material = models.BooleanField(
+        default=False,
+        help_text="This content should only be made available on the frontend to verified professors, staff, or users with editing privilege",
```
lizadaly, line 2036: This help text is a lie; the feature is not currently available to staff/admins. It's an omission but it's a useful one since it lets us easily toggle between professor/non-professor states without having to drop our admin privilege. When it gets closer I'll have the behavior match the docs.

cath9:
@lizadaly how hard would it be to restrict it only to Verified Professors for now, and then potentially add collaborators in future? I ask for a few reasons. First, while you're right that restricting access to only Verified Professors would restrict the utility, I think that might not be a bad way to pilot this and give us a chance to maybe ask for some beta testing to help us think through these scenarios. Second, in most cases, collaborators are either other professors or TA's and I think that would be fine, but some professors are working with large numbers of RA's who are often current students to build books collaboratively, and I don't think they/we would want those kinds of collaborators to have access. Third, this also relates to a conversation that's been coming up in the background for years about whether we want to create different personas in H2O. Would we want to allow professors to designate support staff who would have access to teaching notes? Would we want to have Student accounts that would never have access? etc.

lizadaly:
@cath9 OK great, that definitely simplifies things. I think ultimately more user roles will be valuable for avoiding weird edge cases.

rebeccacremona:
> Any other ideas?

Another idea, probably too complex, pops to mind apropos of https://github.com/harvard-lil/h2o/pull/1703. Maybe we are moving towards having two different actions: clone for annotation, and clone for adaptation. If cloning for annotation (as a student), secret materials are stripped out so that neither they nor any collaborators they add can see secret stuff. If cloning to adapt (as a verified prof), secret materials are retained, and they and any collaborators they add can read and adapt. If a person wants to clone for adaptation and are not a verified professor... they can request that status.

That adds a lot of complexity, but is something like what I would expect, if I were using a feature that I thought was only for professors: if students could just clone to see my prof-only notes, I think I'd find that a bug.

Hmmm. Are there simpler ways.....

web/main/models.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
rebeccacremona, line 1: Huh, we need this with 3.9?

lizadaly:
> Maybe we are moving towards having two different actions: clone for annotation, and clone for adaptation. If cloning for annotation (as a student), secret materials are stripped out so that neither they nor any collaborators they add can see secret stuff. If cloning to adapt (as a verified prof), secret materials are retained, and they and any collaborators they add can read and adapt. If a person wants to clone for adaptation and are not a verified professor... they can request that status.

This is what I was thinking too, and I think it just falls out of the user type rather necessarily mattering which type of cloning behavior they chose? 

I don't think this PR is dependent on this decision since there's no way to add these secret materials in the front-end yet—the redacting logic can be a prereq for adding the frontend control. though.

web/main/models.py
```diff
@@ -2021,12 +2024,33 @@ class ContentNode(
...
+        if user:
+            if user.is_authenticated and (
+                user.verified_professor
+                or casebook.contentcollaborator_set.filter(user=user).exists()
+            ):
+                return ContentNode.objects.filter(casebook=casebook, **kwargs)
+            return ContentNode.objects.filter(casebook=casebook, **kwargs).exclude(
+                is_instructional_material=True
+            )
+        return ContentNode.objects.filter(casebook=casebook, **kwargs)
```
rebeccacremona, line 2052: I'm a little confused by the signature of this method. (I know the type is propagating down from higher up the call chain; commenting here for simplicity.) With `user: Optional[Union[User, AnonymousUser]]`, I would have expected that meant there would be a default `user=None`, so that we could call the method without specifying a user. In what situations do we anticipate calling it intentionally with `None`?

Also just double checking: if we *do* call it intentionally with `None`, should the fallback be to return all the nodes, as though an admin/qualifying user were passed in (as written), or should the fallback be more restrictive, showing only public nodes?

web/main/models.py
rebeccacremona: Love it.

web/main/models.py
rebeccacremona: A `ContentNode` that does not belong to a casebook is a relic of times past, when there was no `Casebook` model: those nodes were the top-level of a a casebook, the root, a Casebook-flavored `ContentNode`, as opposed to a Section- or Resource-flavored node. It's what the [`old_casebook`](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3474) field on the `Casebook` model points too. We checked recently; as of that check, that accounted for all `ContentNode` objects that didn't belong to casebooks except one. That singleton was clearly a messed up thing made by a programmer in the shell, and not an artifact of any real book.

I hope to clean this up soon.

rebeccacremona: This is a thing I'm hoping to clean up soon.

web/main/models.py
```diff
@@ -2403,8 +2444,22 @@ def _delete_related_links_and_text_blocks(self):
...
+            return False
+
+        # Doesn't matter what the permissions are on this node if the casebook itself is not viewable
+        if not self.casebook.viewable_by(user):
+            return False
+
+        if self.is_instructional_material:
+            return user.is_authenticated and (
+                user.verified_professor or user in self.casebook.all_collaborators
+            )
```
rebeccacremona, line 2460: Above in `nodes_for_user_by_casebook` you did:
```
            if user.is_authenticated and (
                user.verified_professor
                or casebook.contentcollaborator_set.filter(user=user).exists()
            ):
```
similar to this. I wonder if there's a way to abstract out; maybe not worth it...

web/main/models.py
rebeccacremona: Yeah, I see how it stopped being useful subsequent to the model refactor.

web/main/models.py
```diff
@@ -3683,41 +3725,44 @@ def _delete_related_links_and_text_blocks(self):
             cls.objects.filter(id__in=ids).delete()
 
     @property
-    def sections(self):
-        return Section.objects.filter(casebook=self).filter(resource_type__isnull=True)
+    def sections(self) -> models.QuerySet[Section]:
+        return Section.objects.filter(casebook=self, resource_type__isnull=True)
```
rebeccacremona, line 3729: I think this method is wrong now (and has long been): there are now Sections that have `resource_type='Section'`, if I recall... but in any case, I don't think this syntax change makes it more wrong 😂

web/main/models.py
```diff
@@ -3683,41 +3725,44 @@ def _delete_related_links_and_text_blocks(self):
...
 
     @property
-    def resources(self):
+    def resources(self) -> ContentNodeQuerySet:
         return ContentNode.objects.filter(casebook=self, resource_id__isnull=False)
```
rebeccacremona, line 3733: Same, I think this method is wrong now... Something else I hope to clean up soon :-)

rebeccacremona (approved):
I am excited about this feature!!! And oof, what a can of worms.

This looks good and safe to merge to me; I asked a couple minor questions and added a little historical comments where I could.

web/main/models.py
lizadaly: Yeah, to be able to declare a type that is defined in the same file.

web/main/models.py
lizadaly: It's confusing because it's wrong!

Earlier in the code it was `user: Optional[User] = None` with a different method name. Removed `Optional` in both cases where it appeared.

web/main/models.py
lizadaly: Yeah it was a tossup for me too. The number of methods on these models is so overwhelming I went with inlining it, but if it has to be added a third time it should probably live somewhere else.

web/main/models.py
lizadaly: https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations (was set to become the default in 3.10 but deferred.)

web/main/models.py
rebeccacremona: Aha!

I think we can lose the `if user` check, then, too? Maybe in a followup PR.

PR #1703, opened 2022-08-11
Emphasizing Annotation Functions

shljessie:
Making the Annotation Function Clear for non-verified professors ( students, RAs ) 

**Before**
Currently, in order to annotate a casebook, students have to go through the user flow of 

1. Creating a Notebook(same as creating a casebook) to store their Annotations 
2. Clone the case into the notebook
3. Scroll down to Click on content within the Annotation

Or cloning the entire book and repeating process 2,3 

Demonstration 
Screen Shot 2022-08-15 at 12 13 00 PM

This is difficult to do because it is hard for first-time users to make a connection between "Clone" and "Annotate" on their own.

**New**
Essentially, Annotating a Casebook Means that a user will clone the entire casebook or clone a case within the casebook into their notebook. 
* The functionalities are the same as Clone 
* The UI will help the users create a mental model of understanding how the process of annotation works on H2O - by cloning a case/casebook. 

Adding an Annotation Button that will be visible only to non-verified professors!

Screen Shot 2022-08-15 at 3 03 38 PM

**_Annotating Entire Casebook_**

https://user-images.githubusercontent.com/59305253/184698767-f9feabfa-7ded-4300-98b4-97c4b8eaa930.mov

**_Annotating Part of a  Casebook_**

https://user-images.githubusercontent.com/59305253/184699314-92c0419a-4329-4d34-9e2f-073c48bbf887.mov

**_Reducing Size of textboxes, adding message_**
Screen Shot 2022-08-15 at 3 35 02 PM


shljessie:
**_Annotating a Case, Section, Link, Text_**
https://user-images.githubusercontent.com/59305253/184698497-4e375248-bf04-4d16-a619-08f8a5987fcc.mov

web/frontend/components/AnnotationCloner.vue
```diff
@@ -0,0 +1,129 @@
...
+    var cookies = document.cookie.split(';');
+    for (var i = 0; i < cookies.length; i++) {
+      var cookie = jQuery.trim(cookies[i]);
+      if (cookie.substring(0, name.length + 1) === (name + '=')) {
+        cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+        break;
+      }
+    }
+  }
+  return cookieValue;
```
lizadaly: I know this pattern is already in the code but there's:

```
export function get_csrf_token()
```

in `frontend/legacy/lib/helpers.js`. Do you think you could use that instead?

shljessie: Should I be importing the function like this to use it? The button doesn't seem to work when I do this though 🤔
Screen Shot 2022-08-15 at 3 25 42 PM

lizadaly: It doesn't take any arguments because it already knows what cookie name to look for. Does it work if you omit the argument? If not, don't worry about it and put it back the way you had it!

lizadaly: Oh! Change the import to:

```
import { get_csrf_token } from "../../frontend/legacy/lib/helpers";
```

shljessie: great it works! I think this PR is ready now! :)

lizadaly:
A couple things related to the specific permissions available:

My user is a verified professor, but I still see it:

image

Here's my status in the admin:

image

Can you verify the same behavior?

Moreover, this is actually my casebook (note how "Revise" shows up), so I think even if I weren't a verified professor, it doesn't make sense for me to see the Annotate button on my own casebook. I think people will just keep clicking on "Annotate" making new clones thinking they were going to start annotating!

**Not a bug**

If I'm not logged in at all, I don't see the button—which I think is correct for now. But maybe in the future we want it to show up and when clicked, prompt you to create an account? My understanding is that a lot of students who use H2O never create accounts in the first place so they're never logged in.

shljessie:
@lizadaly Oops my bad fixed that error! I was going to get to it and totally forgot!😅 

Seems like if you are not logged in, the only button you see is the export button right now
I could look into adding the annotation button & clone button when not logged in and prompting a logged-in

Screen Shot 2022-08-15 at 4 21 17 PM

I think it is because the cloneable is only true when the user is logged in, and if it is not cloneable, the clone and annotate button are set so that they do not show up.

Screen Shot 2022-08-15 at 4 21 27 PM

lizadaly:
Thanks! This does fix the verified professor check.

The "Click on the content below to start Annotation!" message now appears in the edit screen for any casebook modified by a user who isn't a verified professor, including if you create a casebook from scratch and there's no content yet!

image

I think maybe we want this message to only appear if you created the casebook using your new method and not in any other scenario. One way to make that happen by adding a new type of action to the Edit Log, which tracks changes to Casebooks: 

https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3429-L3441

It might be too complicated to wire that up as part of this PR, but you could comment out the messaging on the edit page for now and we could add that back in when the edit log change happens?

shljessie:
Hmm yes I will definitely look into that! 
What do you think about passing along a variable when the annotation button is clicked and only displaying that message then? Could this cause any problems in the future?
https://vuejs.org/guide/essentials/event-handling.html#inline-handlers

shljessie:
Edited to only show edit message when there are casebook contents within the book 

Screen Shot 2022-08-16 at 3 30 57 PM

The toc is the table of content values!
Screen Shot 2022-08-16 at 3 31 17 PM


Screen Shot 2022-08-16 at 3 29 44 PM

Screen Shot 2022-08-16 at 3 30 16 PM

shljessie:
This is ready for review now!

PR #1704, opened 2022-08-12
Allow navigation at the section level in printable HTML view

### LINKED ISSUE #1709: Render only one top-level ordinal and its descendants at a time
Rendering an entire casebook in the browser will be prohibitive in terms of performance both from the server side and from the end user's side—some casebooks when printed run to hundreds of pages.

Render only one top-level ordinal and its descendant children on the page.

Provide navigation items to go next/previous through the top-level ordinals. 

The navigation should only be present in the web view; it should not appear in the printed output.

lizadaly:
This amends the printable-html Casebook view to take a second argument, a "page number", which is passed to a Django Paginator query that retrieves the list of top-level sections from a casebook. A single section is retrieved and then PagedJS renders the descendent nodes as part of a `PagedJS` "book." I think with this change we could enable this feature on prod for admin users.

A screen-only floating navigational bar hangs at the top letting the user move through the sections:

image

## Loading and rendering

The bigger architecture change here is that PagedJS is now a direct dependency (though not part of the main site bundle) and we can hook into its lifecycle methods. Django populates the book content into a `<template>` element, and that's extracted and passed to PagedJS along with the CSS that's paired with the book.

PagedJS has hooks for every stage of its lifecycle and setting things up this way will let us take advantage of that for any transformations we want to apply during rendering.

It also means we could turn this page into a little mini-SPA—we could, for example, pre-populate the next and previous sections in additional templates and have the content just update instantly.

## About the styles

PagedJS works best when the CSS can be manipulated by it directly, and this does not reliably happen when the CSS is delivered as part of a JS bundle. I pulled the book customization out of SCSS/bundling and moved it to live directly as a normal Django static resource. It's a shame to lose SCSS functionality especially because I'm making use of depth selectors a lot, and maybe later I'll return to getting this to work.

Styles applying to the web interface itself like the navigation bar are still in SCSS, but also not part of the main site bundle.

## Limitations 

We lose a few things by breaking up the book arbitrarily like this:

* Page numbering is by default just implemented with a simple CSS `count()` function that now resets for each section, but PagedJS has mechanisms for more complex numbering and I think dealing with this will be tractable.
* If any casebooks have internal hyperlinks, they probably won't work, but I don't know if anyone tries to do this.
* There are PagedJS extensions for generating a table of contents and those won't work out of the box, but I think we'd need to heavily customize the TOC anyway, in which case we can render its hyperlinks however we want to be compatible with this endpoint.
* The URL scheme here is different from the standard site one—it uses these paginator indices instead of ordinals and section names. Since paginator numbering is 1-based, in practice this may not be all that different, but it's worth calling out that the integers represent different things. I don't have to use `Paginator`, but it will come in handy if we want to pre-load next/previous sections, for example.


web/vue.config.js
```diff
@@ -28,67 +28,71 @@ RelativeBundleTracker.prototype.writeOutput = function(compiler, contents) {
...
-    test: 'frontend/pages/test.js',
-    vue_app: 'frontend/pages/vue_app.js',
-    casebook_html: "frontend/pages/casebook_html.scss"
+    application: "frontend/pages/application.js",
+    rich_text_editor: "frontend/pages/rich_text_editor.js",
+    main: "frontend/pages/main.scss",
+    test: "frontend/pages/test.js",
+    vue_app: "frontend/pages/vue_app.js",
+    as_printable_html_styles: "frontend/pages/as_printable_html.scss",
+    as_printable_html: "frontend/pages/as_printable_html.js",
```
lizadaly, line 47: This is the only material change here (Prettier got excited about normalizing the quotation marks and I just let it).

lizadaly (commented):
`webpack-stats.json` needs to be included here because without the new bundle reference the tests won't pass.

rebeccacremona:
> If any casebooks have internal hyperlinks, they probably won't work, but I don't know if anyone tries to do this.

Hmm, I think I have only seen footnotes and the like; I don't think I've ever seen a person attempt to link backwards or forwards to another section/resources. Kind of funny if no one ever has!

web/main/test/test_export.py
```diff
@@ -124,10 +124,11 @@ def test_export(
...
 
 
-def test_printable_html_livesetting_required(admin_user_factory, client, casebook_factory):
+def test_printable_html_livesetting_required(
+    admin_user_factory, client, full_casebook_parts_factory
+):
     """The printable HTML view requires auth and an explicit setting at this time"""
-    casebook = casebook_factory()
-
+    casebook = full_casebook_parts_factory()[0]
```
rebeccacremona: If you want, you can use the [`full_casebook`](https://github.com/harvard-lil/h2o/blob/develop/web/conftest.py#L545) fixture instead: I think it's exactly the same as `full_casebook_parts_factory()[0]`

rebeccacremona (approved):
LGTM!

Clever use of the Django paginator! I feel like I would have done something goofy and manual there.

rebeccacremona:
Let's squash-merge this one

PR #1705, opened 2022-08-15
Ignore more directories during pytest discovery

lizadaly:
This change speeds up test discovery on local machines quite a lot by excluding the large `.mypy_cache` directory, reducing the run time to a third of the previous. It will probably have no major effect on CI runs unless those end up running `mypy` before tests at some point, though it does exclude a few other directories.


matteocargnelutti (approved):
LGTM. Good idea 🏄

Issue #1538, opened 2022-03-09
Export ratelimiting code tweak


rebeccacremona:
We suspect https://github.com/harvard-lil/h2o/blob/1a70a2df63e556d8962501007fa457a8e5ff24da/web/main/models.py#L4090 needs a tweak, to either get a read lock when it starts, or to do it's update with a `F()`, so that if many export threads are going at once, the value increments properly..... but we need to think about it some more.

PR #1706, opened 2022-08-15
Run pytest with the concurrency plugin enabled

lizadaly:
Enables tests to be run concurrently using `pytest-xdist`. Depending on the number of cores on the machine, this reduces test time run by 30% or more. 

Our AWS Lambda emulator cannot serve concurrent requests, so I had to tinker with a few things to make this work reliably for any test that touches the lambda:

* For permission tests that don't actually need to run view internals at all, just mock the call to the export step
* For tests that do exercise the lambda, ensure they run concurrently by putting them in a marked group and asking xdist to sort them into the same worker pool. This necessitated moving any export test out of doctest and into pytest so the mark could be applied.

One export test was entirely a duplicate so I removed it.


web/main/models.py
```diff
@@ -4968,13 +4954,10 @@ class LiveSettings(models.Model):
...
-        >>> ls, full_casebook, resource = [getfixture(f) for f in ['live_settings','full_casebook','resource']]
-        >>> prior_count = ls.export_average_rate
-        >>> _ = full_casebook.export(False)
-        >>> _ = resource.export(False)
-        >>> ls.refresh_from_db()
-        >>> assert ls.export_average_rate == prior_count + 2
+        Determine whether the current export has exceeded the allowable rate, and if not,
+        increment the counter tracking the number of exports over time.
         """
+        # FIXME consider doing this with Django F() expressions to avoid race conditions
```
rebeccacremona, line 5003: Verily: https://github.com/harvard-lil/h2o/issues/1538

rebeccacremona (approved):
Fantastic! Thanks so much for doing this, and untangling the pandoc-lambda stuff necessary to make it happen!

PR #1708, opened 2022-08-16
Remove optionality of user argument in permission-aware auth methods

lizadaly:
Flagged in review of #1318 ; this `user` parameter is never optional in practice because all these methods need to know what access level the user has.


rebeccacremona (approved):
🎉 Thanks!

PR #1715, opened 2022-08-18
Fixing Safari Bug in Create a Casebook

shljessie:
Fixed in Safari
Screen Shot 2022-08-18 at 10 22 04 AM


shljessie:
Works on Chrome too!

Screen Shot 2022-08-18 at 10 23 16 AM

matteocargnelutti (approved):
LGTM 👍

PR #1716, opened 2022-08-18
Limit instructional material to professors; fixes #1713

### LINKED ISSUE #1713: Further restrict instructional material permission check to verified professors
In  #1702 we decided to limit access to instructional material to verified professors only, not just authors, so implement it such that even users with edit access cannot see the material.

When implementing the front-end option for this, similarly only professors should be able to create professor-only content.

lizadaly:
Does what it says on the tin.


rebeccacremona (approved):
LGTM! I think you need to add a migration because of the tweaked help text.

lizadaly:
Oops good catch! I'll wait for https://github.com/harvard-lil/h2o/pull/1652 to land.

PR #1722, opened 2022-08-18
Collapse sections in casebook TOC more efficiently

### LINKED ISSUE #1721: "Collapse all" should collapse large casebooks faster
On a slower machine with very deeply nested casebooks, the "Collapse all" button takes too long to react. 

Either:

* Figure out a way to fire the events more quickly
* Close only the top-level nodes

lizadaly:
Two optimizations to speed up the expand/collapse toggle for large casebooks or slower machines:

* Only selects top-level nodes rather than all descendant nodes
* Performs the mutation in a single transaction batch rather than one-by-one

Tested with Chrome dev tools on CPU mode "6X slower"—it's faster for sure.

If users really want all nodes collapsed it's easy to switch out to the earlier behavior, but top-level nodes will always be more performant, so starting from there.

(Also corrects a dropped property, probably a merge victim.)

fixes #1721

https://user-images.githubusercontent.com/19571/185490372-a5eb6f57-a81e-4a6c-8a68-855b8c0861fb.mov

### DIFF
--- a/web/frontend/components/TheTableOfContents.vue
+++ b/web/frontend/components/TheTableOfContents.vue
@@ -55,7 +55,8 @@ export default {
     CollapseTriangle,
   },
   data: () => ({
-    needsDeleteConfirmation: {}
+    needsDeleteConfirmation: {},
+    tocCollapsed: false
   }),
   directives: {
     focus: {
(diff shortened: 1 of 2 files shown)


matteocargnelutti (approved):
LGTM 👍 !

Very nice addition. 

I think the sheer number of layout & paint operations involved in opening all the descendant nodes is sure to take its toll on older machines, regardless of how efficient the code to do so  is.

Issue #1723, opened 2022-08-19
Ensure exports respect the permissions of individual nodes


lizadaly:
Both the docx export and the printable html export display content nodes that are marked `is_instructional_material` even if the user does not pass the permission check.

PR #1727, opened 2022-08-22
Table of contents in printable HTML view; fixes #1710

### LINKED ISSUE #1710: Generate the table of contents in the printable HTML view
For the printable HTML view, generate a table of contents that is suitable both for printing and for navigating from a browser.

This should perhaps be a unique endpoint that can be linked from the next/previous navigation in the printable HTML screen view.

lizadaly:
Adds a table of contents that's nested at two levels to the printable HTML view. As with the book title, this displays only when requesting the first top-level node of the book.

* The top-level sections are hyperlinked with underlines in the screen view. The underlines aren't visible in the print view.
* Locally adds a font and license so the printable HTML view no longer uses the serif default of the browser.
* Breaks up the monolithic CSS file for book styles a bit.
* This also corrects part of #1723; resources that are marked `is_instructional_content` will not be displayed if the current user is not a verified professor.

image


sabzo (approved):
lgtm. +1

PR #1728, opened 2022-08-22
Some type corrections and mypy config

lizadaly:
Fix up some types (`mypy` now passes 100% for me) and add some config that makes it easier for local editors to discover the right files (a lot of default configurations don't work with the source code in a subdirectory like this).

I tried to upgrade `mypy` here per #1701 but `django-stubs` isn't yet compatible, so that should wait.


rebeccacremona (approved):
LGTM!

PR #1729, opened 2022-08-22
Ensure export respects permissions; fixes #1723

### LINKED ISSUE #1723: Ensure exports respect the permissions of individual nodes

lizadaly:
Adds the user to the export process to determine which content is visible to them.

*Exported as a verified professor*
image

*Exported as another user*
image

### DIFF
--- a/web/conftest.py
+++ b/web/conftest.py
@@ -818,9 +818,7 @@ def assert_num_queries(pytestconfig, monkeypatch):
                         stack_frames = (
                             q["stack"]
                             if pytestconfig.getoption("verbose") > 1
-                            else [q["userland_stack_frame"]]
-                            if q["userland_stack_frame"]
-                            else []
+                            else q.get("userland_stack_frame", [])
                         )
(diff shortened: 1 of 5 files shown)


web/conftest.py
```diff
@@ -818,9 +818,7 @@ def _assert_num_queries(db="default", **expected_counts):
                         stack_frames = (
                             q["stack"]
                             if pytestconfig.getoption("verbose") > 1
-                            else [q["userland_stack_frame"]]
-                            if q["userland_stack_frame"]
-                            else []
+                            else q.get("userland_stack_frame", [])
```
lizadaly: This fixes a bug with this query assertion; passing the `-v` flag is meant to show all queries that ran, but this line could never succeed if `userland_stack_frame` was missing from the dict.

bensteinberg (approved):
LGTM

PR #1730, opened 2022-08-22
Ensure that does_display_ordinals must track instructional content; fixes #1725

### LINKED ISSUE #1725: Enforce constraint that `is_instructional_content` nodes are unnumbered
Though this will be enforced at the UI level on the frontend, it should also be enforced on model save that if `is_instructional_content` is on, `does_display_ordinals` is off.

`does_display_ordinals` should also be available in the admin page for `ContentNodes`, to fix up any issues.

lizadaly:
Enforces that if `is_instructional_material` is true, the node can never be marked as appearing in the ordinal list. This is to ensure that students and non-professors never see skipped ordinal numbers in TOCs or book content.

This does _not_ always renumber the casebook following such a change. I adopted the convention already in the code that `content_tree__repair()` is not called automatically on model `save()`, presumably because it's expensive, or we may be mid-update? 

I think the only two places users will be able to toggle `is_instructional_material` will be on the frontend edit page for a resource (still TODO) and the Django admin. This PR causes renumbering to happen after such an edit in the admin; when I add the frontend toggle, I'll ensure that repair is called only when necessary.

There are other ways this could be implemented that would be more centralized—for example `ContentNode.save()` could fire off a `post_save()` Django signal and renumber the casebook. But we don't otherwise use signals in this repository and they can be hard to reason about. Open to other options!

### DIFF
--- a/web/main/admin.py
+++ b/web/main/admin.py
@@ -440,6 +440,7 @@ class ContentNodeAdmin(BaseAdmin, SimpleHistoryAdmin):
         "headnote",
         "created_at",
         "updated_at",
+        "does_display_ordinals",
         "is_instructional_material",
     ]
     raw_id_fields = ["casebook"]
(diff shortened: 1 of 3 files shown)


matteocargnelutti (approved):
LGTM 😄 . 

This implementation feels correct and easy to understand to me.

PR #1732, opened 2022-08-23
Use a different webpack file in dev

lizadaly:
It's annoying to have to deal with conflicts or upstream changes in `webpack-stats.json`, or to have to stash it every time you pull from upstream.

This change switches to using a different file in dev mode, which is added to `.gitignore`.

Works locally, though I'm not otherwise sure how to test other than to try it in CI/staging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ python-shell.sh
 *.iml
 .vscode
 .mypy_cache
+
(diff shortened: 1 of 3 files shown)


lizadaly:
Unfortunately this approach isn't going to pan out.

PR #1733, opened 2022-08-23
Search inside should check the permissions of the results; fixes #1724

### LINKED ISSUE #1724: "Search inside a casebook" needs to limit results to content nodes available to the user
Content that is tagged as `is_instructional_content` should be indexed (so it's findable when used by professors), but search results should be filtered to match the user's current permission level.

lizadaly:
Adds the `is_instructional_material` flag for textnodes to the full-text search index, and then filters on those if the user is not a verified professor. 

* Centralizes the permission check on the user for whether they can view instructional material. Currently it's just whether they're a verified professor. 
* To make the search API consistent, all content types expose `is_instructional_material` because that is part of the `ContentNode` model, but only `TextBlocks` will actually have the flag set, so we can save having to join on `ContentNode` for legal documents (it vastly increases the search indexing time), and links make no sense in this context. Those two resource types are always `is_instructional_material=False` in the search index.
* I moved the search tests out of doc tests and into their own test module since they were long and duplicative of each other. 

## Deployment note

This requires that `fab create_fts_index` be run at deployment time; until it completes and adds the new field, "search inside" will error. I can't really think of a way around this besides maybe some intermediate step that adds the new field to the index, unpopulated, without actually reindexing the content, but possibly that's more fragile than it's worth?

### DIFF
--- a/web/main/create_fts_index.sql
+++ b/web/main/create_fts_index.sql
@@ -1,7 +1,7 @@
 DROP MATERIALIZED VIEW IF EXISTS tmp_fts_search_view;
 DROP MATERIALIZED VIEW IF EXISTS tmp_fts_internal_search_view;
 CREATE MATERIALIZED VIEW tmp_fts_internal_search_view AS
-    -- seperate category for full-text search
+    -- separate category for full-text search of legal documents
     SELECT
            row_number() OVER (PARTITION BY true) AS id,
            c.id AS result_id,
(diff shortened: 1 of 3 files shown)


lizadaly:
Search results from a verified professor on some test content: 

image


Search results on the same document from a non-professor user:

image

lizadaly:
Something went wrong here with the upstream merge, closing for now.

PR #1735, opened 2022-08-24
Add missing call to create publish_check object; fixes #1734

### LINKED ISSUE #1734: Clicking on a TOC item in edit mode returns a blank page
An error message suggests this might be related to the `publish-check` flag in the Vue component.

lizadaly:
The error in the ticket is coming from a complex object being passed down to `action_buttons.html`; though this code is a bit fragile, I think this is the one view that uses that include that wasn't populating `publish_check` in the template context. This eliminates the error for me locally.

### DIFF
--- a/web/main/views.py
+++ b/web/main/views.py
@@ -2656,6 +2656,15 @@ def annotate_resource(request, casebook, resource):
             "editing": True,
             "edit_mode": True,
             "body_json": body_json,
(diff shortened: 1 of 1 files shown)


lizadaly:
(This is a pretty noticeable bug in prod so would be good to ship quickly today.)

bensteinberg (approved):

PR #1739, opened 2022-08-24
Search inside should check the permissions of the results; fixes #1724

### LINKED ISSUE #1724: "Search inside a casebook" needs to limit results to content nodes available to the user

lizadaly:
Adds the `is_instructional_material` flag for textnodes to the full-text search index, and then filters on those if the user is not a verified professor. 

* Centralizes the permission check on the user for whether they can view instructional material. Currently it's just whether they're a verified professor. 
* To make the search API consistent, all content types expose `is_instructional_material` because that is part of the `ContentNode` model, but only `TextBlocks` will actually have the flag set, so we can save having to join on `ContentNode` for legal documents (it vastly increases the search indexing time), and links make no sense in this context. Those two resource types are always `is_instructional_material=False` in the search index.
* I moved the search tests out of doc tests and into their own test module since they were long and duplicative of each other. 

User on the left is a professor; user on the right is not:

image

## Deployment note

This requires that `fab create_fts_index` be run at deployment time; until it completes and adds the new field, "search inside" will error. I can't really think of a way around this besides maybe some intermediate step that adds the new field to the index, unpopulated, without actually reindexing the content, but possibly that's more fragile than it's worth?

### DIFF
--- a/web/conftest.py
+++ b/web/conftest.py
@@ -587,6 +587,7 @@ def full_casebook_parts_with_prof_only_resource(full_casebook_parts_factory):
     parts: list[ContentNode] = full_casebook_parts_factory(user=prof)
     parts[3].is_instructional_material = True
     parts[3].title = "Instructional material"
+    parts[3].resource.name = "Instructional material"
     parts[3].resource.content = "This is instructional material"
     parts[3].resource.save()
     parts[3].save()
--- a/web/main/create_fts_index.sql
+++ b/web/main/create_fts_index.sql
@@ -1,10 +1,11 @@
 DROP MATERIALIZED VIEW IF EXISTS tmp_fts_search_view;
 DROP MATERIALIZED VIEW IF EXISTS tmp_fts_internal_search_view;
 CREATE MATERIALIZED VIEW tmp_fts_internal_search_view AS
-    -- seperate category for full-text search
+    -- separate category for full-text search of legal documents
     SELECT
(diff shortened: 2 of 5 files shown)


bensteinberg:
I can add `fab create_fts_index` to the deployment orchestration for this deployment alone, or in general.

lizadaly:
> I can add `fab create_fts_index` to the deployment orchestration for this deployment alone, or in general.

That's the one that takes a long time, so I think just the one-time is the right call here.

rebeccacremona (approved):
LGTM!

I love getting a chance to see this to-me-quite-intricate SQL go by; I feel like I'm learning a lot.

Thanks for extracting the tests out; this seems a lot more readable to me too.

rebeccacremona:
Should there be a migration file, since `models.py` was altered? I think yes, even if it a no-op in this case.

lizadaly:
> Should there be a migration file, since `models.py` was altered?

Thanks for checking! In this case no because:

```
    class Meta:
        managed = False
```

But confirmed: 

```
# python manage.py makemigrations
No changes detected
```

rebeccacremona:
> In this case no because:
> 
> ```
>     class Meta:
>         managed = False
> ```

Aha! Thanks!

PR #1740, opened 2022-08-24
Add front-end button for marking a resource as instructional material

lizadaly:
Adds a front-end button to mark instructional material.

The toggle effectively applies only to TextBlock resources, but is on the parent `ContentNode` model, so there's some extra checking here before showing the toggle. The toggle is only rendered to verified professors.

There's a very simple `onclick` handler to indicate to users that instructional material will not be numbered; this is enforced on the backend on save from a prior PR.

image

### DIFF
--- a/web/main/forms.py
+++ b/web/main/forms.py
@@ -209,9 +209,6 @@ class ResourceForm(ContentNodeForm):
         )
         self.helper.form_class = "edit_content_resource"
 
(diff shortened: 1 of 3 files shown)


rebeccacremona (approved):
🎉 

Sorry about the no redirect-after-POST; TBD at some point...

PR #1743, opened 2022-08-25
Allow staff users to see instructional material; fixes #1742

### LINKED ISSUE #1742: is_instructional_material check should be be true for staff users
Currently it's only verified professors, which is fine in dev/staging, but we don't want our staff accounts to be professors on production because our published casebooks would be visible in search, so this check should probably be expanded now to include staff.

lizadaly:
Verified manually.


lizadaly:
(This is not urgent to merge and can wait for other things.)

bensteinberg (approved):

PR #1745, opened 2022-08-29
Add workflow for manual image build

bensteinberg:
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build images
 
-on: [workflow_dispatch]
+on: workflow_dispatch
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,15 @@
+name: Build images
+
+on: [workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Push docker images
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,15 @@
+name: Build images
+
+on: [workflow_dispatch]
```
kilbergr:
```suggestion
on: workflow_dispatch
```

Unless you intend to add other actions, you can take this out of a list as demonstrated here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

kilbergr (approved):
This looks reasonable to me. I had one comment around syntax.

Also, reviewing the tests workflow, I noticed you had the following conditional:
`if: github.event_name == 'push' && github.ref == 'refs/heads/develop'`
for pushing the docker image. Do you want to include that here?

bensteinberg:
> Also, reviewing the tests workflow, I noticed you had the following conditional: `if: github.event_name == 'push' && github.ref == 'refs/heads/develop'` for pushing the docker image. Do you want to include that here?

I think not -- this workflow should never happen automatically, so `github.event_name` will never be `push`, and workflows triggered by `workflow_dispatch` must always be in the default branch, so I think `github.ref` will always point to develop.

kilbergr:
Ah! Thank you for the explanation!

PR #1751, opened 2022-09-12
Add lock icon to instructional material in TOC and content; fixes #1746

### LINKED ISSUE #1746: Identify restricted content in TOC
Teaching notes are great! I think it would be helpful to add some sort of signifier in the table of contents that a piece of content has been restricted to other verified professors. I'm thinking just a lock that would show up next to the title.

lizadaly:
If some content is instructor-only, display a lock icon on both the TOC and the resource page itself:

image

image

Additionally, this changes the resource page to not display the trailing ordinal of a section in the breadcrumb nav if that ordinal is not meant to be displayable.

**Previous**
image

**Now**
image


bensteinberg (approved):
LGTM!

PR #1752, opened 2022-09-13
Add a button for the printable HTML view for staff only

lizadaly:
For staff/admin users, the right-side pane will now look like this:

image

Clicking the button will open the printable HTML view in a new tab.

I dunno what we'll finally call this feature (PDF view? Print preview? HTML export?) but I went with this and this icon for now since it's only internal.


rebeccacremona (approved):
LGTM!

PR #1754, opened 2022-09-13
Add some styles from CAP css and adapt it for print layout

lizadaly:
Port over some of the SCSS from the website used for CAP to the plain CSS used in the printable HTML view (it's plain CSS because it doesn't go through the bundling process).

Generally this is an easier problem for the print CSS than for the Word layout because the styles are either just re-expressions of what's on the website, or expressions of the native markup in the CAP HTML. 

For example, in the Word doc, footnote text is the same size as the body text, but it's `<small>` in the HTML, and so the website and the printable view both display footnote-sized content:

**Word**

image

**Printable HTML**

image

Similarly the text-indent seen here is from repeated `&nbsp;`s in the source; we filter that out in the Word export but not on the website. I think I'll want to filter it and then just replace it with CSS indent, because the indent is expected in print layout but we'd want to apply it consistently across all content.

Finally, I made an editorial call here to start all cases on new pages (unlike the Word version).

### DIFF
--- a/web/static/as_printable_html/book.css
+++ b/web/static/as_printable_html/book.css
@@ -5,12 +5,13 @@
 
   @font-face {
     font-family: 'LibreCaslon';
-    src:  url('../fonts/LibreCaslonText-Regular.woff2') format('woff2');
+    src: url('../fonts/LibreCaslonText-Regular.woff2') format('woff2');
     font-weight: normal;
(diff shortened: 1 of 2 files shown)


kilbergr (approved):
These seem like reasonable choices to me.

PR #1760, opened 2022-09-15
Strip `nbsp;` from content using a template tag; improvements to running heads and other markup

lizadaly:
Kind of a grab bag here:

* Adds a Django template filter to do a simple string replacement process, so far just for stripping `&nbsp;`s from CAP where it's used for paragraph indenting. I might drop this in the future in favor of replacing the `annotated_content_for_export()` method (which is very pandoc-centered) right now.
* Simplifies the template logic which was originally copied from the pandoc export that varies by resource type, to vary less.
* Changes the behavior of the running head: left side is the book title, right side is the node title.
* Handles truncating the running title in Django—ideally this would be done with `text-overflow: ellipses` but that can't be applied to CSS generated content, which is how Pagedjs handles this. 

Other small CSS tweaks, e.g. moving the page numbers to the outer edge of the recto/verso pages rather than always appearing on the lower right in either case.

**Examples**

Running heads:

image

Embedded links are the only resource type that varies now:

image

### DIFF
--- a/web/main/templates/export/as_printable_html/node.html
+++ b/web/main/templates/export/as_printable_html/node.html
@@ -1,5 +1,4 @@
-{% load call_method %}
-
+{% load call_method string_strip %}
 
 
 <section class="{{ node.type }} {{ node.resource_type|default:'' | lower }} depth-{{ node.ordinals | length }}">
@@ -60,6 +59,7 @@
 
     {% elif node.is_resource %}  {# case or textblock #}
 
+      {# TODO understand what is going on here #}
       {# Duplicated in the link block because headnotes should go below a link but above cases and text blocks #}
       {% if node.headnote %}
         <section class="headnote">{{ node.headnote_for_export }}</section>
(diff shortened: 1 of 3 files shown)


matteocargnelutti (approved):
LGTM @lizadaly 😎 ! 

**Very minor consideration:**
For links bearing `target="_blank"`, I'd suggest letting users know that these links open in a new tab / window via their label.

lizadaly:
Thanks! Since this is meant to be a print preview, I don’t want to add any text to the body of the book that would not make sense in a printed context. Does it seem okay to you to omit the label in this case? (It’s an active link in this view just for convenience when proof-reading.)

matteocargnelutti:
Entirely up to you of course, but I'd personally consider removing `target="_blank"` altogether if adding a label for it is not an option :).

lizadaly:
OK! I'm going to leave it as-is since it's just following the behavior for resource links from the website itself, which I think is what users will expect: https://github.com/harvard-lil/h2o/blob/develop/web/main/templates/includes/bodies/link.html#L3

PR #1761, opened 2022-09-19
Update collection stats in search to report full list; some type fixes

lizadaly:
Addresses the confusing UI where the messaging at the top reflects the search the user did, which can make it seem like H2O has few documents.

**Before (prod)**

Note: we have more than two authors!

image

**After (dev)**

image

(The numbers don't match because my local index is behind production by a few weeks.)

While I was there, did some housekeeping on the search method:

* Added types to the parameters
* Removed the default dict and list params ([these aren't safe](https://towardsdatascience.com/python-pitfall-mutable-default-arguments-9385e8265422) though probably wasn't a problem here in practice)
* Added commas to the result numbers 
* Dropped `base_query` as a param; it seemed to have no callers.

fixes ##1759

### DIFF
--- a/web/main/models.py
+++ b/web/main/models.py
@@ -41,7 +41,7 @@ from django.core.exceptions import ValidationError
 from django.core.validators import validate_unicode_slug, MaxLengthValidator
 from django.db import models, connection, transaction, ProgrammingError
 from django.core.paginator import Paginator
-from django.db.models import Count, F, JSONField
+from django.db.models import Count, F, JSONField, QuerySet
 from django.template.defaultfilters import truncatechars
 from django.template.loader import render_to_string
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


web/main/models.py
```diff
@@ -942,14 +951,12 @@ def _search(
                 if order_by in ["created_at", "effective_date"]:
                     order_by_expression = ["-metadata__effective_date", display_name]
         results = results.order_by(*order_by_expression)
-        results = Paginator(results, page_size).get_page(page)
+        paged_results = Paginator(results, page_size).get_page(page)
```
lizadaly, line 954: This is that pattern where mypy got mad because we've reassigned a variable to an object of a different type.

rebeccacremona:
> Removed the default dict and list params

Ooh, good catch.

rebeccacremona (approved):

PR #1765, opened 2022-09-20
Don't offset the action buttons outside the bounding box of the casebook inner section

lizadaly:
Drops the negative top offset for the action buttons which pushes them into the top white area of the casebook page. Adds a positive offset of half the distance between the buttons (16px, halved).

**Before**

image

**After**
image

Does anyone know why this was this way? It [predates the most recent change](https://github.com/harvard-lil/h2o/commit/62176c6b91f8e1d7bd5783643f35a00e8242db52) to this layout to accommodate mobile but I didn't dig back further. This change has no effect on the mobile layout, which doesn't use this selector.

If folks recall this is intentional I can just close this.

### DIFF
--- a/web/frontend/styles/casebook.scss
+++ b/web/frontend/styles/casebook.scss
@@ -109,7 +109,7 @@ main > header.casebook {
     display: block;
     position: absolute;
     right: percentage((-1.5 / $grid-columns));
-    top: 0;
+    top: 8px;
     width: 12.5%;
     margin: 0;
   }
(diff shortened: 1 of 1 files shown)


rebeccacremona:
> Does anyone know why this way this way?

That's actually how the designer [sketched things up](https://drive.google.com/file/d/116lyqJlpeaczS-gZLK_YrfYESeAlFC3G/view?usp=sharing)

E.g.
image

🤷‍♀️ I personally have no horse in this race

lizadaly:
Thanks for finding that. It looks unintentional to me and is something the RAs also asked about, but it's easy enough to revive, so I'll close this for now and we can reopen if someone wants to make the shift.

Issue #1767, opened 2022-09-20
ContentAnnotation model should index on `global_start_offset`


lizadaly:
This field is in the `ordering` property of the Django model, so will be applied as default ordering, but there's no index on it, so this can be very slow when querying the whole table of 2,551,592 rows.

In the context of selecting annotations from within a particular resource this is no big deal, but for any query that selects from the table as a whole, it makes a huge difference if you let Django add the field to its `ORDER BY` statement:

```
>>> default_ordering = ContentAnnotation.objects.all()[1:2]

>>> str(default_ordering.query)
'SELECT "main_contentannotation"."id", "main_contentannotation"."created_at", "main_contentannotation"."updated_at", 
"main_contentannotation"."kind", "main_contentannotation"."content", "main_contentannotation"."global_start_offset",
"main_contentannotation"."global_end_offset", "main_contentannotation"."resource_id" FROM "main_contentannotation" ORDER
 BY "main_contentannotation"."global_start_offset" ASC, "main_contentannotation"."id" ASC LIMIT 1 OFFSET 1'

>>> default_ordering.explain()
'Limit  (cost=48498.40)
```

Compare:

```
>>> no_ordering = ContentAnnotation.objects.all().order_by()[1:2]

>>> str(no_ordering.query)
'SELECT "main_contentannotation"."id", "main_contentannotation"."created_at", "main_contentannotation"."updated_at", 
"main_contentannotation"."kind", "main_contentannotation"."content", "main_contentannotation"."global_start_offset", 
"main_contentannotation"."global_end_offset", "main_contentannotation"."resource_id" FROM "main_contentannotation" LIMIT 1 
OFFSET 1'

>>> no_ordering.explain()
'Limit  (cost=0.02)
```

Because of the table size, adding the index in a migration might take a bit of time to run on prod.

PR #1768, opened 2022-09-20
Add an indexed ordering field; add casebook display; add filtering by…

lizadaly:
I wanted to find more details on content annotations but the admin was slow (see #1767) and didn't have the filtering and field options necessary to actually track down a specific type of annotation to a specific casebook and resource.

* Explicitly orders on `-id`, an indexed field, which drops the query time for the list page from ~1000ms on my machine to 200ms.
* Adds a filter by "kind", meaning whether the annotation is an elision, note, highlight, etc.
* Adds the casebook title and resource name to the list display
* Adds the casebook title to the detail page

**Before**
image

(Takes about 12 seconds to render on prod; TBD how much faster this will be.)

**After**
image

### DIFF
--- a/web/main/admin.py
+++ b/web/main/admin.py
@@ -565,8 +565,6 @@ class AnnotationsAdmin(BaseAdmin, SimpleHistoryAdmin):
     def resource_type(self, obj) -> str:
         return obj.resource.resource_type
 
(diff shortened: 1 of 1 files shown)


matteocargnelutti (approved):
NICE 🚀

Issue #1764, opened 2022-09-20
Visually represent professor-level annotations in some way


lizadaly:
Ensure that professor-added content like notes or corrections are set off from the main text in an intelligible and attractive way.

Where multiple users have contributed, ensure it's obvious who wrote what.

lizadaly:
Representation in the db:

```
postgres=# select kind, count(*) as count from main_contentannotation group by kind order by count desc;
    kind    |  count  
------------+---------
 elide      | 2194541
 replace    |  157587
 highlight  |  129674
 note       |   43579
 correction |   21883
 link       |    4328
```

lizadaly:
Current behavior for the remaining annotation types in the Word export:

## Replacements
Replacements are presented in green, italicized text with dotted underlines.

<img width="760" alt="image" src="https://user-images.githubusercontent.com/19571/192837721-bc5e6e09-8308-4200-b62d-7faa4718de0c.png">

I'm not sure how that design was arrived at; it's not something I recognize.

## Links

See https://github.com/harvard-lil/h2o/issues/1783

## Corrections

These are transparently just replaced (there is no visual indication that the text differed). This seems correct to me.

PR #1769, opened 2022-09-20
Add highlight annotation styles; part of #1764

lizadaly:
Adds a style  for text which had a "highlight" annotation applied to it, as a light gray background behind the highlighted text.

The current serialization for the Word export gives us markup like:

```html

<p>
  <span class="highlighted">The highlighted text.</span>
</p>
```

There are two styles added here: one targeting the highlighted span itself, and an identical one targeting the not-quite-production [`:has()` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:has), which lets the developer target parent nodes. `:has()` is in Chrome, Edge, and Safari, but behind a flag in Firefox, so those users get the span fallback. The difference is whether the whole block is highlighted or not:

**Firefox**

image

**Chrome**
image

When this lands eventually as "export to PDF" the output will probably be driven by some fixed version of Chromium, so I think it's safe to assume any Chrome-supported CSS is in play for the "final" output, as long as the previews look reasonable in Firefox (and it'll support `:has()` soon enough anyway.)

For the block-level version, I used a background plus a variable-sized box-shadow in order to push the highlight outside the paragraph's bounding box; otherwise the whole thing looks like it's inset in a callout. 

I'm open to making this yellow instead and letting the printer deal with transforming it to greyscale; this gives us a little more control over the exact color at the expense of it looking less highlight-y.

Also cleaned up a little bit of the link styling and sorted the styles in a more logical order.

### DIFF
--- a/web/static/as_printable_html/book.css
+++ b/web/static/as_printable_html/book.css
@@ -19,16 +19,14 @@
     --casebook-font-size: 4mm;
     --casebook-line-height: 170%;
     --casebook-font-family: "LibreCaslon";
+    --link-icon-height: 10mm;
+    --link-icon-width: 10mm;
+    --link-icon-margin: 5mm;
+    --highlight-background-color: #ededed;
   }
-
 
   a {
     text-decoration: none;
-  }
-
-  .ordinal {
-    display: inline-block;
-    margin: 0 1em 0 0;
   }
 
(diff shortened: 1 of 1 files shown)


lizadaly:
Oops, I didn't consider the case where only a subset of the paragraph is highlighted; in this case the `:has` selector will erroneously highlight the whole block. Need to rethink this!

Issue #1770, opened 2022-09-20
Investigate using the Range API to rebuild annotations in printable HTML mode


lizadaly:
The Word/pandoc export does a [heroic amount of work](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L2741-L3108) in Python/lxml to construct a new HTML tree with injected nodes to reflect annotated content. This makes sense since the input to the pipeline is a document created by Python; there's no room for any other kinds of transformations.

Since printable HTML has access to a JavaScript runtime, it may be simpler to use the [Range API](https://javascript.info/selection-range) to construct the highlighted/elided/replaced blocks. This might reduce the likelihood of different results between the exported version and the website version, especially around edge cases.

Explore using the Range API to mark up annotated sections _before_ PagedJS starts transforming the HTML (which would invalidate the offsets): https://github.com/harvard-lil/h2o/blob/develop/web/frontend/pages/as_printable_html.js

lizadaly:
Seems to work!

PR #1772, opened 2022-09-22
Add highlighting support to printable HTML view

lizadaly:
As part of adding support for all annotation types to the printable HTML view, start with highlights. This follows the approach of #1770 which to my surprise seems to work.

Annotations created by the frontend are expressed as start and end "global" offsets—character offsets relative to the book content. One challenge that the current frontend and export code has to deal with is that any change to the markup needed to express elisions, highlights, or corrections is itself likely to alter those character offsets, because they modify the DOM tree. A nice feature of the [Range API](https://developer.mozilla.org/en-US/docs/Web/API/Range) is that it sits alongside the DOM, not within it. 

For each "chapter" and for each highlight in that chapter:

* Find the node containing text that lies within the boundary of the start offset for the highlight. Create a Range with a relative start offset that marks where the user began the highlight.
* If the highlight ends in this same text node, close the Range at the point at which the user ended the highlight. Otherwise, close the range when the node ends.
* If the highlight end hasn't been found yet, keep opening new Ranges as text nodes are consumed until the end is found.

Once all highlights have been turned into one or more ranges, for each range, use the native [range.surroundContent()](https://developer.mozilla.org/en-US/docs/Web/API/Range/surroundContents) method to wrap the text content in a styled span. 

I think all annotation kinds (elisions, corrections) should be safe to handle this way, and by deferring DOM mutations until the end, they shouldn't collide with each other because each Range created is relative to its nearest parent. This also lets us [avoid having to decorate nodes](https://github.com/harvard-lil/h2o/blob/4e0ad9597fb2860296da060195528e8cd79391e7/web/frontend/components/TheAnnotator.vue#L103-L107) just for the purposes of skipping them. However, it's possible this implementation will not be totally resilient to the other annotation types and I'll cross the bridge when I get there.

## Testing

At some point printable HTML will need tests, but H2O doesn't have Playwright yet and that's the logical way to test this. For now, some manual diffs between the website highlight views and the PagedJS view (I put it into single-page rather than two-up view to make it easier to compare.)

Screen Shot 2022-09-22 at 12 19 50 PM

### DIFF
--- a/web/main/templates/export/as_printable_html/node.html
+++ b/web/main/templates/export/as_printable_html/node.html
@@ -33,18 +33,22 @@
 
 
     {% if node.is_resource %}
-      <section class="resource">
+     {# Whitespace beteen the <section> element and the content must not be introduced to avoid throwing off the annotation offset values #}
+      <section class="resource" data-node-id="{{ node.id }}" data-resource-id="{{ node.resource_id }}">{{ node.resource.content|safe }}</section>
 
-        {% if node.resource.content %}
-          {% call_method node 'annotated_content_for_export' export_options=export_options as contents %}
-        {% endif %}
-
-        {# Strip out any manual whitespace, and mark the output as safe because this operation should not result in broken HTML #}
-        {{ contents|string_strip:"&nbsp;"|safe }}
-
-      </section>
+      {% for annotation in node.highlights %}
+      <li data-annotation-type="highlight"
+          data-start-offset="{{ annotation.global_start_offset }}"
+          data-end-offset="{{ annotation.global_end_offset }}"
+          class="hidden">{{ annotation.resource_id }}</li>
+      {% endfor %}
     {% endif %}
   </div>
-  </section>
 
-<span class="truncated-title">{{ node.ordinal_string }} {{ node.title|truncatechars:50 }}</span>
+
+</section>
+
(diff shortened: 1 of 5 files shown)


web/static/as_printable_html/book.css
```diff
@@ -152,16 +174,16 @@
...
   }
-
-  sup {
-    font-size: calc(var(--casebook-font-size) / 2.5);
-
+  /* Support for greyscale-only printing */
+  @media not color {
+    .highlighted {
+      background: var(--highlight-background-greyscale);
+    }
```
lizadaly, line 185: I can't find any indication of whether this is supported anywhere (Chrome does not respect it when I print-preview on a b/w printer—I get its own greyscale rendition of the highlight color). But the spec says this should work as intended.

web/main/templates/export/as_printable_html/node.html
```diff
@@ -33,18 +33,22 @@ <h2 class="subtitle">{{ node.subtitle }}</h2>
 
 
     {% if node.is_resource %}
-      <section class="resource">
+     {# Whitespace beteen the <section> element and the content must not be introduced to avoid throwing off the annotation offset values #}
```
lizadaly:
```suggestion
     {# Whitespace between the <section> element and the content must not be introduced to avoid throwing off the annotation offset values #}
```

bensteinberg (approved):

rebeccacremona:
This is way cool.

Issue #1511, opened 2022-02-10
[new export styling] Wishlist: collapse consecutive identical elements without actual content


ChefAndy:
Nobody needs elision marks like they're shown in the screenshot. This also goes for empty paragraphs, paragraphs with just white space, etc.

<img width="706" alt="4" src="https://user-images.githubusercontent.com/1882536/153507366-0c66b26d-3139-49fc-bcfd-add6cc4bf0d4.png">

PR #1773, opened 2022-09-22
Handle elisions in printable HTML as part of #1764

lizadaly:
Extending from #1772 , this generalizes the previous code to consider all annotation types, and adds a serialization mechanism for elisions.

* Uses `<ins>` / `<del>` to represent these semantically.
* Hides adjacent empty elisions (the [current export doesn't do this](https://github.com/harvard-lil/h2o/issues/1511))

## Testing

image

Compare with Word export:

image

## Accessibility and æsthetics

(I don't think this is blocking for this PR, but I'm interested in thoughts!)

I was originally going to set the insertion ellipses as `ins::after { content: "…"}` but `<ins>` has special behavior with regards to [screenreaders](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ins#accessibility_concerns). Even though screenreaders aren't the target of a print preview page, I figure it's useful to think about using this markup in the website itself, and I saw some old discussion in Slack about using these elements.

I'm not happy with a screenreader reading `[…]` out loud (though at least it's a real ellipses character and not a string of dots). A few legal style guides and a free non-H2O casebook I checked seemed to favor excluding the brackets entirely, which would be my preference in the printed layout. On the other hand, screen readers [may skip over reading this content entirely](https://stackoverflow.com/questions/453189/what-do-screen-readers-do-with-an-ellipsis-unicode-character), which suggests maybe we should have a non-visible `ins::after { content: "elided" }` or similar, even though I know it's usually contraindicated to provide alternate content.

### DIFF
--- a/web/frontend/pages/as_printable_html.js
+++ b/web/frontend/pages/as_printable_html.js
@@ -6,18 +6,18 @@ document.addEventListener("DOMContentLoaded", () => {
     const css = tmpl.getAttribute("data-stylesheet");
     const paged = new Previewer();
 
-    // Collect groups of ranges for each highlight offset start and end. Note that start/end offset may
+    // Collect groups of ranges for each annotation offset start and end. Note that start/end offset may
     // cross multiple block-level node boundaries; this will construct one node range for each block-level element.
-    const highlightRanges = Array.from(
-      tmpl.content.querySelectorAll('[data-annotation-type="highlight"]')
+    const annotationRanges = Array.from(
+      tmpl.content.querySelectorAll("[data-annotation-type]")
     ).map((el) => {
       let ranges = [],
         currentOffset = 0,
-        textNode, // The current textNode we're processing
-        startNode, // The text node we've identified as the start of the highlight
-        endNode; // The text node we've identified that contains the end of the highlight
+        textNode, // The current text node we're processing
+        startNode, // The text node we've identified as the start of the annotation
+        endNode; // The text node we've identified that contains the end of the annotation
 
-      // Get the content from the DOM that matches this highlight, and its offset values
+      // Get the content from the DOM that matches this annotation, and its offset values
       const resource = tmpl.content.querySelector(
         `[data-node-id="${el.textContent}"]`
       );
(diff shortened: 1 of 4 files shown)


lizadaly:
(I see that the website uses `aria-label="elided text"` on the elision marker, which is a `span[role=button]`)

rebeccacremona:
re: accessibility for people using screenreaders

These thoughts are peripheral to this PR, but I would truly love to chat with one of our students and brainstorm together about what would actually be a good experience on the website. Like... I have to believe that the toggle buttons are maximally disruptive, probably to the point of making the case text unintelligible. As read, they call WAY too much attention to each elision, whereas the visual effect is (or attempts to be) subtle and easy to gloss over.

Similarly, my guess is that folks would most prefer to be able to toggle audible notifications on and off (either using the built-in handling of `<ins>` in their verbosity settings, or, if that's really only adjusted by power users, which I believe might be the case, via a control in the app), depending on whether they were primarily trying to listen and make sense of the content, or whether they were in that moment concerned with whether the book's author had tweaked the text. Screenreaders still aren't any good at conveying multiple dimensions of information at once, or any nuance of context; I feel like this is -exactly- the sort of situation where you'd love to be able to mark things up to be read _sotto voce_. "**WE THE COURT DECLARE** (pst there is elided text here don't mind me) **THAT ETC. ETC. ETC.**" 🤷‍♀️ But alas.

rebeccacremona:
FWIW my instinct is not to worry too much about screenreaders skipping ellipses entirely. I feel like students who use screenreaders are likely to be aware of that and to already have a strategy if they want them read, just because ellipses and "..." are used so frequently in daily textual conversation (if maybe less so in formal writing).

...but that would be another thing I'd want to confirm with actual law students. I make the assumption that law students are in general more expert screenreader users than the general population of people who use screen readers, but that might not be the case.

rebeccacremona (approved):
Love the collapsing of adjacent elisions!

lizadaly:
Thank you for the thoughtful comments! Ultimately I'd like to try to use this markup and CSS as the default "reading" view on the site (without the PagedJS overlay and pagination, which would just be for "print preview")—maybe when that gets closer it would be a good time to look at optimizing for reading via assistive technologies without needing to accommodate the affordances that support editing.

PR #1774, opened 2022-09-23
Add support for notes; add bold font; refactor annotations collection code

lizadaly:
Mostly this adds support for professor notes (not professor-only content, but notes created with the annotation tool). 

Also adds the bold version of the book font, which I inexplicably omitted last time, and moves the annotation code outside the main body of the `onload` handler for the page, as it was getting quite long. Also dropped the brackets around elision markers to see how that looks.

A note is rendered in the printable HTML output as a regular highlight around the author's selection, plus a new adjacent `<aside>` element containing the note content. 

In this print-centric version, there's no specific styling of the highlighted portion, and the author note is rendered as a footnote at the bottom of the current page.

In a screen-optimized layout we might want to do something with the highlighted text and to put the author text in the margin (more like the current screen version).  

## Sample

image


lizadaly:
Pulling this back to Draft as I realize the existing export renders these as footnotes, which makes more sense in a print context.

lizadaly:
Updated with footnote rendering.

kilbergr (approved):
This looks reasonable to me.

PR #1775, opened 2022-09-26
Add institution model and relationship to user; fixes #1684

### LINKED ISSUE #1684: Add a formal User field of "institution"
We'd like to keep the `affiliation` field open-ended (though maybe add auto-suggest as recommended in #1651 to limit sprawl) and add a formal `institution` field that we'll control.

A few questions to sort out:

* Can a user belong to more than one institution? This has big implications for the implementation—once we decide they can only belong to a single institution it's hard to change, but allowing multiple introduces more complexity. However, if the reality is that many professors teach in multiple institutions over some unit period of time, it's probably worth implementing that way.
* Do we need to track institutional membership over time? If so, that also introduces more complexity, but needs to be captured up-front. If it's important to ask "how many professors published casebooks from HLS in 2022" we'll need to capture this longitudinally rather than just relying on the current value of the field.

*Definition of done:*

* A `User` can belong to zero or one (or more) institutions (see above)
* `institution` is added to the admin list and change views and to the CSV export for professor reports
* If present, the `institution` field appears on the user's public dashboard page.
* If present, it appears in the search results list somewhere (next to the author name in parentheses?)

*How we'll populate it*

* Propose we generate a list of law schools and any other orgs we want to feature in the list.
* The list of allowable institutions is populated in the Django admin (ideally the admin will allow pasting in a list of newline-delimited items rather than forcing us to create a zillion items one at a time)
* Users can update their institution from the populated dropdown and, if none applies, they're directed to fill out `affilitation`.

We may or may not pre-populate institutions for known users.

Would it be madness to try to capture logos/seals for some or all institutions? We could display them on the user's profile page. (We don't need to decide this up-front.)

lizadaly:
Adds a new model, `Institution`, that describes an educational institution to which a verified professor might belong. This is intended to be a controlled vocabulary, unlike `User.affiliation`, which is free-text. See #1684 

* Creates the model
* Runs a data migration to prepopulate some known-good values that map to existing user institutions, including those derived from `EmailWhitelist`.
* Adds the admin for this model

The model includes `email_domains`, which is still unused, but may become a replacement for the `EmailWhitelist` model. I implemented it as an `ArrayField` because there are multiple email domains for existing universities in the `EmailWhiltelist` model. These get populated by the data migration.

Adds a slug field assuming we might want to populate a browsable list of professors who belong to specific institutions (especially Harvard).

Once run, the admin will be populated with values (I added myself to one university):

## Admin list page
image

## Admin change page (example with 2 email domains)
image

No tests, because everything here is pretty much Django built-ins, but I'm open to adding some if you have suggestions!

Nothing appears on the front-end yet.


web/main/admin.py
```diff
@@ -107,7 +107,7 @@ def log_deletion(self, request, object, object_repr):
 
     actions = None  # use ['delete_selected'] to allow delete action
     formfield_overrides = {
-        fields.JSONField: {"widget": JSONEditorWidget},
+        JSONField: {"widget": JSONEditorWidget},
```
lizadaly, line 110: This has no effect on this PR, but I noticed this was now incorrect for Django 3.2 when I first implemented one of the fields as JSON.

matteocargnelutti (approved):
LGTM 🚀 !

Maybe pattern validation could be a nice to have for `email_domains`?

lizadaly:
> LGTM 🚀 !
> 
> Maybe pattern validation could be a nice to have for `email_domains`?

Good idea! I’ll do that when this is wired up to the email verification step.

PR #1779, opened 2022-09-27
Drop docx experimental settings; fixes #1669

### LINKED ISSUE #1669: Remove remaining references to FORCE_DOCX_* settings and related code?

lizadaly:
For awhile now, the `FORCE_DOCX_*` settings for Word export have been `True` in all environments; this formalizes that and drops any code that checked whether the setting is enabled.

If the setting was false, the export used a different set of templates to render the response; these are now deleted.

Mostly relying on tests here (a few times they did fail as I was working through this so I know they touch this pipeline). Also tested manually on a whole casebook, and tested individual content node exports with either annotations enabled or disabled (this is still an option users can choose).

Please do check the lines that have some logic in them to make sure they still resolve logically to the same place as if the setting were enabled.


bensteinberg (approved):
This looks right to me.

bensteinberg:
(`black` has one complaint.)

Issue #1421, opened 2021-10-13
Explain correction UI


holtchesley:
Removing the `v-if="false"` from https://github.com/harvard-lil/h2o/blob/6ed792207e364f22ecab9f4af00281f07de0cc29/web/frontend/components/TheAnnotator.vue#L26 will enable typo correction, but we should also add docs and ui to explain what it does. 

Currently it will allow for a section to be replaced with alternate text on the web and in exports without additional styling/expansion options. We'd like to be able to use it for correcting CAP cases in the future too, or automatically applying corrections other authors have made.

PR #1784, opened 2022-09-28
Add correction and replacement annotation types

lizadaly:
Addresses two more types from https://github.com/harvard-lil/h2o/issues/1764

The Word export treats "replacements" and "corrections" separately. The former is intended to express some opinionated change on the part of the author modifying a case or other imported document. The latter is intended to express a typo correction. The front-end doesn't [strongly push this distinction](https://github.com/harvard-lil/h2o/issues/1421). Since this distinction is largely opaque to both authors and readers, I chose to treat them the same way here and hide the original source. However, it's in the injected markup, so treating it differently is just a matter of adjusting the CSS.

## Source text

image

## Printable view

image

Markup:

```html
<p>This line has 
   <del class="replace" datetime="2022-09-28T16:14:09.617686">replacement</del>
   <ins class="replace" datetime="2022-09-28T16:14:09.617686">even better text than was here before</ins>.
</p>
```

### DIFF
--- a/web/frontend/pages/as_printable_html.js
+++ b/web/frontend/pages/as_printable_html.js
@@ -114,6 +114,7 @@ document.addEventListener("DOMContentLoaded", () => {
             elision.setAttribute("datetime", datetime);
 
             const marker = document.createElement("ins");
+            marker.setAttribute("datetime", datetime);
             marker.classList.add("elision-marker");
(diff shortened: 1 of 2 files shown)


matteocargnelutti (approved):
🙂 Makes a lot of sense to me!

PR #1786, opened 2022-09-29
Handle different link types in the printable view; fixes #1783

### LINKED ISSUE #1783: Handle links in printable HTML view
We have 5 possible ways to make links:

1.  The link resource type, which appears as a standalone item in the table of contents
2.  The link annotation type, which is added using the annotation tool
3.  Adding a link via the WYSIWYG editor tool
4.  Adding a link by pasting existing content with links into the editor
5.  Importing content that has existing links (unsure if this happens in practice with our imported content, but it's certainly possible)

It's tricky to get a count of the number of uses of 3-5, but there are 275,023 link resource types and only 4,000 uses of #2. 

Examples from the Word export:

image

The URL then appears as a footnote:

image

Link resource:

image

It's probably not ideal that these all behave differently, though the link resource type does have special weight and may have associated metadata. The remaining cases should probably all be treated the same.

I think the main decision points are:

* Should the URL be inline with the linked content, or footnoted like the current export?
* Should the link target (the underlined bit) be visually represented in the text?

Should look into how other publishing guidelines handle links in print, especially in a book context.

lizadaly:
Follow the link format we arrived at in #1783.

## Source data

image

## Word export 
image

## Printable export 

* All link types look the same, except internal links which aren't rendered.
* The markup here is anchor tags in both the body text and the footnote, if the previewing user wants to click on them. 
* The footnote date is either derived from the annotation date (if created that way) or from the creation date of the document (which may or may not uniquely correspond with the link itself, but it's a decent proxy).

image

### DIFF
--- a/web/frontend/pages/as_printable_html.js
+++ b/web/frontend/pages/as_printable_html.js
@@ -109,8 +109,8 @@ document.addEventListener("DOMContentLoaded", () => {
     const css = tmpl.getAttribute("data-stylesheet");
     const paged = new Previewer();
 
-    // For any existing URLs, turn them into footnotes
-    tmpl.content.querySelectorAll('a[href^="http"]').forEach((el) => {
+    // For any existing URLs that aren't resources, turn them into footnotes
+    tmpl.content.querySelectorAll('a[href^="http"]:not([data-type])').forEach((el) => {
       hyperlinkFootnote(el, el['href'],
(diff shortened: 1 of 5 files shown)


lizadaly:
Flipping this back to draft as I want to do some more work on the Link Resource type as part of this PR.

lizadaly:
This is ready for review!

kilbergr (approved):
Thank you for walking through!

PR #1789, opened 2022-10-03
Iterate through nodes more carefully when discovering offsets

### LINKED ISSUE #1771: Attempting to highlight across an inline image or elision throws a JS exception
I wouldn't have thought to try this except that the code seems to want to handle the case where a highlight spans an image boundary:

* In the editor, add an inline image
* In annotations mode, try to drag a highlight that crosses over into the image

An exception will be thrown in the JS console and Vue will stop processing until the page is refreshed.

I would expect at least that no exception would be thrown, even if the resulting highlight is somewhat malformed.

```
vue.esm.js?a026:628 [Vue warn]: Error in render: "TypeError: Cannot read properties of null (reading 'textContent')"

found in

---> <TheAnnotator> at frontend/components/TheAnnotator.vue
       <TheResource> at frontend/components/TheResource.vue
         <Root>
warn @ vue.esm.js?a026:628
logError @ vue.esm.js?a026:1902
globalHandleError @ vue.esm.js?a026:1897
handleError @ vue.esm.js?a026:1857
Vue._render @ vue.esm.js?a026:3580
updateComponent @ vue.esm.js?a026:4090
get @ vue.esm.js?a026:4504
run @ vue.esm.js?a026:4579
flushSchedulerQueue @ vue.esm.js?a026:4335
eval @ vue.esm.js?a026:1998
flushCallbacks @ vue.esm.js?a026:1924
Promise.then (async)
timerFunc @ vue.esm.js?a026:1951
nextTick @ vue.esm.js?a026:2008
queueWatcher @ vue.esm.js?a026:4427
update @ vue.esm.js?a026:4569
notify @ vue.esm.js?a026:739
reactiveSetter @ vue.esm.js?a026:1064
proxySetter @ vue.esm.js?a026:4653
selectionchange @ TheAnnotator.vue?ef4c:133
selectionchangeHandler @ TheResource.vue?fb00:51
eval @ selectionchange.js?4e4a:16
vue.esm.js?a026:1906 TypeError: Cannot read properties of null (reading 'textContent')
    at getLength (html_helpers.js?9526:51:1)
    at getOffsetWithinParent (html_helpers.js?9526:74:1)
    at eval (TheAnnotator.vue?ef4c:98:1)
    at Array.map (<anonymous>)
    at VueComponent.offsets (TheAnnotator.vue?ef4c:98:1)
    at Watcher.get (vue.esm.js?a026:4504:1)
    at Watcher.evaluate (vue.esm.js?a026:4606:1)
    at VueComponent.computedGetter [as offsets] (vue.esm.js?a026:4860:1)
    at Object.get (vue.esm.js?a026:2113:1)
    at Proxy.render (TheAnnotator.vue?4360:11:1)
    at Vue._render (vue.esm.js?a026:3578:1)
    at VueComponent.updateComponent (vue.esm.js?a026:4090:1)
    at Watcher.get (vue.esm.js?a026:4504:1)
    at Watcher.run (vue.esm.js?a026:4579:1)
    at flushSchedulerQueue (vue.esm.js?a026:4335:1)
    at Array.eval (vue.esm.js?a026:1998:1)
```

There are similar issues in Sentry but I think they are slightly different cases of trying to highlight across non-text nodes.

lizadaly:
fixes #1771

Addresses a bug mentioned by users and Sentry where highlighting exuberantly across a document can cause the annotation code to throw an uncaught exception and stop processing. The only recourse for users in this scenario is to refresh the page.

This makes a few changes to both make the code easier to follow, and also to fix the bug:

## The bug

The root cause of the actual bug is in a loop that is designed to walk through the HTML tree and count the character offset to the start of the selection range. The tree-walking loop was written such that the iterator could reach the end of its iteration and return `null`, but code in the loop would still attempt to retrieve information about the node. In the JS console this resulted in an uncaught exception trying to get the length of the (null) node's text content.

This was addressed by letting the iteration exit naturally when it completes, or sooner if the loop knows it has exited the relevant part of the DOM. Iterating through all node types, not just text nodes, allows it to capture selections that span image boundaries. 

image

This fixes two cases—highlighting across an image boundary and (more commonly) highlighting across an elision boundary. I think it should generalize to any similar cases.

## Testing

I want to do a little more manual testing, but the new automated test case does fail in `develop` and passes here:

```
  1) getOffsetWithinELement
       returns the correct offset when the selection moves through an image:
     TypeError: Cannot read properties of null (reading 'textContent')
      at getLength (static/dist/js/webpack:/frontend/libs/html_helpers.js:51:1)
      at getOffsetWithinParent (static/dist/js/webpack:/frontend/libs/html_helpers.js:74:1)
      at Context.<anonymous> (static/dist/js/webpack:/frontend/test/libs/html_helpers.test.js:108:1)
      at processImmediate (node:internal/timers:466:21)
```

### DIFF
--- a/web/frontend/components/TheAnnotator.vue
+++ b/web/frontend/components/TheAnnotator.vue
@@ -77,7 +77,7 @@ export default {
       const wrapperRect = this.$parent.$el.getBoundingClientRect();
       const viewportTop = window.scrollY - (wrapperRect.top + window.scrollY);
       const targetRect = this.ranges[1].getBoundingClientRect();
-      
+
       return Math.min(Math.max(targetRect.top - wrapperRect.top, viewportTop + 20),
                       targetRect.bottom - wrapperRect.top).toString(10) + "px";
     },
@@ -92,17 +92,18 @@ export default {
     },
     offsets() {
       if(!this.hasValidRanges) return;
-      
+
       const el = document.querySelector(".case-text");
-      
-      const ret = ["start", "end"].map((s, i) =>
-        this.ranges[i][`${s}Offset`] + getOffsetWithinParent(el, this.ranges[i][`${s}Container`], this.contributesToOffsets)
+
+      const ret = ["start", "end"].map((s, i) => {
+        return this.ranges[i][`${s}Offset`] + getOffsetWithinParent(el, this.ranges[i][`${s}Container`], this.contributesToOffsets)
+      }
       );
       const sel_length = this.selection.toString().length;
 
       let sp = (this.ranges[0].startContainer == this.ranges[0].commonAncestorContainer);
       let ep = (this.ranges[1].endContainer == this.ranges[1].commonAncestorContainer);
-      
+
       if (sp ? !ep : ep) {
         // If the edge of a selection lies on an image, the offsets are calculated incorrectly
         if (sp) {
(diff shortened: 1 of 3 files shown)


matteocargnelutti (approved):
Awesome 👏

PR #1790, opened 2022-10-04
Change layout of printable HTML TOC from grid to simple floats; fixes #1788

### LINKED ISSUE #1788: Printable TOC needs to be implemented without grid
I implemented the layout of the printable TOC with a two-column CSS grid, but this has a side effect when a page break occurs in the middle of a grid row:

Start of one TOC entry, at the bottom of a page:

image

Continuation of this entry on the next page:

image

This is because on the next page, only the second column is inside the grid row now, so CSS forces it into the space that is meant to be occupied by the section title.

Probably a more resilient layout would just be sibling `<p>` elements with fixed blocks for the ordinals, but consult the PagedJS docs if there's more to do to hint to the layout engine what we want to happen here.

lizadaly:
Changes the layout engine of the table of contents in the printable view to use a simple left-float for the ordinal + chapter/section title. 

The previous layout with CSS Grid was subject to an artifact in the PagedJS pagination whereby a long grid row could break across pages, resulting in a new block container (after reflow by PagedJS), in which case CSS would shove the longer section title into the small first column intended for the section number. This changes the layout to a less sophisticated float + span which should behave less weirdly with breaks.

This is a little tough to A/B because any change to the markup like this will reflow everything differently, but on spot-checking about 5 different books, things seem to break as I'd expect.

Here's how it looks:

image

(as before, the underlines are only visible in the screen view as this is the primary way for previewing users to skip ahead in the book)

### DIFF
--- a/web/main/templates/export/as_printable_html/casebook.html
+++ b/web/main/templates/export/as_printable_html/casebook.html
@@ -58,9 +58,11 @@
         </ul>
       </div>
       <section class="casebook headnote">{{ casebook.headnote|safe}}</section>
+      {% if page.number == 1%}
       <div id="toc">
         {% include "export/as_printable_html/toc.html" with toc=toc %}
       </div>
+      {% endif %}
     </div>
 
     {% for child in children %}
--- a/web/main/templates/export/as_printable_html/node.html
+++ b/web/main/templates/export/as_printable_html/node.html
@@ -22,13 +22,13 @@
         <section class="headnote">{{ headnote }}</section>
       {% endif %}
 
-      <a href="{{ node.resource.url }}" target="_blank" data-type="resource">{{ node.resource.url }}</a>
+      <a href="{{ node.resource.url }}" target="_blank" data-type="resource link">{{ node.resource.url }}</a>
 
     {% else %}
 
         <h1 class="{{ node.type }} title">
           <span class="{{ node.type }} ordinal" data-toc-idx="{{ index }}">{{ node.ordinal_string }}</span>
-          {{ node.title }}
+          <span>{{ node.title }}</span>
         </h1>
 
         {% if node.subtitle %}
--- a/web/static/as_printable_html/book.css
+++ b/web/static/as_printable_html/book.css
@@ -153,20 +153,23 @@
   .node-container > p, section > p {
     text-indent: 2em;
   }
-  .ordinal:not(:empty) {
-    display: inline-block;
-    margin: 0 1em 0 0;
+  .title > .ordinal:not(:empty) {
+    float: left;
+    width: 10mm;
   }
-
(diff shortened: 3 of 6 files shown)


web/main/templates/export/as_printable_html/casebook.html
```diff
@@ -58,9 +58,11 @@ <h2 class="casebook subtitle">{{ casebook.subtitle|default_if_none:""}}</h2>
         </ul>
       </div>
       <section class="casebook headnote">{{ casebook.headnote|safe}}</section>
+      {% if page.number == 1%}
```
lizadaly, line 61: Only displays the TOC markup on the first section; previously this was just hidden by CSS but I wanted to add a page break before it, and I couldn't get PagedJS to ignore it even if it was hidden.

web/static/as_printable_html/book.css
```diff
@@ -153,20 +153,23 @@
...
   }
-
   .link-container {
     min-height: var(--link-icon-height);
     width: 100%;
   }
-  .link-container p {
-    margin-left: calc(var(--link-icon-width) + var(--link-icon-margin));
+
+  a[data-type~="link"] {
```
lizadaly, line 169: Unrelated correction to the link markup/styling that was missed in the previous link PR.

web/main/templates/export/as_printable_html/node.html
```diff
@@ -22,13 +22,13 @@ <h2 class="subtitle">{{ node.subtitle|default_if_none:"" }}</h2>
...
 
-      <a href="{{ node.resource.url }}" target="_blank" data-type="resource">{{ node.resource.url }}</a>
+      <a href="{{ node.resource.url }}" target="_blank" data-type="resource link">{{ node.resource.url }}</a>
 
     {% else %}
 
         <h1 class="{{ node.type }} title">
           <span class="{{ node.type }} ordinal" data-toc-idx="{{ index }}">{{ node.ordinal_string }}</span>
-          {{ node.title }}
+          <span>{{ node.title }}</span>
```
kilbergr, line 31: Not blocking at all but for my curiosity: what's the value of adding a span here?

kilbergr (approved):
Thank you for the PR info :) Made it very easy to parse!

web/main/templates/export/as_printable_html/node.html
lizadaly: So that it's addressable by this rule:

https://github.com/harvard-lil/h2o/pull/1790/files#diff-ecb321e58f1547e83335bbc1669dfcf87b70a66e423d70a2ac09b2dca4ec888aR36

Otherwise it would inherit the default span behavior of being an inline. Off the top of my head I don't remember how it flowed differently as an inline though!

PR #1792, opened 2022-10-05
[DRAFT] Upgrade django-webpack-loader and dependencies to latest

lizadaly:
Our versions of django-webpack-loader and its related JS library were quite old and no longer matched the site documentation. 

Draft pending more testing, but this changeset worked locally both in dev mode and with built assets (which I've committed here, as otherwise the test suite probably couldn't run.) I also would like to better understand some of the behavior differences.

The new bundle structure did not include one called `chunk-common` so I removed the reference to that from `vue.config.js`. I want to better understand what's in that in prod and compare it with what is ending up in `chunk-vendors.js` in the new version.


lizadaly:
I'm going to close this for now as I think things aren't in an optimal state to move forward here.

I believe this upgrade (and further upgrades to vue-cli) can work, but as configured this is fighting the tools a bit:

* The manually-specified page bundles aren't really reflective of how the code is structured anymore. It would probably be better to let tools handle code-splitting for us.
* Newer versions of vue-cli want to include or exclude polyfills based on a current list of supported browsers. We import `babel/polyfill` directly in the code, which squashes this behavior: 

```
  When setting `useBuiltIns: 'usage'`, polyfills are automatically imported when needed.
  Please remove the direct import of `@babel/polyfill` or use `useBuiltIns: 'entry'` instead.
```

* Part of the upgrade meant not needing to include our own [custom relative path bundler tracker](https://github.com/owais/webpack-bundle-tracker/issues/25), but out of the box, this upgrade generated almost 20MB of polyfills (see previous bullet). When I reimplemented the custom relative path tracker, the huge polyfill dropped out. I kind of suspect we're both including too many polyfills, and then accidentally omitting them, which might leave some marginal browsers out of the loop. 

I think I can still achieve my initial goals of fine-tuning the asset loading in the HTML using the current version of the tooling, so moving forward that way instead.

PR #1794, opened 2022-10-07
Add playwright and live server tests; fixes #1291

### LINKED ISSUE #1291: Set up headless browser tests
Maybe use Puppeteer?

lizadaly:
In preparation for using playwright for PDF generation and because we would like browser-based testing anyway, adds the playwright configuration borrowed from Perma and a simple test.


web/conftest.py
```diff
@@ -1,44 +1,46 @@
...
 from main.utils import re_split_offsets
 
-from test.test_helpers import dump_casebook_outline
-
-
 # This file defines test fixtures available to all tests.
 # To see available fixtures run pytest --fixtures
 
+# Allow setup of live server test cases; see https://github.com/microsoft/playwright-python/issues/439
+os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "true")
```
lizadaly, line 42: This is only required in Django 3.x; Perma is still on 2.2.

lizadaly:
I'm not sure it's the same Chromium that we add that is also added by Playwright?

```
#12 [ 8/10] RUN apt-get update && apt-get install -y chromium=106.0.5249.91-1~deb11u1     chromium-common=106.0.5249.91-1~deb11u1     chromium-driver=106.0.5249.91-1~deb11u1     chromium-l10n=106.0.5249.91-1~deb11u1     chromium-sandbox=106.0.5249.91-1~deb11u1

[...]

#14 6.103 Chromium 106.0.5249.30 (playwright build v1024) downloaded to /root/.cache/ms-playwright/chromium-1024
```

Are these for different purposes?

web/conftest.py
rebeccacremona: I did this in [Capstone too](https://github.com/harvard-lil/capstone/blob/develop/capstone/docker-compose.yml#L47); I don't remember if there is any reason I did it in `docker-compose.yml` instead of testing setup like you did. Hmmm, probably not. 

I don't know if you'll end up needing that in other contexts for the PDF export... Probably not, because at least if mediated by celery workers, that will be running in a separate python process... though might be needed in dev with `CELERY_TASK_ALWAYS_EAGER = true`? We'll have to see!

bensteinberg:
> Are these for different purposes?

Yes -- the Chromium we specify in the Perma Dockerfile is for capturing; if in this context, only playwright is using Chromium, you don't need that one. (If you do need a non-playwright Chromium, though, I think you need to add our registry, as in https://github.com/harvard-lil/perma/blob/develop/perma_web/Dockerfile#L68-L70)

docker/Dockerfile
```diff
@@ -21,3 +21,7 @@ COPY web/requirements.txt /app/web
 RUN pip install pip==21.3.1 \
     && pip install -r requirements.txt \
     && rm requirements.txt
+
+# Install Playwright browsers for integration tests
+RUN apt-get update && apt install --no-install-recommends libdbus-glib-1-2
```
bensteinberg:
```suggestion
RUN apt-get update && apt install -y --no-install-recommends libdbus-glib-1-2
```

rebeccacremona:
> Are these for different purposes?

Yeah, if I can remember... when we first switched to Chromium from Chrome (because there was no build of Chrome that worked inside Docker on our M1 Macs), we considered using the Playwright script to install it, but we didn't like that we couldn't specify which version we would get: you can only point the Playwright install script at a particular channel, like stable or canary. For something as core to Perma as "which browser is being used for capture?", we wanted it to be completely deterministic, so that new capture workers could be spun up with old versions even if new versions had subsequently been released and the whole fleet upgraded at once; we wanted to be able to downgrade if it turned out a new version wasn't capturing well, etc.

I also think I spent some time trying to reuse that Chromium executable with the Playwright tests, so that there wasn't more than one installed, but I remember that running aground/not feeling worth the effort to get working.

If you are going to be using Playwright + chromium in prod to make PDFs, I'm not sure which way you want to go: use the Playwright scripts to install the latest versions of Chromium on production minions, or try and wrangle Playwright to talk to a separately installed and updated one.

rebeccacremona (approved):
LGTM!

If of interest, I also found this useful in Perma:

Patch to get [full-screen screenshots](https://github.com/harvard-lil/perma/blob/develop/perma_web/conftest.py#L10-L18) instead of just the viewport, and config for [automatic screenshots on failure or error](https://github.com/harvard-lil/perma/blob/develop/perma_web/setup.cfg#L16).

(I wasn't getting full-screen screenshots with CAP when I added the patch, even though it works in Perma; didn't spend a lot of time working out why)

lizadaly:
Thanks to both of you for your help!

PR #1795, opened 2022-10-11
Drop livesetting and additional permission checks for printable view

lizadaly:
When the printable HTML view was first added, a feature flag was put around it (to only enable it in certain environments) and an additional permission check was added to ensure the user was an admin.

The feature has been enabled on prod for months with no issues, so it should be safe to remove these extra checks, even if we don't yet expose the view via a visible link to non-admin users.

This changeset:

* Drops the requirement for the feature flag in the view
* Removes the field from the `LiveSettings` model
* Drops the additional requirement that the user be an admin

Now the URL is technically available to any user who can pass the `viewable_by` check on the Casebook instance, including unauthenticated users if the casebook is public. However, this PR does not expose the link to the view to end users, so it's merely obscured rather than protected (which I happen to think carries no risk but happy to talk about it).

This is happening now in part to facilitate allowing a playwright-driven client to request the printable HTML endpoint without its user having to be an admin, though still TBD how it will inherit author permissions to render a casebook that is not yet public.


bensteinberg (approved):
LGTM

PR #1803, opened 2022-10-14
Upgrade pytest and related dependencies to latest; migrate some doctests to pytest

lizadaly:
This upgrades the testing dependencies pytest, pytest-django, and playwright to the latest versions. 

Doing so broke behavior in the doctests that depended on the fixture `reset_sequences`, which is meant to revert any primary key values between test runs. To do so it needs to run before any other fixtures that create db objects. Something about this version of pytest/pytest-django no longer guarantees the order in which the fixtures are run (or perhaps changes the transaction behavior), since in manually debugging the doctest assertions, the only changes were the computed pk values.

One complication is that many of the failures weren't in unit tests at all, but in assertions in doctests in test helpers that the tests call.

Rather than try to restore the original behavior, I opted to make a few structural changes that IMO reduce surprise:

* Remove doctest assertions from any affected test helpers. Including assertions in helpers violates the general principle that [unit tests should be limited in the scope of their test coverage]( https://learning.oreilly.com/library/view/unit-testing-principles/9781617296277/OEBPS/Text/kindle_split_013.html\#:-:text\=When%20you%20see%20multiple%20act,id%20such%20a%20test%20structure.). In addition, these assertions were being called multiple times per test run, and while they're fast, that's unnecessary work.
* Move complicated tests out of doctests and into pytests. The cloning doctest was (at least) two distinct tests and is now two pytest functions.

I don't believe I otherwise changed any test behavior except to remove no-op or unused statements, which were flagged by flake8 now that they're visible to tooling. I don't believe test helpers themselves need test coverage, but since the documentation aspect is useful I retained the example output in each helper's doc string, but now as plain documentation.

(It might also be better in general to avoid relying on assertions related to primary key values, since requiring specific fixture order is a code smell on its own.)


lizadaly:
This is now ready for review!

bensteinberg (approved):

PR #1804, opened 2022-10-17
[DRAFT] Generate a PDF from a Playwright script; allow serving whole books

lizadaly:
(This depends on #1802 so still draft until that is approved and merged.)

Trigger PDF generation through a Playwright client pointed at an arbitrary URL.

The URL is expected to be the printable HTML endpoint with the new (to this PR) `whole_book` option, where all sections are rendered at once. For large books this endpoint is not ideal for humans to access, but is necessary to give PagedJS the ability to paginate the entire thing in one go.

This can be invoked as:

```
python main/pdf.py http://opencasebook.test:8000/casebooks/6105-annotations-tests/as-printable-html/all/ \ 
   annotations.pdf
```

A few related changes in this PR:

* The navigation footer in the printable view has been moved to the bottom of the page and varies if the user is in "whole book" mode. In both modes it lets the user exit out of the print preview and go back to the main casebook edit screen. 
* Added `defer` to the JS imports which means they'll run when the DOM is ready, obviating the need a listener for `DOMContentLoaded` (thanks Http 203!)
* Added an additional public casebook to the functional test fixtures for use by the PDF test.

Includes a simple Playwright test that verifies that a PDF file (or something resembling it) was created.

## TODO

Very long (1,000+ page) PDFs will often fail to render in limited-resource environments. I think the problem is with PagedJS itself failing to complete pagination rather than anything to do with PDFs or Playwright, but this needs more testing. (A book that renders fine when running directly from my laptop's virtualenv may fail to complete on the same computer inside the Docker container.)

There's currently no ability for the PDF generation routine to share the authentication credentials of the user, so only published books can be generated this way.

### Footer sample
image

### PDF sample
[annotations.pdf](https://github.com/harvard-lil/h2o/files/9803728/annotations.pdf)


lizadaly:
Gonna reapply these changes to a clean branch.

PR #1808, opened 2022-10-21
Stop "npm run serve" overwriting webpack-stats.json

jcushman:
Friday funday coding project! A small quality of life improvement to stop `fab run_frontend` from overwriting `webpack-stats.json`. This is a second attempt at #1732. I called the new stats file `webpack-stats-serve.json` to emphasize that it's the file created by running `npm run serve` instead of `npm run build`, and added a `LIVE_JS_ASSETS` environment variable (set by `fab run_frontend`) to tell Django that it should look for live assets from `npm run serve` instead of built assets from `npm run build`.

Note, I think it might be considerably clearer to call this file `vue-bundle-paths.json` (and `vue-bundle-paths-serve.json`) instead of `webpack-stats.json`. But that seemed confusing to mix in here, so, eh, something to consider.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ python-shell.sh
 *.iml
 .vscode
 .mypy_cache
+
+# Dev-only webpack stats
+web/webpack-stats-serve.json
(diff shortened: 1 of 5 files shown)


lizadaly (approved):
Thanks! This worked as I expected when I ran it locally. It'll need to merge in the bcrypt update to deploy though.

PR #1816, opened 2022-11-09
Add first published and last modified dates to casebook page

### LINKED ISSUE #1303: Add date info to Casebooks
Perhaps Year Published and then also a last updated date? 

(note that I see this in the search area, but not when viewing a casebook)

lizadaly:
Puts some key dates for the Casebook on the "entry page" for the book (the page you land on from search).

image

I put the date info on the right-hand side next to the title.

The history model (`CasebookEditLog`) only tracks "first published", not "most recently re-published", which is a somewhat nebulous concept since republished drafts are new books merged back into old ones. Hopefully this value is good enough here.

Most-recently-modified is the more recent of either when the casebook was edited or when it was annotated.

There weren't good hooks into the edit log model from the Django template side—the History page uses a somewhat complicated function to group changes. So I added two straightforward properties on the `Casebook` model.

## Testing 

I didn't include unit tests here though I could—since these properties are only called by a template (which produces no output if it errors) and are just encapsulated ORM queries, I felt it was safe enough to just manually check.

More screenshots in the linked issue.

Fixes https://github.com/harvard-lil/h2o/issues/1303

### DIFF
--- a/web/frontend/styles/casebook.scss
+++ b/web/frontend/styles/casebook.scss
@@ -295,14 +295,16 @@ form.edit_content_resource, form.edit_content_section, form.edit_content_caseboo
     margin-bottom: 2px;
     padding: 20px 20px 20px 30px;
     background-color: $white;
+    display: flex;
+    justify-content: space-between;
+
     .casebook-title {
       @include sans-serif($bold, 20px, 24px);
       color: $dark-gray;
     }
-    .reading-time {
-      float: right;
+    .casebook-date {
       color: $dark-gray;
-      font-size: 16px;
+      font-size: 14px;
       font-weight: 500;
     }
   }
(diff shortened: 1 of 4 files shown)


web/frontend/styles/casebook.scss
```diff
@@ -295,14 +295,16 @@ form.edit_content_resource, form.edit_content_section, form.edit_content_caseboo
...
     padding: 20px 20px 20px 30px;
     background-color: $white;
+    display: flex;
+    justify-content: space-between;
+
     .casebook-title {
       @include sans-serif($bold, 20px, 24px);
       color: $dark-gray;
     }
-    .reading-time {
```
lizadaly, line 302: This appeared to be unused—`reading-time` only appears on the section pages, and styles for that are defined further down in the file.

rebeccacremona (approved):
Nice!

I like the `space-between`.

Issue #1050, opened 2020-01-10
Modals and redirects


rebeccacremona:
From https://github.com/harvard-lil/h2o/blob/develop/web/frontend/legacy/lib/requests.js#L35-L45
```
      // This code actually expects successful AJAX requests to receive
      // a redirect as a response. Axios follows the redirect, performing
      // a GET of the page we want the browser to display next. Here, we
      // examine the URL of we just retrieved, and compare it to the URL
      // of the page we are presently on. If we are already on that page,
      // we perform a refresh. Otherwise, we direct the browser there.
      // In all cases, this performs a duplicate GET.
      //
      // Let's have a think on how we might redesign this in the future.
      // Probably a mix of more Vue, and switching from AJAX to standard
      // form submissions.
```

lizadaly:
Good to clean up when working with the JS again in a major way.

rebeccacremona:
Re-opening because we just got very confused by this behavior, when trying to add a flash message after the publication of book. (PATCH -> reload; which looks a lot like, but is not, a normal top-level form submission)

lizadaly:
There are likely more instances of this than the publish endpoint, but I think now that we have some examples we can open specific tickets for any additional cases we want to modify.

PR #1817, opened 2022-11-14
Use django's redirect shortcut rather than raw HttpResponseRedirect

lizadaly:
A little bit of cleanup in advance of #1050.

Use Django's [redirect() helper function](https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#redirect) instead of `HttpResponseRedirect` directly to eliminate some clunky boilerplate with `reverse()` and arg-passing.

There is one functional change here: two places where the code indicated that the redirect was to a "canonical" URL now redirect with a 301 moved permanently rather than 302. No tests detected this change but I verified in the network console that the behavior still worked (for the case of redirecting a resource URL to the canonical form with the title slug).

### DIFF
--- a/web/main/views.py
+++ b/web/main/views.py
@@ -24,11 +24,10 @@ from django.http import (
     HttpResponse,
     HttpResponseBadRequest,
     HttpResponseForbidden,
-    HttpResponseRedirect,
     HttpResponseServerError,
     JsonResponse,
 )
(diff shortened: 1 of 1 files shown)


matteocargnelutti (approved):
LGTM 👍 !  Good catch re 301 vs 302!

PR #1818, opened 2022-11-14
Drop Link.content_type field that is unused; fixes #1109

### LINKED ISSUE #1109: Get rid of content_type from Link
From #872 : `Why do "Link" objects have a "content_type" field? Do we need it?`
Looking at this, it doesn't look like we're using it anywhere, or setting it now. At some point in the future if we look at embedding content we should revisit.

lizadaly:
Drops this field that is sparsely populated and not displayed in the UI. 

See https://github.com/harvard-lil/h2o/issues/1109#issuecomment-1179372547 for an analysis of the existing data.


rebeccacremona (approved):
🙌

PR #1824, opened 2022-11-16
Fix "remove myself from casebook" and inviting duplicate collaborators

lizadaly:
Two unrelated bug fixes on the casebook settings page:

* Provide a CSRF token for the "remove myself" flow.
* Check whether a collaborator exists for a given user/casebook combination before trying to create a new one.

This message added in #1823 now works, as does the feature:

image


bensteinberg (approved):
🤘

PR #1826, opened 2022-11-17
Redirect-after-POST for edit_casebook; reduce duplicate code in sources check

lizadaly:
The main thing here is continuing #1050 work to make Django view pages redirect as normal, this time for the casebook edit page.

This also reduces a bunch of repeated code to draw from the list of possible sources for searching using the quick add. I suspect that this code could actually be excised ultimately—if we want to add beta search sources going forward I think we should use a more traditional feature flag approach rather than custom permission checks. All I did here was package up the repetition into a class function that should have the same logic.

### DIFF
--- a/web/main/views.py
+++ b/web/main/views.py
@@ -1718,7 +1718,7 @@ def create_draft(request, casebook):
 @require_http_methods(["GET", "POST"])
 @requires_csrf_token
 @hydrate_params
-def edit_casebook(request, casebook: Casebook):
+def edit_casebook(request: HttpRequest, casebook: Casebook):
     """
     Given:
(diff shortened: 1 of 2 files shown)


web/main/views.py
```diff
@@ -1731,7 +1731,7 @@ def edit_casebook(request, casebook: Casebook):
     ...    content_includes=[private.title, "You are viewing a private casebook"],
     ... )
     >>> check_response(
-    ...     client.post(private.get_edit_url(), {'title': new_title}, as_user=private.testing_editor),
+    ...     client.post(private.get_edit_url(), {'title': new_title}, as_user=private.testing_editor, follow=True),
```
lizadaly, line 1734: This tells the Django test client to follow the redirect-after-post, so the `check_response` function eventually gets the 200 OK response it's expecting.

matteocargnelutti (approved):
Nice 😎

PR #1830, opened 2022-11-22
Update Pillow

bensteinberg:
https://pillow.readthedocs.io/en/stable/releasenotes/9.3.0.html


lizadaly (approved):
I'm unfamiliar with what we actually use Pillow for here, but looking at the changelist I'm reasonably confident this should be a fine upgrade.

PR #1835, opened 2022-12-05
Make section headings searchable in "Search Inside"

### LINKED ISSUE #1832: Search Inside doesn't seem to pick up on text in section headings

lizadaly:
Fixes #1832 

Adds section title content to the "Search Inside" feature.

This was a little less straightforward than I'd assumed at first because a `TextBlock` and a section aren't the same kind of thing. Since the `Section` proxy class is DOA, this code only cares about `ContentNodes` with the right (usually null) resource type.

I assumed we'd want to group section results in with "Text Passages" so it looks like this (the first two results are sections):

image

I considered labelling the sections as such but it doesn't seem necessary. Clicking on the section title takes you to the section.

Also shortened this function mostly by making some function calls guaranteed to return empty string or empty list results so a number of if/else checks could be eliminated.

### DIFF
--- a/web/main/create_fts_index.sql
+++ b/web/main/create_fts_index.sql
@@ -43,6 +43,24 @@ UNION ALL
         INNER JOIN main_contentnode cn ON cn.resource_id = t.id AND cn.resource_type = 'TextBlock'
     GROUP BY t.id, cn.id
 UNION ALL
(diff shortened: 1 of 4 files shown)


bensteinberg (approved):
👍

Issue #1841, opened 2022-12-08
Add "reader view" mode 


lizadaly:
This experimental mode will leverage the printable HTML view, with the following characteristics:

* It should use the printable HTML Django templates and JS that adds annotations/elisions.
* It should _not_ display the page through PagedJS, but instead use a standard single-column scrolling layout.
* It should be paginated at the top-level "section" level, not at the individual resource level.
* It should be possible to toggle in and out of this mode.
* It _should_ re-use all the custom print CSS associated with the printable view that are supported by browsers (e.g. not polyfilled by PagedJS). To the extent possible this should give us high-quality printed output that just misses out on some of the book-ish affordances of PagedJS.

Some example differences from the PagedJS layout:

* Annotations will not be rendered as footnotes, but as web-friendly sidebar annotations. 
* Hyperlinks will be rendered as normal underlined and active links, not as footnotes.
* Elisions can (potentially) be toggled open.

lizadaly:
Still to do:

* Change hyperlinks rendering to normal, active links
* Ensure Matomo tracker is in the custom HTML head (or use the site-wide `base.html` if that doesn't interfere with PagedJS).

lizadaly:
* Consider using the left-nav for persistent section/resource headings
* Margins are overly generous
* Small viewport layout

lizadaly:
> one thing that might be a helpful add on would be an easy way to navigate back to the toc once you're in a specific resource on the reader page

PR #1842, opened 2022-12-08
Add basic version of "reader mode" for viewing casebooks in longer form

lizadaly:
First steps in making a usable "reader mode" for students or people just looking to read a casebook end-to-end, or print it.

This is now the default mode for the "printable HTML" view—the two column print layout is now triggered with an explicit "Print preview" button. This view is still only available to staff users.

Some of this is a little janky and is starting to push at the edges of not using Vue for the client-side controls. When the functionality gets nailed down more I can look at whether it makes sense to put this under Vue's control. Alternatively I could just polish the rendering steps up so there's fewer repainting/layout changes.

When you click "HTML preview" now you get this view:

image

The footer lets you toggle among the different modes, or page through the casebook.

Notes and highlights are rendered somewhere between the default edit view, and the print-preview view. The note target here is dotted-underlined; we can look at other kinds of indicators that are less likely to be confused with hyperlinks.

Clicking "Print Preview" takes you to the two-up view:

image

Note that the annotation is now a footnote. Same markup though!

 
part of #1841


bensteinberg (approved):
Cool!

PR #1845, opened 2022-12-12
Add discovery/featured books page; fixes #1838

### LINKED ISSUE #1838: Add Discovery page and copy to home page
We'd like to feature some specific casebooks that are good starters for professors looking to adapt content.

* Hand-curate a set of casebooks, organized by class or topic, to suggest as books to clone. Potentially include headshots and information about the authors.
* Write a standalone page on H2O to list the full set of books to feature.
* Also develop copy or short previews that can be placed inline on other pages—the home page, the "create a casebook" modal.

One thing we should keep in mind is that anyone who has already logged in almost never sees the unauthenticated home page, so we probably want to feature this content somewhere that authenticated users can find it too.

Goal is to have this up before the end of the year break.

lizadaly:
Adds a featured casebooks page to be linked from the home page and other places. Currently it's not linked from anywhere.

image

This is implemented as a template tag which accepts a casebook ID as input, and any of the displayable fields as optional parameters. If those are provided, they'll be used as-is instead of the default values from the casebook itself.

A few casebooks look missing here because my local database doesn't have them; this is a good test of what happens if they aren't available. In view source you'd see:

```html
    <!-- Error: Casebook ID 6195 was not found in this environment-->
```

For casebooks which exist but aren't publicly viewable, they are also omitted but show something different in view source:

```html
<!-- Error: Casebook ID 6094 is not publicly viewable--> 
```

If we want to keep this information private I could remove it from the template output, but I think it's safe enough—we already use auto-incrementing IDs which are easily discoverable.

## Analytics

Adds the analytics doodads based on the Matomo documentation, but that's not testable locally.

## Mobile

Two columns become one:

image

## Testing

Added a few unit tests for the template tag, which is the one moving part here.

# Additional styling 

Also rationalizes the widths of all the static pages because this was depressing me:

## Before

image

## After

image

Lastly, I removed a wrapper div `content` which was in all the static pages but didn't actually do anything and had no references in the relevant SCSS.

### DIFF
--- a/web/frontend/styles/header.scss
+++ b/web/frontend/styles/header.scss
@@ -37,7 +37,7 @@ body {
         // The text is hidden and is replaced visually by an image
         background-image: url('~static/images/logo.svg');
         background-position: 10px center;
-    background-repeat: no-repeat;
+        background-repeat: no-repeat;
         background-size: initial;
         color: transparent;
 
--- a/web/frontend/styles/pages.scss
+++ b/web/frontend/styles/pages.scss
@@ -1,10 +1,24 @@
 
 .pages-content {
-  padding: 20px 20px;
-  width: 80%;
+
+  padding: 5vh 15px 10vh 15px;
+
   margin: 0px auto;
   font-family: $font-family-sans-serif;
 
+  @media (min-width: 768px) {
+    width: 750px;
+  }
+  @media (min-width: 992px) {
+    width: 869px;
+  }
+  @media (min-width: 1200px) {
+      width: 1170px;
+  }
+
+  h1 {
+    margin: 0 0 4rem 0;
+  }
   h2 {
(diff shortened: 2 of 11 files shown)


rebeccacremona (approved):
This looks great! Love the tiling.

Musing... I wonder if "Upper Level Courses" and "Graduate and Undergraduate Courses" level headings might want a little more separation from the preceding content. In this glance, I feel like they aren't standing out. But I don't know if that musement is correct :-)

> Also rationalizes the widths of all the static pages because this was depressing me:

Yes! Thanks!

cath9:
Would echo Becky's musing about spacing, particularly beneath the "Graduate and Undergraduate Courses" heading (which I understand is probably different bc there is no subheading, but if there is a way to pad it that would be great.)

PR #1848, opened 2022-12-13
Reader mode header and other enhancements

lizadaly:
## Analytics and tracking 

* Move Sentry code to an include and use it in both the general site header and the reader view
* Ensure that Matomo analytics are capturing the reader view

## Mobile layout

* One-column layout for mobile view
* Don't show print-preview button in mobile view (the two-up view is not useful here)

## Enhancements

Hang the current section title in the left nav. I'd like this to be `sticky` and be present as the user scrolls down, displaying the current section, but that needs some JS. TODO.

image


bensteinberg (approved):
👍

Issue #1782, opened 2022-09-27
Allow a Casebook to be tagged with one or more areas of interest by the author


lizadaly:
From a controlled vocabulary, allow users to tag Casebooks with one or more areas of interest.

Tab 3 on [this spreadsheet](https://docs.google.com/spreadsheets/d/1OOpsVo0eyJjkPDhVeRHNho0maUbZohIe2twoI_xnKDI/edit#gid=1972631599) lists areas of interest that a Casebook could be tagged with.

* The controlled vocabulary should be a new type of model field
* The controlled vocabulary should be editable in the Django admin
* Adding or removing tags from Casebooks should be possible in the Django admin 
* Adding or removing tags from Casebooks should be possible for the author on the front end (TBD where)
* It should (eventually) be possible to search on or browse through public Casebooks based on their tags

There's probably existing Django packages that do this, so look at some prior art and decide whether it's simpler to use a third-party package or roll our own.

PR #1855, opened 2022-12-19
Add back-end support for tagging casebooks

varun-magesh:
This partially addresses #1782, which says:
> From a controlled vocabulary, allow users to tag Casebooks with one or more areas of interest.

Though there are a few different out-of-the box solutions for tagging in Django, there is only one project with a large community and continued maintenance: django-taggit. In general, tagging projects including taggit seem to focus on flexible, social-media type tagging rather than the controlled vocabulary we are interested in. It seemed like taggit would take substantial customization at the level of its most basic models to support a fixed vocabulary properly. In light of that cost, it seemed to make the most sense to roll our own simple tagging solution.

This PR adds backend support for Casebook tagging. It creates a new `CasebookTagNames` model which represents the list of permitted tags. It creates a named many-to-many relationship between tag names and casebooks. This allows us to take advantage of Django's helpers and when the time comes to display or filter by tags we will be able to do convenient things like `casebook.tags.add()` or `tag.casebooks.all()`.

The list of tag names and each casebook's tags are editable through the Django admin interface.

No tests on this one because it's mostly Django boilerplate, but I could add some if they would be helpful.

2022-12-18-234704_1411x394_scrot
2022-12-18-234650_1013x418_scrot


web/main/models.py
```diff
@@ -1429,6 +1429,29 @@ class Meta:
...
+        "CasebookTagName",
+        on_delete=models.CASCADE,
+    )
+
+    class Meta:
+        unique_together = (("tag", "casebook"),)
+
+
+class CasebookTagName(models.Model):
+    slug = models.CharField(max_length=100, blank=True, null=True, unique=True)
```
lizadaly: `SlugField` here? ([ref](https://docs.djangoproject.com/en/3.2/ref/models/fields/#slugfield)). Also in general you probably don't want both `unique=True` and nullable. Typically I think of slug-type fields as being required but [auto-populated](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.prepopulated_fields).

web/main/models.py
```diff
@@ -1429,6 +1429,29 @@ class Meta:
...
+    tag = models.ForeignKey(
+        "CasebookTagName",
+        on_delete=models.CASCADE,
+    )
+
+    class Meta:
+        unique_together = (("tag", "casebook"),)
+
+
+class CasebookTagName(models.Model):
```
lizadaly: I was a little confused briefly by the model naming here—I think maybe just call this a `Tag` and call the other one `CasebookTag`, as you have. The convention for many-to-many through models is to name the thing the concatenation of the two model names, so `Casebook + Tag` makes sense but then I had trouble parsing what I assume is  `CasebookTag [a type of tag] Name`.  Just calling it `Tag` resolves the issue, plus it leaves open the possibility of using it for sub-casebook level tagging.

web/main/models.py
```diff
@@ -1429,6 +1429,29 @@ class Meta:
         unique_together = (("user", "casebook"),)
 
 
+class CasebookTag(models.Model):
```
lizadaly: Usually the reason to use a through model rather than just a direct many-to-many is to add additional fields on the relationship. However, it's so annoying to change your mind later that I usually recommend starting with a through model—so I think this is right. Did you have additional fields in mind you might want to add here? I could imagine, for example, recording the user who did the tagging operation.

lizadaly (commented):
This looks like a terrific start, and I agree with your instinct that the use case here is small enough that rolling our own is probably the right course.

I'm not sure what's up with that failing test; I don't know what those "historical casebook" model tests are meant to check for, and it's possible they've outlived their usefulness.

One thing I'd strongly recommend adding are creation and modification date fields to both models. There might be a base model class we have that already include them; otherwise. you can probably just copy them from another model to have consistent naming.

lizadaly:
Oh and I agree that the final version of this PR, assuming equivalent coverage, won't need tests yet.

web/main/models.py
varun-magesh: Yes, the unique/null issue makes sense, thank you for catching it! Didn't know about either `SlugField` or `prepopulated`, I'll modify this to use those.

web/main/models.py
varun-magesh: Yeah I was really unsure about the naming --- I tried a few different variants all of which were equally confusing. `Tag`/`CasebookTag` makes sense; it is also more clear to me now from @cath9's outline that we are hoping to tag more than just `Casebook`s. It seems like I should also make `Tag` use `GenericForeignKeys`? Or perhaps a real `ForeignKey` to a `ContentNode` would be better, since I think (not quite sure) we could infer that a top-level node represents a casebook.

web/main/models.py
varun-magesh: Yes, that's what I was thinking. I didn't have anything specific in mind, more just future-proofing. I think I also just like being able to see the table for myself, in a way.

User makes sense; I'll add that in! The other thing I was thinking of was `type` if we are going to link to more than just `Casebook`s.

varun-magesh: Having looked at the plans for this, it seems like it would be helpful to have a type for the tag itself, rather than just the object it points to. E.g. topic and institution tags both point to casebooks, but they represent something different, so maybe se should note that --- probably on the `CasebookTag` model (the old `CasebookTagName`)

web/main/models.py
lizadaly: I'm not a big fan of `GenericForeignKey` unless you really definitely absolutely have no idea what the connecting models might be. Here's a good writeup of the problems and some alternative architectures: https://lukeplant.me.uk/blog/posts/avoid-django-genericforeignkey/

For this case I think it'd be perfectly okay to add `ContentNode` later as a second kind of relation, sort of like [Alternative 5](https://lukeplant.me.uk/blog/posts/avoid-django-genericforeignkey/#:~:text=Alternative%205%20%2D%20multiple%20linked%20models) here. For your purposes, I recommend just pretending like we're only worrying about Casebooks for now—I think that'll help our decision-making on a lot of fronts to start simple.

web/main/models.py
```diff
@@ -1429,6 +1429,45 @@ class Meta:
...
+    display_text = models.CharField(max_length=100, unique=True)
+    casebooks = models.ManyToManyField("Casebook", through="CasebookTag", related_name="tags")
+
+    class Category(models.TextChoices):
+        INSTITUTION = "institution", "Institution"
+        RESOURCE_TYPE = "resource_type", "Resource Type"
+        INTEREST_AREA = "interest_area", "Interest Area"
+
+    category = models.CharField(
+        max_length=16,
```
lizadaly: There's no advantage to limiting `max_length` to small values in most cases—it doesn't save resources (Postgres doesn't care about the length) so the only use case is for user validation. For example, a US social security number is always 9 digits; having anything longer would be an error, and that's useful to enforce.

So make it 100 or so so we don't get annoyed when we want to add a multiword category that's longer than 16?

lizadaly (approved):
Pending the test update and one minor suggestion, this looks good!

For the failing test:

There's a few things going on here. One is that the `assert_num_queries` code itself has a bug when called with `pytest -vv`, which is the default in CI. This [throws an exception](https://github.com/harvard-lil/h2o/actions/runs/3836232977/jobs/6530180734) instead of actually reporting:

```python
UNEXPECTED EXCEPTION: KeyError('stack')
  File "/usr/local/lib/python3.9/doctest.py", line 1334, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest main.models.HistoricalCasebook.instance_type.delete[9]>", line 2, in <module>
  File "/usr/local/lib/python3.9/contextlib.py", line 126, in __exit__
    next(self.gen)
  File "/app/web/conftest.py", line 823, in _assert_num_queries
    q["stack"]
KeyError: 'stack'
```

Running pytest locally but without the verbose flag lets you see what it's trying to report:

```
Unexpected queries: expected {'delete': 14, 'select': 20, 'insert': 36}, 
got {'select': 20, 'delete': 16, 'insert': 36}
```

I was not able to figure out how to get the verbose reporting to work.

After some unpacking:

* The inserts come from `HistoricalCasebook`/`django-simple-history`. They're recording change records to the history log when the casebook is modified. 
* The selects haven't changed. 

There are now two new deletions. This potentially makes sense because there's a `CASCADE DELETE` setting for `CasebookTag`, but at first I was confused by this because the test itself doesn't add any tags  to be deleted.

I enabled verbose query logging for the application as a whole and see:

```
DELETE FROM "main_casebooktag" WHERE "main_casebooktag"."casebook_id" IN (1); args=(1,)
DELETE FROM "main_casebooktag" WHERE "main_casebooktag"."casebook_id" IN (2); args=(2,)
```

So I think what's happening here is that it's optimistically calling deletes based on the database relationship but not actually doing any work, because there are no records that need deleting.

**TL,DR**: Increment the expected number of deletes by 2 in https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3666-L3667

web/main/models.py
varun-magesh: That makes sense!

varun-magesh:
> **TL,DR**: Increment the expected number of deletes by 2 in https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L3666-L3667

Ooh, that is a very interesting interaction. The reasoning behind the optimistic delete makes sense but it never would have occurred to me without being told. Thank you so much for the investigation and detailed explanation!

lizadaly:
Thank you, this looks great! I really appreciate you sticking with it through all the suggestions!

PR #1858, opened 2023-01-04
Change "Search Inside" tab to "Find"

lizadaly:
Per @cath9: 

image


lizadaly:
@cath9 Should the placeholder text also say "Find in <title>" instead of "Search in <title>"?

matteocargnelutti (approved):
👍

cath9:
Very open to conflicting opinions, but I was imagining the tab would just say "Find".

cath9:
Ignore the previous comment. Yes, the text in the field can say "Find in <title>"

lizadaly:
image

Issue #1861, opened 2023-01-09
Case header is omitted in reading mode


lizadaly:
Example: Casebook 7606, section 1.3.2, the case header is omitted, and the metadata from the case itself is hidden by a CSS rule (originally from the edit view to prevent duplicating information).

Ideally see if the case header custom rendering from the main site can be reused as an include.

PR #1863, opened 2023-01-09
Disambiguate headnotes from other text in reading mode

### LINKED ISSUE #1862: In reading mode, provide differentiation between headnote content and case/custom content
See casebook 7606 section 1.3.2, where it's not possible to tell where the author's headnote ends and the case text begins (this is exacerbated by #1861 but would be true regardless).

In the Word export this is differentiated based on a sans-serif font (headnote) and a serif font (case). On the main site the headnote is in the tan background color (`<section class="headnote">`).

lizadaly:
In reading mode, put a light shadow around case content to set it apart from the headnote:

image

This is meant to evoke the notion that this is a whole document embedded inside the casebook.

The treatment needs to be different in the paginated, printable view, because the border ends up appearing on every page and isn't padded correctly. In this view instead I just set it off with a light line:

image

This also fixes an unintentional regression in the PagedJS layout where styles meant to target the print CSS weren't made available to PagedJS.

Fixes #1862 and #1861

### DIFF
--- a/web/frontend/pages/as_printable_html.js
+++ b/web/frontend/pages/as_printable_html.js
@@ -268,7 +268,8 @@ annotationRanges.forEach((rg) => {
 });
 
 function renderAsPagedJS() {
-  paged.preview(tmpl.content, [], main).then((flow) => {
+  const css = tmpl.getAttribute("data-stylesheet");
+  paged.preview(tmpl.content, [css], main).then((flow) => {
     console.log("Rendered", flow.total, "pages.");
     main.classList.add("preview-ready");
   });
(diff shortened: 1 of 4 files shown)


cath9:
one thing I like about the website design is that the document is the thing apart, ie it is against a white background while the reset of the site is the tan you mentioned. So it looks like an embedded legal document, which is what it is. If what you propose above is the easiest format for now, perhaps the document is outlined, rather than the headnote.

lizadaly:
@cath9 I can do whatever! That’s a helpful framing for the intent. I have some alternate ideas to try.

lizadaly:
This is ready for review. @cath9 approved the layout of the reading view.

rebeccacremona (approved):

PR #1867, opened 2023-01-11
In reading mode, don't set height to an invalid value

### LINKED ISSUE #1866: Reading mode TOC doesn't expand in Safari because Safari
The toggle toggles, but nothing appears (probably the selector doesn't work).

lizadaly:
fixes #1866 

warning very complicated PR!!!!1


matteocargnelutti (approved):
@lizadaly Nice, but can you simplify a bit?

PR #1868, opened 2023-01-11
Add links to subsections in the nav in reader mode, fixes  #1860

### LINKED ISSUE #1860: In reading mode, the TOC should be linked at full depth
Currently only the top-level chapter heading is linked. All the sub chapter links should also work and direct people to the correct anchor on the page.

lizadaly:
A bit of a rewrite in how the TOC is laid out:

image

I changed the markup in the TOC to be less fragile, then added indentation to the sub-section items. As before, the TOC doesn't descend further than 2 levels, though it could.

Each item in the toc is now hyperlinked, not just the top items. If you follow them, you go to the section via an anchor link.

As before, the underlining is not visible when printing: 

image

Also modifies the display of `Link` items, which I think wasn't working well. Now it's less exuberant:

image

### DIFF
--- a/web/main/serializers.py
+++ b/web/main/serializers.py
@@ -102,6 +102,7 @@ class ContentNodeSerializer(serializers.ModelSerializer):
             "ordinal_string",
             "ordinals",
             "is_instructional_material",
+            "slug",
         )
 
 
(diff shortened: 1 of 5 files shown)


matteocargnelutti (approved):
Nice 😎 !

Issue #1801, opened 2022-10-12
Add a celery job that will generate a PDF on behalf of a user


lizadaly:
The job should take as input:

* The casebook to be rendered
* The user on behalf of which to render it

If the user does not have access to the casebook, it should not generate a PDF.

Otherwise it should:

* Invoke playwright with the URL to the casebook to the full-book PDF preview
* Store the PDF on the S3/minio filesystem

It should return:

* A success or error message based on the playwright response 
* The user-accessible URL to the pdf file 

The URL may be time-limited for private PDFs.

lizadaly:
This is done!

PR #1882, opened 2023-01-18
Integrate Celery for running async tasks

### LINKED ISSUE #1798: Add a dev redis container for use by developers

rebeccacremona:
This PR adds Celery for [the running of asynchronous tasks](https://github.com/orgs/harvard-lil/projects/5/views/1).

See the README for more info on how to interact with it locally. This setup is identical to the one we used with the [prototyped perma-capture service](https://github.com/harvard-lil/perma-capture/tree/new-arch-with-vue) and is very similar to Perma.

I included one "demo" task and scheduled it to run once a minute in dev, a) to demo config, b) to make it easy to see if your dev setup is working, and c) to have a task available for the error email test.... but.... it might not want to stay in the repo long-term, once we have at least one real task.

~I set `CELERY_TASK_TIME_LIMIT` to a ridiculously short 2 seconds so that we will be forced to consider what an appropriate time limit is, once we have real tasks.~

I think this gets us in good shape to start [making PDFs asychronously](https://github.com/harvard-lil/h2o/issues/1801).

Closes https://github.com/harvard-lil/h2o/issues/1798.

### DIFF
--- a/web/config/settings/settings_base.py
+++ b/web/config/settings/settings_base.py
@@ -239,7 +239,7 @@ WEBPACK_LOADER = {
 ### CELERY settings ###
 ###
 
-CELERY_BROKER_URL = "redis://redis:6379/1"
+CELERY_BROKER_URL = "redis://redis:6379/0"
 CELERY_ACCEPT_CONTENT = ["json"]
 CELERY_TASK_SERIALIZER = "json"
 CELERY_RESULT_SERIALIZER = "json"
--- a/web/config/settings/settings_base.py
+++ b/web/config/settings/settings_base.py
@@ -235,6 +235,46 @@ WEBPACK_LOADER = {
     }
 }
 
+###
+### CELERY settings ###
+###
+
+CELERY_BROKER_URL = "redis://redis:6379/1"
+CELERY_ACCEPT_CONTENT = ["json"]
+CELERY_TASK_SERIALIZER = "json"
+CELERY_RESULT_SERIALIZER = "json"
+CELERY_TIMEZONE = "UTC"
+
+# These default time limits are useful in Perma, where capturing tasks
+# sometimes hang, for still-undiagnosed reasons, and where all tasks
+# are expected to be short-lived.
+#
+# It remains to be determined whether they will prove appropriate here.
+#
+# If a task is running longer than five minutes, ask it to shut down
+# CELERY_TASK_SOFT_TIME_LIMIT=300
+# If a task is running longer than seven minutes, kill it
+# CELERY_TASK_TIME_LIMIT = 420
+CELERY_TASK_TIME_LIMIT = 2
+
+# Control whether Celery tasks should be run asynchronously in a background worker
+# process or synchronously in the main thread of the calling script / Django request.
+# This should normally be False, but it's handy to not require the broker and a
+# celery worker daemon to be running sometimes... for instance, if you want to drop into pdb.
+CELERY_TASK_ALWAYS_EAGER = False
+CELERY_TASK_EAGER_PROPAGATES = True  # propagate exceptions when CELERY_TASK_ALWAYS_EAGER=True
+
+CELERY_TASK_ROUTES = {
+    "main.tasks.sample_scheduled_task": {"queue": "background"},
+}
+
+# from celery.schedules import crontab
+CELERY_BEAT_SCHEDULE = {}
+
+###
+### /END CELERY settings ###
+###
+
 CAPAPI_BASE_URL = "https://api.case.law/v1/"
 CAPAPI_API_KEY = ""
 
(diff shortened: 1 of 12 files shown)


bensteinberg (approved):
Cool! Before we merge this, let me set up a Redis instance for stage.

web/main/tasks.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@task_failure.connect()
+def celery_task_failure_email(**kwargs):
+    """
+    Celery 4.0+ has no method to send emails on failed tasks.
+    This event handler provides equivalent functionality to Celery < 4.0,
+    and reports truly failed tasks, like those terminated after CELERY_TASK_TIME_LIMIT
+    and those that throw uncaught exceptions.
+    From https://github.com/celery/celery/issues/3389
```
lizadaly: Is this an opportunity to start to move away from this reporting method? The reasons for removing the feature in Celery in the linked issue seem like good ones.

web/config/settings/settings_base.py
```diff
@@ -235,6 +235,46 @@ class LoggerConfig(TypedDict, total=False):
...
+# sometimes hang, for still-undiagnosed reasons, and where all tasks
+# are expected to be short-lived.
+#
+# It remains to be determined whether they will prove appropriate here.
+#
+# If a task is running longer than five minutes, ask it to shut down
+# CELERY_TASK_SOFT_TIME_LIMIT=300
+# If a task is running longer than seven minutes, kill it
+# CELERY_TASK_TIME_LIMIT = 420
+CELERY_TASK_TIME_LIMIT = 2
```
lizadaly: I would start with whatever the default base time limit is, and use the soft default rather than the hard kill, so more reporting is potentially captured. I dunno that 2 seconds would tell us much of anything? Rendering a PDF for a 500 page book will definitely take longer than that!

rebeccacremona: Ah, that was on purpose: I should have put my reasoning in a comment instead of the PR description. I set that to a ridiculously low number to forcibly remind us to think about and pick reasonable values when adding the actual PDF-making task.

[There is no default time limit](https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_time_limit); you have to set one explicitly. I definitely agree that we should set the "soft" limit to whatever we actually want the cap to be, and only use the hard limit to clean up clearly hanging tasks.... which might be less of an issue here than for Perma's capture process: I imagine there's less code that might potentially swallow the "soft" request to stop.

Happy to set some sensible values here instead if you like! What would you suggest? I haven't tried exporting any books yet.

web/main/tasks.py
rebeccacremona: Ah, did not even think about that once.

rebeccacremona: To me, it's confusing to have some kinds of things that result in error emails and some that don't. If we remove this code, which does sound sensible to do either now or soon, it seems to me that we should simultaneously stop H2O from sending error emails at all.

rebeccacremona: Either way sounds good to me!

lizadaly: Happy to work with you all to set up Slack notifications from Sentry next week.

web/config/settings/settings_base.py
lizadaly: Ah sorry, I see the default is no-limit; that's actually consistent with other job runners like Airflow and Prefect so I should've assumed that.

I would typically set the global timeout starting with double the expected duration of the longest task. We haven't audited all the possible tasks, but if we include the ones currently implemented with cron:

* Site-wide search indexing - fast, just a few seconds
* Full-text searching - slow, ~10 minutes
* Reporting tables - fast, but could do more when run async
* PDF generation - seconds and probably not minutes, but not benchmarked

Playwright will have its own timeout since it's a browser initiating a request. How fast it runs will be very environment dependent. I have observed PagedJS failing to render very large books if it's deprived of RAM, but we'll want Playwright to handle its own timeout anyway so it can complete writing any screenshots or video files. 

I like to treat global task timeouts as a measure of last resort to stop literally stuck jobs, and use reporting tools to evaluate whether task duration is creeping up or otherwise seems unhealthy. If one particular task has the characteristics of both being typically very short and safe to kill (an example might be some kind of notification counter—a nice-to-have that is okay if it's a bit stale) then I might set a per-task timeout that's short assuming it'll just catch up on a subsequent run.

I find it's better to err on the side of longer timeouts, that give infrequently-long tasks room to complete, than very short timeouts that repeatedly kill jobs that can't possibly finish until the max is increased.

bensteinberg:
Staging infrastructure is set up.

web/main/tasks.py
rebeccacremona: Removed. And, since this code included a test of the setup, added a test to the demo task instead.

web/config/settings/settings_base.py
rebeccacremona: Switched to a default soft time out of 20 minutes (twice the ~10min mentioned for full-text indexing) and a hard time out of 22 minutes.

Issue #1879, opened 2023-01-17
Streamline links in navigational footer in reading mode


lizadaly:
Implement this careful design exactly as proposed:

![image](https://user-images.githubusercontent.com/19571/212983667-28fa3235-6e13-4723-81c1-4a0e721cff33.png)

For the "1" in "1 of 24", make a dropdown that would switch to the numbered chapter on change.

Issue #1875, opened 2023-01-17
Remove the print preview button for now from the reader mode nav bar


lizadaly:
The print preview button was just kind of hung here to put it somewhere. It's out of place, so remove it. 

(Possibly it doesn't need to go anywhere; ultimately this is what will come out of the "generate PDF" workflow and probably doesn't need to be navigated to by a human at all.)

<img width="990" alt="image" src="https://user-images.githubusercontent.com/19571/212977966-eb0e5e1d-de1a-4426-b74b-bbf3d9e4506e.png">

Issue #1874, opened 2023-01-17
Change style of TOC shortcut link in reader view 


lizadaly:
The current presentation makes it looks like a button that will pop out the TOC in place. The TOC will be very long and wide, so treating it like a modal doesn't make sense.

Change the style to look more like a "back to top" link so the behavior (scrolling to the top of the page) is less unexpected.

PR #1883, opened 2023-01-19
Reading mode footer improvements

lizadaly:
Addresses #1874, #1879, #1875

Changes to the footer of reading view to streamline the experience:

image

* Removed print preview button (the entry point for this is going to be the PDF generation flow)
* Changed offset mechanism for the "Table of contents" link—previously it would get tucked behind the nav bar on short viewports. Now it's a fixed offset from the bottom.
* Moved the "Exit reading mode" link out of the nav bar and positioned it parallel to the TOC link
* Added a non-enabled greyed-out style for next/previous buttons when no subsequent chapter is available.
* Added forward/backward triangles to the next/previous buttons when active, to better suggest their purpose.
* The "1 of n sections" indicator is now a dropdown; changing it will navigate the user to the appropriate section.

Added a Playwright test for the dropdown changer.

Note that in Safari, there's a very aggressive shiny style applied to dropdowns:

image

I _can_ get rid of this, but doing so also removes the up/down arrows that hint that it's a dropdown, and is not generally recommended.

cc @cath9


lizadaly:
(looking into the test failure, which didn't happen locally)

bensteinberg (approved):

PR #1884, opened 2023-01-19
Drop animation in toc expansion in reading mode

### LINKED ISSUE #1880: Height calculation for TOC not working in reading mode on first load
Something's funny here; only happens occasionally so maybe a problem with timing:

image

on staging: casebooks/4243-administrative-law-372

lizadaly:
Fixes #1880 by removing the feature

This gives up on an accordion animation for the TOC when opening or closing it in reading mode. Dynamically expanding a variable-height div is very tricky, especially if the div needs to sometimes initially render as closed, as here when not visiting the first chapter. 

I tried a lot of different approaches but this is a well-trod tricky problem that even modern CSS can't solve well (TL;DR you would like to be able to animate `auto` properties; [current still-active spec thread](https://github.com/w3c/csswg-drafts/issues/626).)

Since casebook chapters can be _very_ long, I think it's especially problematic here to do any of the height/JS-based workarounds, because each transition involves a repaint. On mobile devices I think it would have been unworkably slow even without the buggy behavior that generated the related ticket.

Now when clicking the open/close toggle, the TOC is simply shown or hidden in a single tick.

### DIFF
--- a/web/frontend/pages/as_printable_html.js
+++ b/web/frontend/pages/as_printable_html.js
@@ -15,39 +15,10 @@ class TocControl extends HTMLElement {
     this.tocList = this.querySelector('ol');
     if (this.tocList) {
       this.querySelector('.toc-opener').addEventListener('click', () => {
-        window.requestAnimationFrame(() => this.toggleAttribute('open'));
+        this.toggleAttribute('open');
       })
-      this.tocList.addEventListener('transitionstart', () => {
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
👍

rebeccacremona (approved):
> (TL;DR you would like to be able to animate auto properties

Huh, didn't know about this, pretty interesting

PR #1887, opened 2023-01-25
Fix elisions in reading mode by moving DOM modifications later

lizadaly:
I thought this was complicated but it was simple, and it is because I did something silly.

## Before

Note the elision placement at the end of the paragraph (this is correct):

image

Reader view was placing the elision off into the middle of the next paragraph:

image

## After

image

Offset calculations for annotations in reading mode are represented as [Ranges](https://developer.mozilla.org/en-US/docs/Web/API/Range), which can be computed before a document tree is inserted into the DOM. Doing it this way means I don't have to account for newly-inserted notes changing the offsets, or for repaint events. Only when all the ranges are calculated in memory does the code walk through and start inserting or eliding nodes.

HOWEVER, at some point I added code that inserted footnote nodes for use in the print output, and I put that before the range calculation instead of afterward, and then missed the regression because there weren't yet tests. I'll add tests for this soon. In this fix I moved it after the range calcs and also inside a block that applies only to PagedJS because it's unnecessary work in reading mode.

Also:

* Fixes an off-by-one error that was leading to elisions spilling over one character into the next node. This was usually harmless but meant that the elision marker `[...]` was sometimes falling inside the wrong paragraph, resulting in a misleading suggestion that text had been cut from that paragraph.
* Changed the `id` values I added in a previous PR to data attributes, because these values will sometimes duplicate when elisions or highlights span elements.
* The code was previously adding an elision marker for each node that contained an elision; now it only adds one when the elision is fully closed. `<p>This <i>should</i> be one elision</p>` was resulting in `<p>[...] <i>[...]</i> [...]</p>`; now this will print `<p>[...]</p>` (visually; the markup is more complicated).
* In cases where an elision fell inside markup that resulted in italics, the elision will now not be in italics (this case was common).
* In cases where the elision spans an entire paragraph, the elision would inherit the indent of the paragraph. Now it will be always fully out-dented.


matteocargnelutti (approved):
NICE

PR #1892, opened 2023-01-31
Add an endpoint that will generate a PDF; part of #1801

lizadaly:
Working sketch of generating a PDF from a user endpoint, with the following limitations:

* The export will fail for non-public casebooks because the Playwright browser can't access privileged resources yet.
* Any output from Playwright including errors aren't yet reportable to the front end.
* The user-facing endpoint will always just return "OK" immediately.
* The PDF is written to the temp filesystem on the host, which isn't accessible to the end user.
* Large 500+ page PDF generation fails because PagedJS doesn't think it's completed layout yet. Still to be investigated (this was a known issue).

If you copy the generated PDF off the Docker container, you can see that it works:

image

There's a button on the frontend to trigger the endpoint, only visible to admins:

image

I switched the label to "Beta" from "admin only" in preparation for making some of this eventually available to end users, and changed "HTML preview" to "Reading mode", a name which I also don't like.

## Next steps

I think this could be merged once approved. I'm curious to see how error reporting ends up working Sentry for the task— requesting a non-public casebook should be an easy way to trigger an error.


matteocargnelutti (approved):
Awesome 😎

Issue #1891, opened 2023-01-27
Poll or wait for completion of async PDF generation job


lizadaly:
Generating a PDF title will trigger the following workflow:

* An HTTP request to the "generate PDF" endpoint 
* A celery job will be queued to start the Playwright (#1801) 

The job will either write a completed PDF, error, or time out.

If the job succeeds: 

* A PDF will be available on block storage.
* A URL to the PDF should be delivered to the end user that they have permission to access.

If the job errors: 

* The user should be notified and the error should be propagated.

If the job times out:

* The user should be notified and the timeout should be raised as an error.

lizadaly:
There are lots of implementation options for polling for job completion. What do the other projects do?

rebeccacremona:
> There are lots of implementation options for polling for job completion. What do the other projects do?

Perma tracks every request to make a Perma Link in a [CaptureJob](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1835) table. As the job runs, we update its [status](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1842) field, along with some [progress indicators](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1855). That info is exposed in an [API route](https://github.com/harvard-lil/perma/blob/develop/perma_web/api/views.py#L259). The frontend's JS hits that route [every so often](https://github.com/harvard-lil/perma/blob/develop/perma_web/static/js/create-link.module.js#L70) and advances a progress bar; we've heard some feedback that the progress bar isn't the best design choice. 

The [prototype capture service](https://github.com/harvard-lil/perma-capture/tree/new-arch-with-vue) is set up the same way, but is tidier, including a [Vue frontend](https://github.com/harvard-lil/perma-capture/tree/new-arch-with-vue/web/frontend).

As far as I know, CAP doesn't have any user-facing tasks or polling; probably worth confirming that with someone else.

lizadaly:
This is done!

PR #1898, opened 2023-02-06
Monitor PDF export result status

lizadaly:
Part of #1801; does most of #1891

Adds job monitoring for PDF creation and a front-end component to alert the user when it is complete. This is a larger-than-usual PR because all the parts are needed to do something useful.

Includes:

* Adds [django-celery-results](https://django-celery-results.readthedocs.io/), the Celery-endorsed method for monitoring job statuses using the Django ORM. 
* Adds a new Vue component, `ExportButton` which triggers and monitors the PDF export process.
* Replaces the synchronous endpoint URL with an API endpoint that accepts POST requests (to create a new export task) and GET (to check the status of a task)

`django-celery-results` is generic and doesn't do everything you might want in a task monitoring table, but it seems like a good start and it's nice that it provides immediate visibility in the Django admin for any celery job:

image

When a PDF job is triggered, a simple polling mechanism starts in Vue to check the status. The absence of any task result is assumed to be a pending job; when a result is present it'll be in either the SUCCESS or FAILURE state.

This gets reported back to the user as one of three cases:

* The PDF export succeeded, and it prints the path to the temp filesystem where it was generated (this isn't useful yet).
* The PDF export failed because Playwright threw an exception. Usually this is because the casebook isn't public and Playwright can't reach it.
* The polling timed out. This could be any reason, like the Celery task failing to start, or because (as of right now) the polling timeout is shorter than the Playwright timeout. Playwright will time out on long casebooks, but long after the front end has given up on waiting.

## Success

https://user-images.githubusercontent.com/19571/217097270-95a49e22-331c-413f-9b5c-4d10c08d472f.mov

## Failure (non-public casebook)

https://user-images.githubusercontent.com/19571/217097424-d51dced0-5468-4212-bbc0-3618389f6e1e.mov

## Tests

None specifically yet—in large part because the full set of authorization decisions haven't been made—for example, will this be available to anyone, even bots, like the Word export? 

There are no permission checks on the task_id GET endpoint at the moment; you need to know the `task_id` to query for anything, and nothing can be done with the output. We need to think through what _should_ be true for this endpoint, though.

The POST endpoint has the usual "user must be able to view casebook" check.

### DIFF
--- a/web/config/settings/settings_base.py
+++ b/web/config/settings/settings_base.py
@@ -31,6 +31,7 @@ INSTALLED_APPS = [
     "webpack_loader",
     "django_json_widget",
     "simple_history",
+    "django_celery_results",
     # built-in
     "django.contrib.admin.apps.SimpleAdminConfig",
     "django.contrib.auth",
@@ -262,6 +263,9 @@ CELERY_TASK_ROUTES = {
 }
 
 CELERY_BEAT_SCHEDULE = {}
+CELERY_RESULT_BACKEND = "django-db"
+CELERY_CACHE_BACKEND = "django-cache"
+CELERY_RESULT_EXTENDED = True  # Include full metadata in TaskResult entries
 
 ###
 ### /END CELERY settings ###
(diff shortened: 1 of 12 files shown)


bensteinberg (approved):
LGTM

PR #1899, opened 2023-02-07
Add Minio locally; write PDF output to "bucket"

lizadaly:
* Adds a new bucket to the minio filesystem
* Writes the completed PDF to it
* Return a URL to the object in the bucket
* Updates tests to match expectation
* Updates the UI to give the user a download link

I set an expiration time on the URL as 60 minutes, which happens to be the default anyway. Minio is returning content type `application/pdf`, which is correct; we'll see if S3 does the same thing.

image

Considerations for actual S3:

* The bucket needs to exist
* The workers will need to be able to access the bucket (recommend an ACL like bucket READ/WRITE and no specific object permissions)
* Recommend setting up a lifecycle policy on objects in the bucket (they should probably start by expiring in 24 hours, which is likely to be far longer than the URL will be active)

Until S3 is set up, merging this will break the PDF functionality on stage/prod. I think that's OK since it's currently not useful.


bensteinberg (approved):
Looks good -- I'll set up S3 bucket etc. before we merge -- stay tuned, and/or I may have questions.

web/main/tasks.py
```diff
@@ -43,10 +44,12 @@ def generate_pdf(
     page.on("console", lambda msg: logger.warning(f"From browser console: {msg}"))
     expect(page.locator(selector)).to_be_visible(timeout=timeout)
     pdf = page.pdf()
-    output_file.write_bytes(pdf)
-
+    storage = get_s3_storage(bucket_name=os.environ["PDF_EXPORT_BUCKET"])
```
rebeccacremona: Suggestion: for consistency, can you make this a Django setting that is set to `os.environ["PDF_EXPORT_BUCKET"]` by default? That will make it easier for us to override in tier-level settings.py

rebeccacremona:
One more suggestion: IIRC, `settings_common` needs to have [`AWS_S3_SIGNATURE_VERSION = 's3v4'`](https://github.com/harvard-lil/perma/blob/081939e1afc39080a48774b18e29acac47547f38/perma_web/perma/settings/deployments/settings_common.py#L67); otherwise, we get an old-style signed URL instead of the more desirable modern version.

Old format: `http://perma.minio.test:9000/perma-storage/generated/warcs/SJ/7C/ST/SJ7C-STN5.warc.gz?AWSAccessKeyId=accesskey&Signature=z9UAtQ%2B5hj2KBkZnc8iTedy3Vuo%3D&Expires=1651681642`

New format: `http://perma.minio.test:9000/perma-storage/generated/warcs/SJ/7C/ST/SJ7C-STN5.warc.gz?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=accesskey%2F20220504%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20220504T171155Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=1e08b1bdabfe9f4a59ad98e1e867092e4c3309fbda43683f599f1018a40f3b96`

lizadaly:
> One more suggestion: IIRC, `settings_common` needs to have [`AWS_S3_SIGNATURE_VERSION = 's3v4'`]

Confirmed now getting URLs like:

http://opencasebook.minio.test:9000/h2o.pdf_exports/this-is-a-test-20230207T185658.pdf?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=accesskey%2F20230207%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20230207T185700Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=5d47d237370655e856ba8c01527c909cb5bde4128b1672389a2f9128f68845ca

bensteinberg:
I think I have this set up, so we can try it out anytime.

Issue #1651, opened 2022-07-21
Add suggestions for institution


dongkyunc:
`SELECT affiliation FROM main_user WHERE verified_professor=true;` returns 141 different values for 228 verified professors. Some of the affiliations are 

```
Harvard University
Harvard Law School
Harvard Law School, Berkman Center
HLS/HBS
hls
HLS
Berkman Klein Center for Internet & Society
Berkman Klein Center

Professor, George Washington University

None

[redacted professor1's cv]
[redacted professor2's website]

Magistrate Judge

Financial Regulation: Law and Policy
```

This makes grouping by university kind of hard (unless we want to do it via user email domain?). Providing options assuages the issue somewhat and it seems to be possible from either the [Django](https://django-autocomplete-light.readthedocs.io) end or from [Vue](https://github.com/darrenjennings/vue-autosuggest). I tried wrangling the form for a couple hours but got nowhere.

lizadaly:
@cath9 As a first step, what do you think about cleaning these up manually? We could export these values as a CSV and then make changes in the admin.

cath9:
yep I think that's the only way. Need to think a bit about how to communicate these changed to profs who have already been using the platform. but we should move forward!

lizadaly:
This is fixed; the list of schools in search is now reasonable.

PR #1907, opened 2023-02-14
Switch site search to use Institution to populate school dropdown

lizadaly:
Currently, the site search uses the user-supplied "affiliation" field to drive the "School" dropdown used for faceting casebook results. There are a few issues with this:

1. The results list includes many variations on the same university name, see #1651
2. It's freeform text, so not every value here is really a "school"
3. The index is built at the casebook level; the indexing code uses `string_agg` to aggregate multiple affiliation fields into one comma-delimited field. This in turn introduces more duplicates based on whichever order was picked at indexing time.

A sample of what the dropdown contains now:
image

It's also the case that selecting from the dropdown will only **exactly** match the set of affiliations associated with that casebook. For example, [filtering by "Columbia Law School"](https://opencasebook.org/search/?type=casebook&q=&sort=display_name&author=&school=Harvard%20University,%20Columbia%20Law%20School) returns only two results, when a [third result](https://opencasebook.org/search/?type=casebook&q=&sort=display_name&author=&school=Columbia%20Law%20School,%20Harvard%20University) exists where a Columbia Law School professor was a co-author.

This PR:

* Switches the indexed field to be from the controlled vocabulary of the `Institution` model. @cath9 manually populated all the verified professors with this value awhile back.
* In the full-text index, populates the institution field as an array rather than a string, one entry for each represented institution.
* On the front end, populates the school dropdown with unique values from the set of institutions in the search index.

**My local institution data isn't reflective of prod and is made up**; before merging we should update the staging site DB so it incorporates Catherine's data work, and then really test this before going live.

## New dropdown

Dropdown is coming from the Institution table:

image

Filtering on one institution will also show other institutions from that result set:

image

Of the 4 results here, 3 are from coauthors and one is solo.

If we switch to Santa Clara, we get 3 of the same results, but we're missing the 4th, because that's CUNY only:

image

### DIFF
--- a/web/main/models.py
+++ b/web/main/models.py
@@ -41,7 +41,6 @@ from django.core.paginator import Paginator, Page
 from django.core.validators import MaxLengthValidator, validate_unicode_slug
 from django.db import ProgrammingError, connection, models, transaction
 from django.db.models import Count, F, JSONField, Q, QuerySet
-from django.db.models.expressions import RawSQL
 
 from django.template.defaultfilters import truncatechars
 from django.template.loader import render_to_string
@@ -886,7 +885,7 @@ class SearchIndex(models.Model):
             if k == "institution":
                 # Institutions are arrays, but the ORM won't know that by default.
                 # It does know that it's a JSONB blob, and a `contains` query does the right thing.
-                base_query = base_query.filter(**{f"metadata__institution__contains": v})
+                base_query = base_query.filter(**{"metadata__institution__contains": v})
             else:
                 base_query = base_query.filter(**{f"metadata__{k}": v})
 
--- a/web/main/models.py
+++ b/web/main/models.py
@@ -19,7 +19,6 @@ from typing import (  # noqa: F401 workaround for django-stubs#1022 until the fi
     Union,
 )
 from urllib.parse import urlparse
-
 import lxml.etree
 import lxml.sax
 import requests
@@ -38,10 +37,12 @@ from django.contrib.postgres.search import (
     SearchVectorField,
 )
 from django.core.exceptions import ValidationError
-from django.core.paginator import Paginator
+from django.core.paginator import Paginator, Page
 from django.core.validators import MaxLengthValidator, validate_unicode_slug
 from django.db import ProgrammingError, connection, models, transaction
 from django.db.models import Count, F, JSONField, Q, QuerySet
+from django.db.models.expressions import RawSQL
+
 from django.template.defaultfilters import truncatechars
 from django.template.loader import render_to_string
 from django.urls import reverse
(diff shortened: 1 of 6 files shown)


web/main/views.py
```diff
@@ -3076,19 +3076,22 @@ def internal_search(request):
...
         page=page,
         query=query,
         filters=filters,
-        facet_fields=["attribution", "affiliation"],
+        facet_fields=["attribution", "institution"],
         order_by=request.GET.get("sort"),
     )
     full_counts = SearchIndex.counts(query=SearchIndex.objects.all())
 
-    results.from_capapi = False
```
lizadaly, line 3091: Enabling mypy on these functions flagged that this property wasn't on the object, and indeed I couldn't find the string in the code anywhere.

web/main/models.py
```diff
@@ -858,19 +849,8 @@ def refresh_search_index(cls):
         with connection.cursor() as cursor:
             try:
                 cursor.execute("REFRESH MATERIALIZED VIEW CONCURRENTLY internal_search_view")
-            except ProgrammingError as e:
```
lizadaly, line 861: This doesn't actually work—for one thing the view name is `internal_search_view`, so the clause is never caught. If that's corrected, it _still_ doesn't work because the database is mid-transaction and can't commit the new indexing step. So I just removed this wrapped and put the "private" method at the top.

web/main/models.py
```diff
@@ -858,19 +849,8 @@ def refresh_search_index(cls):
...
-                if e.args[0].startswith('relation "internal_search_view" does not exist'):
-                    cls.create_search_index()
-
-    @classmethod
-    def search(cls, *args, **kwargs):
-        try:
-            return cls._search(*args, **kwargs)
-        except ProgrammingError as e:
-            if e.args[0].startswith('relation "internalsearch_view" does not exist'):
```
lizadaly, line 870: This doesn't actually work—for one thing the view name is internal_search_view, so the clause is never caught. If that's corrected, it still doesn't work because the database is mid-transaction and can't commit the new indexing step. So I just removed this wrapper and put the "private" method at the top.

bensteinberg (approved):
Looks good. I will update the db on stage. Stay tuned.

web/main/views.py
rebeccacremona: Ah, I bet that is a relic from before we had `LegalDocument` and `LegalDocumentSource` classes.

rebeccacremona (approved):

web/main/views.py
rebeccacremona: yeah https://github.com/harvard-lil/h2o/blob/315c8d4c7b6e314ec7ced886daca8cce33135821/web/search/views.py#L51

PR #1910, opened 2023-02-17
Move printable layout to static handling; move HTML out of template

lizadaly:
This makes a number of simplifying and performance-enhancing changes to reading mode. It also firmly breaks reading mode away from PagedJS/PDF rendering in favor of choices that benefit screen reading only.

This does not truly break PDF output—you can still get a file from the export button—but that file looks very primitive now. I'm going to add back in PDF-centric styles in further work.

This PR:

* Puts the book content directly into the browser-viewable DOM, rather than in a `template` element. The former step was an optimization designed to inject annotations synchronously before PDF rendering; now the content goes straight to the browser and the annotations are added more opportunistically.
* Remove the bundled common resources from reader view; this was a big payload that was unused.
* Makes internal IDs from case content unique across the document so internal hyperlinks continue to work.

And lots of small-ish performance optimizations:

* Font preloading
* Add annotations using `insertAdjacentHTML` rather than `insertAdjacentElement` (Mozilla says it's faster and my benchmarks agreed)
* Uses `requestIdleCallback` for a few mutations that don't visibly affect layout 

# Lighthouse metrics before/after 

## Before

image
image

## After
image
image

(The absolute numbers vary a lot between casebooks so this is more about the relative difference.)


lizadaly:
Looking into that failing test; it's passing locally.

bensteinberg (approved):
🚤

web/static/as_printable_html/as_printable_html.js
```diff
@@ -208,64 +188,59 @@ annotationRanges.forEach((rg) => {
...
       });
-      break
+      break;
   }
 });
 
 // Any existing URLs that are external and don't open in a new tab should
-tmpl.content.querySelectorAll('a[href^="http"]:not([target])').forEach((el) => {
-  el.setAttribute('target', '_blank');
+main.querySelectorAll('a[href^="http"]:not([target])').forEach((el) => {
```
matteocargnelutti, line 214: Potential trade-off to be made here: If you're solely focusing on time to interactive, a broad event delegation filtering clicks on `a` could work.

matteocargnelutti:
This is excellent, @lizadaly .

> Add annotations using `insertAdjacentHTML` rather than `insertAdjacentElement` (Mozilla says it's faster and my benchmarks agreed)

I would have assumed the opposite!

web/static/as_printable_html/as_printable_html.js
lizadaly: Oh that's a really good point, especially since most of these will not be clicked and there can be hundreds.

PR #1912, opened 2023-02-21
Miscellaneous package updates

lizadaly:
What it says on the tin.


lizadaly:
Whoops, looks like the compatible django-stubs version will come out soon, so will wait on these updates.

PR #1915, opened 2023-02-22
More reading mode updates

lizadaly:
* Fixes a bug where replacement text was being injected for each range composing a replacement (for large selections this resulted in duplicated replacements)
* Handles more deeply nested headers
* Makes link icons more predictably sized
* Better resets for some elements which can inherit styles from parents


matteocargnelutti (approved):
LGTM. Tests are still running but I don't expect any of these changes to affect them.

PR #1916, opened 2023-02-24
Replace Fabric with Invoke

### LINKED ISSUE #1810: Consider replacing local task runner with `invoke`
Local tasks like running builds or performing maintenance use the `fabric3` library, which works well and developers are comfortable with. There are some reasons to consider switching tooling here though:

* `fabric3` is at end-of-life, because the parent [Fabric](https://www.fabfile.org/index.html) library is now under active development again.
* Fabric is primarily a remote code execution tool, but deployment is now handled in other ways. Requiring Fabric brings along a lot of heavyweight dependencies like [paramiko](https://www.paramiko.org/) that require version updates despite being unused.

[Invoke](https://www.pyinvoke.org/) is close to a drop-in replacement for Fabric when used for only local tasks. It has no dependencies.

(It might be convenient or it might be confusing but it would be possibly to retain `fab` as an alias so people can continue to use the commands they're accustomed to during development!)

bensteinberg:
Since we only use Fabric to run things locally, we can switch to Invoke, removing a few dependencies. Closes #1810.

Notable changes for developers who use Invoke tasks include replacing underscores with hyphens: `invoke run-frontend` instead of `fab run_frontend` -- and boolean arguments are really boolean now: `invoke run --debug-toolbar` instead of `fab run:debug_toolbar=True`

When writing new Invoke tasks, the two things to remember are including the context as the first argument, and that arguments are not necessarily strings, as they are in Fabric3 -- see https://docs.pyinvoke.org/en/stable/concepts/invoking-tasks.html#type-casting

See also https://github.com/harvard-lil/perma-payments/pull/165 -- there's some discussion there about whether and how to change the names of files in which Invoke and Celery look for tasks. The short version is that we are going to let the less-configurable Invoke have `tasks.py`, and I've chosen to give Celery `celery_tasks.py`.

This PR also updates the `@sentry/tracing` JS package, and removes a PDF file generated in testing that snuck in a few months ago.


lizadaly:
It looks like this'll be a nice improvement (I especially like that it can use real command-line args).

The celery change wasn't happy though:

```
[2023-02-24 21:16:24,144: ERROR/MainProcess] Received unregistered task of type 'main.tasks.demo_scheduled_task'.
The message has been ignored and discarded.

Did you remember to import the module containing this task?
Or maybe you're using relative imports?

Please see
http://docs.celeryq.org/en/latest/internals/protocol.html
for more information.

The full contents of the message body was:
b'[[], {}, {"callbacks": null, "errbacks": null, "chain": null, "chord": null}]' (77b)

Thw full contents of the message headers:
{'lang': 'py', 'task': 'main.tasks.demo_scheduled_task', 'id': '5d8e8cfe-c5dc-44c2-931f-e0d4a3547382', 'shadow': None, 'eta': None, 'expires': None, 'group': None, 'group_index': None, 'retries': 0, 'timelimit': [None, None], 'root_id': '5d8e8cfe-c5dc-44c2-931f-e0d4a3547382', 'parent_id': None, 'argsrepr': '()', 'kwargsrepr': '{}', 'origin': 'gen33@1454fbf2a403', 'ignore_result': False}

The delivery info for this task is:
{'exchange': '', 'routing_key': 'background'}
  File "/usr/local/lib/python3.9/site-packages/celery/worker/consumer/consumer.py", line 591, in on_task_received
    strategy = strategies[type_]
KeyError: 'main.tasks.demo_scheduled_task'
```

Maybe a simple solution would be to set an alias in the Docker user's `.bashrc`?

e.g.:
```bash
alias invoke="invoke -c invoke_tasks"
```

and then leave `tasks.py` to celery? Alternatively update any references to `main.tasks`.

lizadaly:
Looks like you fixed it while I was still babbling!

lizadaly:
```
inv run
```

This works, but other tasks aren't found:

```
root@1454fbf2a403:/app/web# inv run_frontend
No idea what 'run_frontend' is!

root@1454fbf2a403:/app/web# inv refresh_search_index
No idea what 'refresh_search_index' is!
```

On first glance I can't see why.

lizadaly:
I'm an idiot!

lizadaly (approved):
This looks good, sorry for the noise!

I assume the deploy/cron jobs will update too?

bensteinberg:
Yeah, I will update deployment config before I merge this.

PR #1920, opened 2023-02-28
Refactor docx html export code to its own module

lizadaly:
# Background 

For PDF export of very long books, handling annotations in the browser is prohibitively time-expensive. In earlier work I optimized this for screen display by deferring expensive transformations, but in the PDF export, all transformations need to be done up-front before the document is produced.

Luckily, we already have a server-side mechanism for introducing annotations that's known to work—the transformations that are called by the docx export pipeline that produces HTML for consumption by pandoc.

My current plan is to use the browser-friendly and accessible markup only for reading mode, and to reuse the markup generated for docx export for PDF export. I'll use PagedJS + Playwright for this because that's known to work, even if it isn't extremely fast. (I think the absolute best option for headless PDF generation in 2023 is probably https://weasyprint.org/, which unlike pandoc's PDF export is self-contained, and unlike most other open source PDF tools is still in active development.)

# This actual PR

In order to support footnotes in PagedJS, the export flow will have to take an option to insert notes and links adjacent to their source rather than at the bottom of the document. But as written this export code is a little hard to work with—it's a method on a model containing an inner class containing long methods, with some complicated docstrings for testing.

Before I make any changes to it I wanted to have it in a state that was a little easier to work with, so this moves the code out to its own module and and refactors the assertions to individual pytests with smaller scopes more appropriate for unit tests. I also added type annotations to much of the code, and wrote a template tag to call the new function since it's no longer a callable method. Actual test coverage and functionality should be identical, and I verified it with a few large casebooks. I'll also test on staging once merged.

### DIFF
--- a/web/main/export.py
+++ b/web/main/export.py
@@ -1,0 +1,282 @@
+import re
+from typing import Any
+
+from django.utils.safestring import SafeText, mark_safe
+from lxml import html, sax
+from pyquery import PyQuery
+
+from main.models import ContentAnnotation, ContentNode
+
+from .utils import block_level_elements, remove_empty_tags, void_elements
+
+SortedAnnotation = tuple[int, bool, ContentAnnotation]
+ 
+
+class AnnotationContentHandler(sax.ContentHandler):
+    def __init__(self, annotations: list[SortedAnnotation], postfix_id: int, export_options: dict):
+        # internal state:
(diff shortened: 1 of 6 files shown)


rebeccacremona (approved):
Huzzah!

I particularly like the parameterized pytests; nice and clean!

Issue #1540, opened 2022-03-15
Improve Django admin performance for specific model pages


rebeccacremona:
Needs re-thinking, subsequent to the decision to remove proxy models. Not performant at present: for instance, attempts to load the resource list 504.

lizadaly:
Pages that work for me locally, if slowly, but do time out on prod:

* Resources
* Content annotations
* Content nodes

rebeccacremona:
I suspect something like this will might come in handy https://github.com/harvard-lil/perma/pull/2926/commits/a2e02f69c7bf017c4d8809a8e0c236ee49a41c16

PR #1924, opened 2023-03-01
Drop expensive queries from contentnode admin

lizadaly:
This should vastly improve the speed of the ContentNode admin page such that it becomes useful—at the moment the list page just times out on prod. 

Removes two very expensive queries:

* Drops the count of annotations per content node for each row of the list page. This is very very slow and not that interesting anyway.
* Adds the estimated paginator taken from Perma for the result set.

Also adds two fields on the change page that were probably omissions: `resource_id` and `resource_type`. They were listed in `readonly_fields`, but those values have no effect if they aren't also in `fields`. (This behavior of Django's is silly.)

### DIFF
--- a/web/main/admin.py
+++ b/web/main/admin.py
@@ -4,20 +4,22 @@ from django.conf import settings
 from django.contrib import admin, messages
 from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
 from django.core.mail import send_mail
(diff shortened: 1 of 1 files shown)


bensteinberg (approved):
:fire:

lizadaly:
c.f. https://github.com/harvard-lil/h2o/issues/1540

PR #1929, opened 2023-03-13
Handle long URL strings in annotations with scrolling and generated hyperlinks in reading mode; fixes #1926

### LINKED ISSUE #1926: Move reading mode to public beta
After some internal and external testing, we're nearly ready to make reading mode available to all users.

* Is this still labeled "beta"?
* Do we want any help content available? 
* Any other required features or fixes before this is applied?

Rolling back will be a trivial change to restore the admin-only check, so this is low risk overall.

lizadaly:
In both the regular and reading mode UIs, long hyperlinks can break out of their containers.

Ideally authors would have creates these as the hyperlink type, but currently in the platform adding rich/HTML content in author notes is not supported; they are always plain text.

In both the platform UI and reading mode, this will force scrolling inside the annotation if the browser could not line-break it.

In reading mode only, this will also transform plain text URLs into hyperlinked ones. This is harder to do in Vue because Vue enforces not modifying the DOM in templates. Since this case is rare overall, this only does the DOM transformation in reading mode.

## Before

Platform UI—the first link does break because the browser can wrap on the hyphen:

image

## After

Platform UI—the bottom one will scroll:

image

Reading mode—the bottom scrolls and the links are clickable:

image

### DIFF
--- a/web/frontend/components/NoteAnnotation.vue
+++ b/web/frontend/components/NoteAnnotation.vue
@@ -145,7 +145,7 @@ a:active {
     padding: 10px 15px;
     width: fit-content;
   }
-
+ 
   .button {
     margin-top: 1em;
   }
@@ -166,6 +166,7 @@ a:active {
   padding: 10px;
   background-color: $white;
   color: $black;
+  overflow: scroll; 
 }
 .top {
   z-index: 10;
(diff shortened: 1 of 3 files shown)


matteocargnelutti (approved):

rebeccacremona:
> if the browser could not line-break it

I am really amused that we have never spotted this in H2O before!!

Love the scrolling, but just throwing it out there: this might be a good time for `word-wrap: break-word`... if we suppose that long words like "supercalifragilisticexpialidocious" will be rare in notes...

before
image

after
image

PR #1931, opened 2023-03-15
Hide/show reading mode TOC

lizadaly:
Couple features/extensions to reading mode:

* Show/hide the TOC based on a button. This doesn't animate, though it could if we wanted.
* Fixes a bug with the links in the TOC, which @cath9 thinks "should actually work." Ugh fine now they do.
* Is more explicit about the desired behavior of scrollbars. Because this layout is familiar to our users from Google Docs and elsewhere, it hides the intra-div scrollbars on browsers that might otherwise show them. We can change this behavior if users find it confusing.
* Is more explicit about how the dividing line between the sections is drawn. Before it was something of a side effect that the active TOC area was not offset with a line. Thanks @matteocargnelutti for looking at that with me.

## TOC open

image

## TOC closed
image

### DIFF
--- a/web/main/templates/export/as_printable_html/casebook.html
+++ b/web/main/templates/export/as_printable_html/casebook.html
@@ -72,9 +72,7 @@
     </nav>
     <div>
       <header class="site-header">
-        
-    
-          <p >
+          <p>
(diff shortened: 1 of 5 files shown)


bensteinberg (approved):
👍

PR #1932, opened 2023-03-17
Reading mode: reflow on small devices; move nav to right

lizadaly:
The people have spoken so the nav moves to the right. This also changes the behavior of the "hide toc" button to match.

Following stuff I learned from "[Every Layout](https://every-layout.de/)" this relies on flex to bounce the nav to the bottom if there isn't enough room in the viewport for both columns.

I am still using media breakpoints for internal sizing changes to the content window in small viewports. (This would be an ideal place to use container queries but they may be just a bit too new to rely on for production software.)

Other changes:

* Adds the title heading "Table of contents" to the TOC mostly for screenreader announcements
* Swaps to the logo that's just the wordmark
* Fixes incorrect placement of author names in headnotes
* Moves the copyright notice to the nav and reduces the content
* Changes the "learn more" link to point directly to our copyright statement

## Desktop
image

## Mobile

Top and bottom of viewport

image image

### DIFF
--- a/web/main/templates/export/as_printable_html/casebook.html
+++ b/web/main/templates/export/as_printable_html/casebook.html
@@ -49,28 +49,7 @@
 
 
   <main>
-    <nav class="left">
-      <div class="metadata-block">
 
-          <a href="{% url 'casebook' casebook %}">
-            <img src="{% static 'images/ui/casebook/prev_page.svg' %}" alt="Back to site"  width="32">
-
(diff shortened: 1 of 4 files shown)


matteocargnelutti (approved):
🤝 _Democracy_ ✨

lizadaly:
Made a tiny additional change to fix some padding.

Issue #1934, opened 2023-03-20
ContentNode.resource_type should be indexed


lizadaly:
It's common to filter on this string field (which is effectively an enum of a known set) but it's not indexed, so this operation can be very slow when run across casebooks.

This comes up in the Django admin especially: any time `contentnode.resource_type` is referenced as something to filter on, Django will automatically want to do `select distinct resource_type from content_node` and then need to iterate over 1,001,118 rows to retrieve 5 or so known values. This particular case can be mitigated by a custom filter.

Note that like #1767, adding the index now will probably take quite some time and might need to be done carefully as part of a migration.

PR #1935, opened 2023-03-20
Remove unused admin pages and optimize slow-loading ones

lizadaly:
This should make several admin pages which time out on prod now load, and removes parts of the admin that were unused.

* Removes the admin pages that apply to proxy models—while these models aren't being removed imminently, their admin pages don't perform well and aren't strictly necessary because you can just use the resource type filters.
* Works around #1934 by statically providing the list of resource types to be filtered, specific to the model admin pages.
* Removes the queryset override in the contentnode admin; it was prefetching tables we don't display in the list view (anymore?) 
* Uses fast, estimation-style pagination and removes full counts for large admin pages where that value isn't useful because it's usually in the millions of rows.

(If the query is filtered in any way,  it will typically do full counts because I think this is otherwise fast enough now—it will only do estimated pagination for the unfiltered list view. 

**Edited**: the annotations model looked like it would time out on prod in this case, so I've provided a mechanism to list models that should always be estimated.)

Performance counts on my laptop (which is much faster than prod so this is more about the relative difference):

**ContentNode admin**

Before: 7 queries, 435ms (times out on prod)
After: 5 queries, 9ms

**ContentAnnotations admin**

Before: 7 queries, 350 ms (times out on prod)
After: 5 queries, 5ms

### DIFF
--- a/web/main/admin.py
+++ b/web/main/admin.py
@@ -29,12 +29,8 @@ from .models import (
     Institution,
     LegalDocument,
     LegalDocumentSource,
-    Link,
     LiveSettings,
-    Resource,
-    Section,
     Tag,
-    TextBlock,
     User,
 )
 from .utils import clone_model_instance, fix_after_rails
@@ -182,7 +178,7 @@ class InputFilter(admin.SimpleListFilter):
     https://hakibenita.com/how-to-add-a-text-filter-to-django-admin
     """
 
-    template = "admin/input_filter.html"
+    template = "admin/input_filter.html" 
 
     def lookups(self, request, model_admin):
         # Dummy, required to show the filter.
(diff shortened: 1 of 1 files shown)


web/main/admin.py
```diff
@@ -507,65 +495,9 @@ def save_model(self, request, obj, form, *args, **kwargs):
         ):
             obj.casebook.content_tree__repair()
 
+    ordering = ("id",)
```
lizadaly: The default ordering for this model is `ordinals`, which is an unindexed `ArrayField`. This was very slow when run across all content nodes (it's fine in the normal case when querying within in a single casebook, though ideally it would be some kind of an indexable value in all cases.)

web/main/admin.py
```diff
@@ -45,16 +41,18 @@
...
+
     @cached_property
     def count(self) -> int:
         cursor = connection.cursor()
         cursor.execute(
             f"SELECT reltuples AS estimate FROM pg_class WHERE relname = '{self.object_list.query.model._meta.db_table}';"
         )
         estimate = int(cursor.fetchone()[0])
-        if estimate > 10000:
+        if estimate > 10_000 and not bool(self.object_list.query.where):
```
rebeccacremona: Nice! If it turns out this works, I might see if we can do similarly in Perma...

rebeccacremona (approved):
Hooray for admin optimizations! It was nice and zippy back when the proxy models worked (and the ContentNode admin was not in use), but I think hasn't received a lot of attention since.

> Removes the queryset override in the contentnode admin; it was prefetching tables we don't display in the list view (anymore?)

Hmm, yeah, here's the [commit where the ContentNode admin was added](https://github.com/harvard-lil/h2o/commit/92deefca05fcf8d89b5cc72c8d64dd1daab9f860); looks like maybe those things were never displayed.

PR #1938, opened 2023-03-21
Reading mode to public: Add new blue logo; move reading mode link outside of the admin check

lizadaly:
This makes reading mode public! 🎉 

* Swaps the two-color logo for the one-color blue logo
* Removes the is-staff check around the reading mode button
* Removes (entirely) the PDF export button for now; the feature is still there, but this cleans up the view so admins see the same thing as regular users

This puts reading mode out totally public; it does not require authentication.

## Anonymous view:

image

## Logged-in view:

image

## Reading mode with new logo:

image


matteocargnelutti (approved):

bensteinberg (approved):
:100:

lizadaly:
Gonna give @cath9 a last chance to have feedback.

lizadaly:
n.b. we decided to keep the "beta" label for now not so much because it's unfinished but because it might cause people to actually notice it among the sea of other buttons there.

cath9:
This looks so good, let's do it

PR #1945, opened 2023-03-22
Remove exemption for superuser for legacy legal doc source

### LINKED ISSUE #1936: Remove role-based access to "legacy" source field
[This method](https://github.com/harvard-lil/h2o/blob/develop/web/main/models.py#L1112-L1118) causes the source values returned to be different for admin users from regular users. I found this confusing as I tried to down a JS error in the resource-add modal, when in fact it was only appearing for me as an admin user.

Simply removing the `is_superuser` check from this function unfortunately causes an unrelated, broadly-scoped doctest to fail unexpectedly:

```
___________________________________________________ [doctest] main.models.cleanse_html_field ____________________________________________________
[gw0] linux -- Python 3.9.9 /usr/local/bin/python
214     >>> assert node.headnote == same_after_sanitizing
215     >>> caplog.clear()
216 
217     Optionally, sanitize the field to remove potentially dangerous HTML before cleaning up whitespace:
218     >>> node.headnote = same_after_cleansing
219     >>> caplog.clear()
220     >>> with caplog.at_level(logging.DEBUG):
221     ...     cleanse_html_field(node, 'headnote', True)
222     >>> assert len(caplog.record_tuples) == 3
223     >>> assert caplog.record_tuples[0][2] == 'Normalizing newlines in class cl headnote'
UNEXPECTED EXCEPTION: AssertionError()
  File "/usr/local/lib/python3.9/doctest.py", line 1334, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest main.models.cleanse_html_field[19]>", line 1, in <module>
AssertionError
/app/web/main/models.py:223: UnexpectedException
```

The legacy source support for admin users can probably be removed at this point, and this test should probably be rewritten to only make assertions about the feature it's meant to test.

lizadaly:
Currently superusers will see three sources for legal documents: CAP, GPO, and "Legacy." The legacy type isn't exposed to non-admins already; this just puts admins on the same footing.

This will result in a small performance optimization for staff users doing searches—the front-end API is written such that it issues one distinct HTTP query for each source type, so are waiting for 3 HTTP responses to get results instead of 2 (one which is always empty of results). 

In the [related ticket](https://github.com/harvard-lil/h2o/issues/1936) I noted a test failure following this change, but I can't reproduce this now.

### DIFF
--- a/web/main/models.py
+++ b/web/main/models.py
@@ -1110,12 +1110,9 @@ class LegalDocumentSource(models.Model):
             cls.source_apis[api.details["name"]] = api
 
     @classmethod
-    def active_sources(cls, user: Union[AnonymousUser, User]) -> QuerySet[LegalDocumentSource]:
-        """Return the queryset of active sources based on the user's current permissions"""
-        search_sources = LegalDocumentSource.objects.order_by("priority")
-        if user.is_anonymous or not user.is_superuser:
-            search_sources = search_sources.filter(active=True)
-        return search_sources
+    def active_sources(cls) -> QuerySet[LegalDocumentSource]:
+        """Return the queryset of active sources"""
+        return LegalDocumentSource.objects.order_by("priority").filter(active=True)
 
     def api_model(self):
         # short_description, long_description, bulk_process, search(long_citation_json), import(id)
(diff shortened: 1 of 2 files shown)


kilbergr (approved):
Just to make sure I get it from your description/code:
this is changing active_sources method so that no matter what type of user you are, you are only getting active sources returned rather than superusers seeing both active and inactive sources.
If that's it, LGTM! If I've missed something, can you explain again? I'm not quite grasping how this PR would stop admins from receiving legacy type.

lizadaly:
That's it!

PR #1956, opened 2023-03-28
Add endpoint that will fetch and add a resource to a casebook in one roundtrip

lizadaly:
To optimize the add-legal-doc workflow, this adds a new API endpoint that performs the work of currently done by https://github.com/harvard-lil/h2o/blob/develop/web/main/views.py#L2591-L2630 and https://github.com/harvard-lil/h2o/blob/develop/web/main/views.py#L1915-L1939 as separate HTTP requests.

This endpoint simplifies and optimizes that flow in these ways:

* For legal docs that have previously been imported, it won't try to check the remote API for an updated version unless it's more than "max age" old. Per @cath9 we set this date at one year.
* It returns a JSON response rather than an HTTP redirect, including the new resource ID and the redirection URL so the client may redirect, but isn't required to
* Leverages the DRF automatic mechanism to transparently support JSON requests and responses rather than hardcoding form or JSON input

It includes the same logic as the dual endpoint it replaces, including accepting an optional section to add the new resource in a particular part of the casebook, or default to adding it at the end.

This doesn't add the frontend consumer for this endpoint, but does include a number of tests which should fully cover its functionality.

### DIFF
--- a/web/conftest.py
+++ b/web/conftest.py
@@ -327,7 +327,7 @@ class LegalDocumentFactory(factory.DjangoModelFactory):
     effective_date = datetime(1900, 1, 1)
     publication_date = datetime(1901, 1, 1)
     updated_date = datetime(1902, 1, 1)
-    source_ref = factory.Sequence(lambda n: {"id": n})
+    source_ref = factory.Sequence(lambda n: n)
     content = factory.Sequence(lambda n: f"Dubious legal claim {n}")
     metadata = {}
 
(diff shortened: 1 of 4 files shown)


web/conftest.py
```diff
@@ -327,7 +327,7 @@ class Meta:
     effective_date = datetime(1900, 1, 1)
     publication_date = datetime(1901, 1, 1)
     updated_date = datetime(1902, 1, 1)
-    source_ref = factory.Sequence(lambda n: {"id": n})
+    source_ref = factory.Sequence(lambda n: n)
```
lizadaly, line 330: This was just incorrect; the `source_ref` is a normal char field in the database:

```
 source_ref       | character varying(10000)   |           | not null | 
```

bensteinberg (approved):

Issue #1959, opened 2023-03-29
Links to GPO from search results don't work


lizadaly:
Not sure when this broke, but the API response from the backend doesn't return any values for GPO `urls`, so the front end has nothing to point to when the user tries to click the external links on the right:

<img width="938" alt="image" src="https://user-images.githubusercontent.com/19571/228647690-e7e6d98a-39a8-4ae2-a8fb-63c1f59db68a.png">

Example from CAP:

```
<a data-v-5412d700="" href="https://cite.case.law/so-2d/782/339/" target="_blank" rel="noopener noreferrer">CAP</a>
```

Example from GPO:

```
<a data-v-5412d700="" target="_blank" rel="noopener noreferrer">GPO</a>
```

The API response isn't adding a `url` value here:

```json
{"fullName": "Copies of Supplements to Code of Laws of United States and of District of Columbia Code and Supplements; conclusive evidence of original", 
"shortName": "Copies of Supplements to Code of Laws of United States and of District of Columbia C...", 
"fullCitations": "1 U.S.C. \u00a7 209", 
"shortCitations": "1 U.S.C. \u00a7 209", 
"effectiveDate": "2018-01-01", 
"url": null, 
"id": "USCODE-2018-title1-chap3-sec209"}
```

lizadaly:
Fixed by removing the anchor when no link is present; if there could be a link here, it needs to be mapped from the data that we have.

PR #1963, opened 2023-03-31
Rewrite frontend of legal document search + add; fixes #1933

### LINKED ISSUE #1933: Improve performance and responsiveness of "add content" button when selecting cases
When adding new case content using the Quick Add or Add Content feature, the user is prompted to do a search for a case, then select the matching case by clicking on it.

The behavior of this feature could be improved:

* Some UI hints are missing that the region is clickable
* There is no visible response from the UI after you click
* There is a significant delay between clicking and loading the case

The last two items together mean users often click repeatedly and overall the app looks buggy and not responsive.

lizadaly:
This is the frontend half of #1956 to complete #1933. 

There are two places in the code where users can add legal documents—from the "Add Content" button on the right, and from the "Quick Add" field in the table of contents. This is intended to replace both, ultimately, but this PR only includes the "Add Content" functionality.

Some of this code is duplicative of the existing legal doc flow. The final PR will make this work net-negative in terms of lines of code, but for now this is mostly additive.

This is a fairly big PR but the work was hard to roll out incrementally since it replaces a feature already there.

The UI looks very similar to before, with changes that are designed to aid readability and bring these results more in line with the presentation of the casebook search results.

## Before 
image

## After 
image

The most substantive UI change happens after a case is selected. Previously the page would do a full refresh and take them to the edit-resource page without any notification of what occurred (the legal doc was added to their casebook at the end or in the current section).

Now when they select a search result, they are notified about what happened and prompted to make a choice about what to do next:

image

## Full changelist

* The new code does not make use of the Vuex store. It does all its rendering and event bus management within the scope of the container, search form, and results components. The round-trip to the store seems to cause a lot of browser instability and may have been a source of the performance bottlenecks @cath9 was experiencing when demoing over Zoom. Since the search information is confined to these parent/child relationships, there's no need to use a store.
* This makes use of the new endpoint from #1956 to handle the addition of the legal doc in one HTTP request rather than 2 serial ones.
* When querying multiple sources (the default behavior), the existing code runs the queries serially. This runs them in parallel and waits for both to resolve (h/t @matteocargnelutti). 
* Supports full keyboard navigation, including result selection.
* For results without external links (see #1959), omits the hyperlink and external link icon.
* Avoids use of backwards-facing shims in favor of markup and APIs that are now solidly browser-native (a function of the time elapsed since this was first written). Specifically:
  * Uses `fetch()` rather than `Axios` (and since it uses a new API endpoint, it avoids any of the monkeypatching to handle Rails-compatible REST behavior)
  * Uses only the native `date` input rather than adding a shim
  * Uses flex rather than tabular layout for results
  * Includes only semantic markup (no divs!)
  * Uses `URLSearchParams` rather than a custom function

## TODO

There's explicitly no exception handling on the fetch requests right now. They should recover if the backend errors so the UI doesn't freeze, but still report up to Sentry. I'd like to research how best to do this.

There are unit tests for only one component. There should be some coverage for all three. (IMO this work is better suited for unit tests + mocks than Playwright tests, which would require a lot of backend fixtures when the backend is already covered by Python tests.)

I'd like to put this branch up on staging for some human testing before even considering merging, given the size of the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Run these from inside the container.
 1. `pytest -n auto --dist loadgroup` runs python tests with concurrency (faster, same config as CI)
 1. `flake8` runs python lints
 1. `npm run test` runs javascript unit tests using [Mocha](https://mochajs.org)
+1. `npm run test-watch` runs javascript unit tests with the `--watch` option to auto-rerun on test changes
 1. `npm run lint` runs javascript lints
 1. `pytest -k functional` runs the Playwright tests only.
 
--- a/web/frontend/components/AddContent.vue
+++ b/web/frontend/components/AddContent.vue
@@ -23,12 +23,9 @@
         </div>
 
         <div class="add-resource-body" v-if="caseTab">
-          <case-searcher
-            v-model="caseQueryObj"
-            @choose="selectCase"
-          />
-          <case-results :queryObj="caseQueryObj" @choose="selectCase"/>
+          <legal-document-search :casebook="casebook" :section="section" @close="showModal = false"/>
         </div>
+        
         <div class="add-resource-body" v-else-if="textTab">
           <form ref="textForm" class="new-text" v-on:submit.stop.prevent="submitTextForm()">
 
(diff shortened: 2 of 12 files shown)


lizadaly:
This is ready for review before merging to staging for manual testing.

bensteinberg (approved):

PR #1971, opened 2023-04-03
Update the TOC in the current section or casebook after an addition

### LINKED ISSUE #1965: Ensure that the newly-added doc is visible when the add-legal-doc modal is dismissed
Follow-up to #1963 

After a legal doc is added, the user may dismiss the search modal in four ways:

* Hitting ESC
* Hitting the close X on the top right
* Clicking anywhere outside of the modal
* Clicking the gray "close" button inside the modal

All of these ultimately trigger the same "close" event on the `<Modal>` component.

In any of these cases, the user's newly-added document will not appear automatically in the TOC, and it could seem like the action did not work. The document will be present when the page is refreshed.

Ideally:

* The added-legal-doc event will notify the upstream TOC component that it should refresh itself behind the modal as soon as the add is complete.

If that is not readily possible:

* Modify the gray "close" button to read "Back to casebook" and have that button execute a full page refresh.

lizadaly:
…fixes #1965

When a legal doc is added, this will update the TOC in the current section or root of the casebook such that when the user closes the add-content modal, the new document is already in place in the table of contents.

### DIFF
--- a/web/frontend/components/LegalDocumentSearch/LegalDocumentSearch.vue
+++ b/web/frontend/components/LegalDocumentSearch/LegalDocumentSearch.vue
@@ -17,8 +17,10 @@ import SearchForm from "./SearchForm";
 import ResultsForm from "./ResultsForm";
 import url from "../../libs/urls";
 import { get_csrf_token } from "../../legacy/lib/helpers";
+import { createNamespacedHelpers } from "vuex";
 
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):
Cool! Kind of awesome that it's that easy to achieve!

lizadaly:
I was pleasantly surprised by how little code this was!

Issue #1977, opened 2023-04-07
Add unit test coverage to QuickAdd.vue


lizadaly:
This component is being reworked as part of the Legal Doc add changes, but has no test coverage. Add some tests that cover the functionality that is being updated as part of this work.

Issue #1976, opened 2023-04-07
Backport "clone from link" feature to new legal doc search component


lizadaly:
A useful but hard-to-find feature in the existing Quick Add is that if the user pastes in a URL to a resource from H2O, it will import that resource automatically as a clone.

This feature doesn't work in the new legal doc add component because the original implementation mutated the state of the old component in a tricky way.

Re-add this feature but do so in a way that is more explicit.

Issue #1975, opened 2023-04-07
Upgrade Temp resource flow to use new legal document search component


lizadaly:
Before https://github.com/harvard-lil/h2o/issues/1969 can be complete, the flow for "update a temp resource" needs to be updated to use the new legal doc search. After https://github.com/harvard-lil/h2o/issues/1968 is complete, the only way to get a resource in this state will be from pasting an outline, so this is lowest priority. The main reason to do it would be to remove all of the associated code for https://github.com/harvard-lil/h2o/issues/1969 including the Vuex store pertaining to this feature.

lizadaly:
Before #1969 can be complete, the flow for "update a temp resource" needs to be updated to use the new legal doc search. After #1968 is complete, the only way to get a resource in this state will be from pasting an outline, so this is lowest priority. The main reason to do it would be to remove all of the associated code for #1969 including the Vuex store pertaining to this feature.

mdellabitta:
Reference: #1943

Issue #1968, opened 2023-04-03
Replace quick-add legal doc search with new flow


lizadaly:
Continuing #1933 to handle the quick add feature. This should happen after the add-content feature has been live for a few days and we're certain that it's OK.

Without fully rethinking the UX here, try to reduce the number of clicks needed to go from a search term entered into the quick-add box to a results list.

lizadaly:
This will need to modify the new add legal doc endpoint to accept an optional ContentNode id, representing the Temp node that was created by Quick Add, rather than always creating a new one. This would replace the previous loop of several round trips to first import the legal document, then patch in the association between the two.

PR #1980, opened 2023-04-10
Refactor Quick Add to require fewer clicks and reduce boilerplate

lizadaly:
This is another somewhat monolithic change to the frontend, to rewrite the Quick Add component that appears at the bottom of the table of contents.

Includes the following issues: 

* #1968 
* #1975 
* #1976 
* #1977 

## Before

Status quo is the user enters a case name in the Quick Add, then a temporary item is added, then they click "Specify Document", then they click "Search", then they pick the results (they have to click directly on the title, not the row), then the item is changed in place.

https://user-images.githubusercontent.com/19571/230987676-05cdcfad-0c69-40e1-b71c-7ac311c07ee4.mov

## After

Now the search form is embedded in the Quick Add, so when either automatically (by a matching regular expression) or manually (by pulling the dropdown to Legal Documents), the search happens inline, and the matching item is inserted in place. This reduces the number of clicks from at least 4 to 2.

https://user-images.githubusercontent.com/19571/230987720-74c6b7d3-8395-4bb1-aada-612c0fb38a70.mov

The other functionality is unchanged: users can still paste links from within H2O to auto-clone, paste whole outlines, or add sections or custom content as before.

## Under the hood changes

* Advanced search from Add Content has been refactored into a discrete component and is used in both places. Similarly the actual search API call is now its own function.
* Dropped unnecessary external libraries like lodash and Axios in favor of native browser calls.
* QuickAdd now has some unit tests (it had none before)
* Before, the Clone and Outline options were implicit and then mutated the resource type model in confusing ways. They are now first-class citizens which makes user-facing documentation more straightforward.
* Dropped some dead code (`unWait()`?)

@cath9 and I will probably iterate on this on staging before it gets to prod.

### DIFF
--- a/web/frontend/libs/legal_document_search.js
+++ b/web/frontend/libs/legal_document_search.js
@@ -1,0 +1,153 @@
+import url from "./urls";
+import { get_csrf_token } from "../legacy/lib/helpers";
+
+export const jurisdictions = [
+    { val: "ala", name: "Alabama" },
+    { val: "alaska", name: "Alaska" },
+    { val: "am-samoa", name: "American Samoa" },
+    { val: "ariz", name: "Arizona" },
+    { val: "ark", name: "Arkansas" },
+    { val: "cal", name: "California" },
+    { val: "colo", name: "Colorado" },
+    { val: "conn", name: "Connecticut" },
+    { val: "dakota-territory", name: "Dakota Territory" },
+    { val: "dc", name: "District of Columbia" },
+    { val: "del", name: "Delaware" },
+    { val: "fla", name: "Florida" },
+    { val: "ga", name: "Georgia" },
+    { val: "guam", name: "Guam" },
+    { val: "haw", name: "Hawaii" },
+    { val: "idaho", name: "Idaho" },
+    { val: "ill", name: "Illinois" },
+    { val: "ind", name: "Indiana" },
+    { val: "iowa", name: "Iowa" },
+    { val: "kan", name: "Kansas" },
+    { val: "ky", name: "Kentucky" },
+    { val: "la", name: "Louisiana" },
+    { val: "mass", name: "Massachusetts" },
+    { val: "md", name: "Maryland" },
+    { val: "me", name: "Maine" },
+    { val: "mich", name: "Michigan" },
+    { val: "minn", name: "Minnesota" },
+    { val: "miss", name: "Mississippi" },
+    { val: "mo", name: "Missouri" },
+    { val: "mont", name: "Montana" },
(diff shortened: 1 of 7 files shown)


bensteinberg (approved):
👍

bensteinberg:
`unWait()` sounds like it's up Matteo's alley, like `unClude()`

web/frontend/libs/legal_document_search.js
```diff
@@ -0,0 +1,164 @@
...
+      new URLSearchParams({
+        q: query,
+        jurisdiction: jurisdiction || "",
+        before_date: beforeDate || "",
+        after_date: afterDate || "",
+      });
+    sources.push({ url, id, name, order });
+    order += 1;
+  }
+  return await Promise.all(
```
matteocargnelutti, line 111: ❤️

web/frontend/libs/legal_document_search.js
```diff
@@ -0,0 +1,164 @@
...
+      "Content-Type": "application/json",
+      "X-CSRF-Token": get_csrf_token(),
+    },
+    body: JSON.stringify({
+      source_id: sourceId,
+      source_ref: sourceRef,
+      section_id: sectionId,
+    }),
+  });
+  const body = await resp.json();
```
matteocargnelutti, line 158: I am not sure how the API you're interacting with behaves so this might be irrelevant, but given how frequently I forget about this myself, it's probably worth mentioning: `fetch()` doesn't throw when it encounters HTTP != 2XX. 

With the code above, if your API returns something like: 
```javascript
HTTP 500 
{ "error": "Something terrible happened." }
```

`fetch()` would still resolve, so would `response.json()`, and I think this function would return something like:
```javascript
{
  resourceId: undefined, 
  redirectUrl: undefined, 
  sourceRef: "value of sourceRef"
}
```

I tend to add a status code check that throws if status != 200 on whatever comes out of `fetch` for that reason.

Cheers.

matteocargnelutti (approved):

web/frontend/libs/legal_document_search.js
lizadaly: How am I supposed to know that when it's only in **very noticeable bold type** in MDN?? 

image

Thanks, I did miss this. It's our API so we'll know from Sentry that it threw on the server side, but agreed the client should recover better especially since it's possible the client just passed some unexpected parameters (which does cause a 500, ask me how I know) and the rest of the site might be working fine. I'll address this in a fast follow since I want to get this on staging for Catherine to review.

PR #1985, opened 2023-04-11
Handle non-HTTP 200 responses in fetch calls; fixes #1981

### LINKED ISSUE #1981: Handle non-20x return values in fetch() calls
In both the search and add-document API calls in the new content add components, ensure that the flow just terminates if the API on the back end doesn't return an expected 20x.

In the earlier Axios code, errors just went to the console; we should probably add some simple user notification this go-around.

> `fetch()` doesn't throw when it encounters HTTP != 2XX. 
>
> With the code above, if your API returns something like: 
> ```javascript
> HTTP 500 
> { "error": "Something terrible happened." }
> ```
> 
> `fetch()` would still resolve, so would `response.json()`, > and I think this function would return something like:
> ```javascript
> {
>  resourceId: undefined, 
>  redirectUrl: undefined, 
>  sourceRef: "value of sourceRef"
> }
> ```

_Originally posted by @matteocargnelutti in https://github.com/harvard-lil/h2o/pull/1980#discussion_r1162077787_

lizadaly:
For the three new API calls that have been added using `fetch()`, handle cases where the response that comes back is not OK. 

In most cases this just passes along an error message. In the case of search results, because this expects a list of lists, the error message will be inside the results data structure. 

The error message gets displayed to the user under the text box in most places.

### DIFF
--- a/web/frontend/components/LegalDocumentSearch/LegalDocumentSearch.vue
+++ b/web/frontend/components/LegalDocumentSearch/LegalDocumentSearch.vue
@@ -9,6 +9,7 @@
       :added="added"
       :selected-result="selectedResult"
     />
+    <p class="message" v-if="message">{{ message }}</p>
   </section>
 </template>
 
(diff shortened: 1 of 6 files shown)


matteocargnelutti (approved):
NEAT ✨ !

PR #1987, opened 2023-04-11
Limit text input to appropriate URL types

lizadaly:
If the user selects type Link or Clone (a Clone is a link to another H2O resource), change the input type to `url` and let the browser validate.

For type Clone, applies the `pattern` attribute with the current origin so the browser handles client-side validation. (This pattern matches the internal check on whether a link is considered external or a clone.)

In Chrome and Firefox, the browser will make use of the `title` attribute to hint to the user what the required pattern is. In Safari, naturally, the validation is enforced but the pattern is not displayed. (This is impossible to screenshot.)

With some custom validation we could make this experience a little nicer, but this is almost all declarative markup and very much an edge case. If the user somehow blows past the client-side validation, they'll get the server-side error from #1985 anyway, rather than no response.

### DIFF
--- a/web/frontend/test/components/QuickAdd.test.js
+++ b/web/frontend/test/components/QuickAdd.test.js
@@ -44,14 +44,14 @@ describe("QuickAdd", () => {
   it("loads the quick add form with expected defaults", () => {
     const wrapper = mount(QuickAdd, { store, localVue });
     expect(wrapper.find(".resource-type option:checked").element.textContent).toContain("Section")
-    expect(wrapper.find("[type='text']").element.placeholder).toContain("Week One")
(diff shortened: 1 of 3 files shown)


web/frontend/test/components/QuickAdd.test.js
```diff
@@ -71,7 +71,7 @@ describe("QuickAdd", () => {
     expect(wrapper.vm.results.length).toBe(2);
   });
 
-  it.only("adds an error message if the search fails", async () => {
+  it("adds an error message if the search fails", async () => {
```
lizadaly, line 74: EVERY TIME

matteocargnelutti (approved):

PR #1994, opened 2023-04-13
CourtListener API integration prototype

lizadaly:
This functions as a proof of concept, but it's very rough. There's a lot to do in terms of mapping CourtListener's API fields to our own, and figuring out what transformations need to happen on the content itself.

What this does do:

* Adds CourtListener as a new legal document source. It's set `active=False` and that can be overridden per environment in the admin.
* Supports two search types: a pass-through to their [general search query](https://www.courtlistener.com/?q=%22santa+claus%22) API by default, or if the input seems citation-like, specifies a citation query.
* Supports pulling the content as HTML and inserting it to a casebook. 

What this doesn't do:

* Map anything to our internal `metadata` field for `LegalDocuments`.
* Try to do anything special with the markup like supporting obvious styling hints (`class="indent"`).
* Try to influence the search filtering or ordering of their results.

What this happens to do:

* Pulls the Legal Document Source mappings out of `models.py`. These aren't Django models and it's already hard to reason about that file because it's so long.

Screenshot 2023-04-13 at 1 47 36 PM

## Next steps

I think we'll need a few days to decide whether we want to merge a version of this. The code can easily go live as-is and as long as the source is left in the default `inactive` state, it won't be visible on prod, but could be enabled on staging. We'd want to give all parties a heads-up before doing so and maybe see if there's a more performant API key available; the response time from the one I set up is pretty variable but perhaps it's being rate-limited.


lizadaly:
This is ready to review. It should be deployable to staging without the API key; since the source defaults to `inactive` the relevant code won't be called and this should just be a no-op. 

I created a separate issue for generating and populating API keys for staging and prod and then we can flip the source to "active" on staging and start reviewing the actually behavior and content for what final acceptance criteria should be.

bensteinberg (approved):
👍

PR #1996, opened 2023-04-13
Make Matomo usage reporting async

lizadaly:
Something I should've done a long time ago for @cath9 but fortunately for her now it was annoying me, so I'm improving this.

In the H2O reporting dashboard, high-level stats from Matomo were being fetched synchronously during the request cycle, which made the dashboard page load slowly. For awhile I was considering having the stats fetched and cached via a recurring celery job, but this is simpler, if more ridiculous.

Now the page loads instantly, which gives you immediate access to the Django-side reports, while the Matomo reporting is fetched async and pops in after a few seconds. 

This involved writing some wee little Django scaffolding for the endpoint, and porting the Django template code to JS which modifies the DOM once the response comes back. There are also a few mechanical changes to make the Matomo results pickleable; before they could be Django model instances.

Two small functional changes:

* This reports casebook authorship with the newer controlled-vocabulary of the `Institution` model rather than the older user-supplied `affiliation` field; this should help identify where authors might need an institution applied.
* Changed the messaging when Matomo returns no results to a warning which should hopefully avoid Sentry noise.

### DIFF
--- a/web/main/test/test_permissions.py
+++ b/web/main/test/test_permissions.py
@@ -35,6 +35,10 @@ def get_permissions_tests():
 
         view_func = path.callback
 
+        # don't run tests on anything that's not a pattern (like an include)
+        if not hasattr(view_func, "__name__"):
+            continue
+
         # don't run tests on built-in views:
         if view_func.__name__ in ["RedirectView", "TemplateView"]:
             continue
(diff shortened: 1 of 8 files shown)


lizadaly:
Got some test work to do here, re-drafting.

web/main/test/test_permissions.py
```diff
@@ -35,6 +35,10 @@ def get_permissions_tests():
 
         view_func = path.callback
 
+        # don't run tests on anything that's not a pattern (like an include)
+        if not hasattr(view_func, "__name__"):
```
matteocargnelutti, line 39: @lizadaly: You know what I am going to say.

matteocargnelutti (approved):
NICE 😎

web/main/test/test_permissions.py
lizadaly: :spray-bottle:

Issue #1979, opened 2023-04-07
Reporting: add cumulative totals for Verified Professors with Published casebooks


cath9:
Looking to know the number of Verified Professors in the platform with published casebooks in any given year. 

This should be cumulative. i.e if in 2012, 5 verified professors published casebooks through the platform, and those all remained published in 2013 and also 6 more verified professors published casebooks in 2013, then the numbers I'm looking for would be:
2012 = 5
2013 = 11

Thanks!

lizadaly:
It's surprisingly hard to know when a casebook was unpublished, because for some reason that state is not recorded in the edit log (it does record "first published").

I can tell if a casebook is _currently_ unpublished, but e.g. if I published in 2012, unpublished in 2013, then republished in 2023, there's no way to know that it was unpublished for the intervening 10 years.

There are only ~16 casebooks by professors that were ever published and are now unpublished (out of 417 casebooks ever published by verified professors). If it's important to drop those 16 somewhere in this report, I might be able to provide a guesstimate about which year they dropped out, but it wouldn't be accurate.

Issue #1978, opened 2023-04-07
Reporting: Casebooks published by Verified Professors segmented by year


cath9:
Looking to know how many casebooks were published by a verified professor in a given year. This should not be cumulative - just looking for a discrete number for each year. Ideally it will only include the first time a professor clicked "publish", and not include the publication of revision drafts.

lizadaly:
We can't pull real revision drafts cumulatively because the publish edit log only goes back to 2021. Before that I can only tell you when the casebook was first created, not when it was first published, so I'll have to use "first-created" dates for all values.

lizadaly:
Revised this: it looks like the edit log was backfilled, so while published dates prior to 2021 are estimates based on the status of the casebook at the time, after that they should be reliable, so I will use the edit log here.

(In other words, before 2021, we can't track when a book was published-then-unpublished here, but after 2021 this report should include casebooks that were since unpublished, or published in a different year than they were first created.)

PR #1998, opened 2023-04-17
Generate time-series reports for casebooks and professors 

lizadaly:
This generates three new CSV-only reports that compute casebook and professor counts of three types over a period of the last 10 years:

* A count of how many casebooks were published each year by verified professors [#1978] 
* A cumulative running total of how many verified professors have published casebooks by each year [#1979] 
* For each verified professor, a count by year of their published casebooks.

The last item was not requested; I just misread one of the tickets. But perhaps it'll be useful! 

I did some spot-checking of the counts and compared them across each other and against counts from the search page, and I think this is broadly correct. It's impossible to know when a casebook was not in the published state historically. As mentioned in the ticket, only 16 casebooks by verified professors have since been unpublished, so hopefully the absence of this information is not material. Also prior to 2021 we have only guesstimate dates for first-published dates—in reality some of this casebooks were likely published one year later than reported, because we only have creation date for that period.

## Casebooks published by verified professors by year

year | Casebook count
-- | --
2012 | 4
2013 | 9
2014 | 25
2015 | 33
2016 | 35
2017 | 16
2018 | 34
2019 | 31
2020 | 81
2021 | 53
2022 | 79
2023 | 17

This sums to 417—which is close to but doesn't exactly match the count of currently published casebooks because:

* This data is slightly stale
* It does not exclude casebooks in series
* It does not exclude casebooks which were published but have been unpublished

## Cumulative professors with published casebooks

Year | Casebooks by professors
-- | --
2012 | 3
2013 | 11
2014 | 35
2015 | 56
2016 | 77
2017 | 90
2018 | 106
2019 | 128
2020 | 173
2021 | 205
2022 | 259
2023 | 271

* Does not drop out unpublished casebooks because tracking that by date is unreliable.

### DIFF
--- a/web/reporting/views.py
+++ b/web/reporting/views.py
@@ -7,12 +7,9 @@ from dateutil.relativedelta import relativedelta
 from dateutil.rrule import *
 from django.contrib.admin.views.decorators import staff_member_required
 from django.db import connection
-from django.db.models import Count
-from django.db.models.functions import ExtractYear
 from django.http import HttpRequest, HttpResponse, JsonResponse
 
 from main.test.test_permissions_helpers import no_perms_test
-from reporting.create_reporting_views import PUBLISHED_CASEBOOKS
 from reporting.matomo import usage
 
 from .admin.usage_dashboard import DateForm
@@ -40,7 +37,7 @@ def sql_to_csv_response(sql: str, filename: str) -> HttpResponse:
 
 @no_perms_test
 @staff_member_required
-def matomo_stats(request: HttpRequest):
+def matomo_stats(request: HttpRequest) -> JsonResponse:
     """When requested as on page load, retrieve Matomo analytics for the given time period"""
     form = DateForm(request.GET)
     if form.is_valid():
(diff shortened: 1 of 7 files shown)


web/reporting/views.py
```diff
@@ -1,17 +1,43 @@
...
+from django.http import HttpRequest, HttpResponse, JsonResponse
 
-from reporting.matomo import usage
 from main.test.test_permissions_helpers import no_perms_test
+from reporting.matomo import usage
 
 from .admin.usage_dashboard import DateForm
 
 
+def sql_to_csv_response(sql: str, filename: str) -> HttpResponse:
```
bensteinberg, line 18: I'm sure this isn't necessary, but it crossed my mind after a frisson of horror on reading "Given a blob of SQL, execute it": should `sql_to_csv_response()` ensure that it can only run select queries?

lizadaly: I'm not sure how I'd do that. This is just a standalone function, not a view or anything directly callable from the outside—it's just preventing 3 repetitions of the same pattern in the views that follow it. 

If you'd prefer I could unwind it and put it inline in each of the views.

bensteinberg: Nah, it doesn't make sense -- I'm just twitchy.

lizadaly: I'm glad you asked!

bensteinberg (approved):

lizadaly:
Catherine will need to weigh in on whether this is what she's looking for and if the numbers seem truthy to her, but we can do that in a followup since this isn't user-facing anyway.

PR #2001, opened 2023-04-18
Allow new users to self-identify as a particular user type

lizadaly:
This leverages the Django built-in `Group` model (which H2O did not previously use) to let users select from a predefined list of groups to which they may belong. Django Groups are typically used to gather up permissions, as for the admin, but they aren't restricted to that use and are a convenient ManyToMany relationship already available to the `User`.

This PR:

* Runs a data migration to pre-populate the desired group types. The groups have no special behavior; they are just labels.
* Runs a second migration to categorize all verified professors as group "Professor."
* Enables the group mechanism in the changelist and list views of the admin—all groups are filterable in the list view, and can be modified on a per-user basis.

Smaller items:

* Cleans up the list view a bit to remove less-useful fields—it was getting long enough to scroll horizontally on even a wide viewport. Also reordered the list filters to privilege the newer ones.
* Sets the HTML input type for the email address field to type `email`. This would've happened automatically if `User.email_address` were an [EmailField](https://docs.djangoproject.com/en/4.2/ref/models/fields/#emailfield), but since it's not I set the widget type manually.
* Corrects some typographic errors in the help text of the signup form and makes the error message less specific for non-educational domains, since we allow more than just .edu and .gov now.

## Signup form

Looks like this now: 

image

## Admin list page

Added the new groups filter on the right, and customized the display of the groups in the list view to be comma-delimited vlaues.

image


web/main/migrations/0044_add_user_groups.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.18 on 2023-04-18 18:17
```
matteocargnelutti, line 1: @lizadaly Is this migration still correct after your recent edit? (I think so -- but double checking)

lizadaly: It is (but the test failure is real so fixing that).

matteocargnelutti (approved):

Issue #1814, opened 2022-11-09
Clarify or remove the word "original" in the credits tab


lizadaly:
The word "original" appears parenthesized here, but I don't think that term is used elsewhere on the site, and if it's supposed to distinguish imported or caselaw content from self-authored content, I don't think it's working right:

<img width="381" alt="image" src="https://user-images.githubusercontent.com/19571/200913699-c7b3f1dd-8818-46c1-a2fb-d1ebec2871bd.png">

I checked several cases and everything is marked "original."

lizadaly:
I see, it's a link to the "original" version of the content if the casebook was cloned. That wasn't obvious to me at all.

lizadaly:
Especially since in many cases the original version is not public, in which case it leads you to a login page and then probably a permission denied message.

lizadaly:
Rolling this into #1942

PR #2013, opened 2023-04-25
Improve attribution and appearance of credits tab

### LINKED ISSUE #2012: Improve appearance and behavior of credits/attribution page on a casebook
Followup to https://github.com/harvard-lil/h2o/issues/1942

* Foreground contributions from earlier authors rather than burying them in the credits paragraph
* Don't include unpublished drafts in attributions list
* Never link to unpublished sections 
* Don't unnecessarily expand TOC subtrees to the point that the list becomes unwieldy

lizadaly:
fixes #2012 

The Credits page on production has the following limitations:

* Authors who contributed to casebooks further back in the ancestry tree get limited credit.
* Casebook TOCs are expanded in full depth, making for a cluttered display that does not add information.
* Oftentimes the immediate parent of a casebook (the one that was cloned from) is a non-public draft, leading to links that 404 or return permission denied.
* The labeling can be confusing (#1814).
* Depending on the user's context, the Credits tab may apply to the whole casebook or just the last-viewed section, but this distinction is opaque.

In addition, the Credits view has some complex business logic but no tests.

This PR:

* More clearly assigns credit to earlier authors.
* Clarifies and links the original document sources from which those contributions were drawn.
* Limits the the information to only the most relevant resource pages.
* Always apply at the Casebook level.
* Better handle cases where previous revisions were drafts that were subsequently published, or were never published drafts.
* Adds a test suite.

I'm sure there are still some edge cases here but I think this smooths over some of the most common flaws in the status quo.

## Before

Casebooks with lots of contributors generated hugely long lists, and obscured individual contributions. Many of these links 404 or lead to login pages because they refer to now-private drafts.

imageimage

## After 

This shows the immediate parent casebook, plus an additional section below for earlier contributors that includes the top-level casebook or sections that were cloned:

image

If we explore the n-1 casebook, the immediate parent, now the two earlier contributors are foregrounded, and the specific resources are listed:

image

### DIFF
--- a/web/main/templates/includes/credits.html
+++ b/web/main/templates/includes/credits.html
@@ -1,15 +1,26 @@
 <section class="credits">
     <div>
         {% if contributing_casebooks %}
-        <h3>Content from the following sources has been used in the creation of this {{type}}:</h3>
+        <h3>Content from the following sources has been used in the creation of this casebook</h3>
+
+        </p>
         <ul class="casebook-list">
             {% for row in contributing_casebooks %}
             <li>
                 <div class="casebook-credit-header">
-                    <a href="{{row.casebook.get_absolute_url}}"><h4> {{ row.casebook.title }} </h4></a>
+                    <header>
+                        <h3><a href="{{ row.casebook.get_absolute_url }}">{{ row.casebook.title }}</a></h3>
+                        {% if row.casebook.first_published  %}
+                            <span>(First published {{ row.casebook.first_published.entry_date | date:"M Y" }})</span>
+                        {% else %}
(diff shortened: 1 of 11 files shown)


web/main/templates/includes/credits.html
```diff
@@ -1,44 +1,71 @@
 <section class="credits">
     <div>
         {% if contributing_casebooks %}
-        <h3>Content from the following sources has been used in the creation of this {{type}}:</h3>
+        <h3>Content from the following sources has been used in the creation of this casebook</h3>
```
bensteinberg: This looks slightly odd to me without a colon or a period at the end.

bensteinberg (approved):
🔥

PR #2014, opened 2023-05-02
Improve appearance of credits page with multi-source work

lizadaly:
Part of #2012 

Further refinements to the credits page based on feedback:

* Adds a small book cover icon to clarify that the source is a book
* Groups ancestor casebook contributions by casebook

## Example

image

### DIFF
--- a/web/main/templates/includes/credits.html
+++ b/web/main/templates/includes/credits.html
@@ -61,11 +61,13 @@
                                             (first published {{ clone.first_published.entry_date | date:"M Y" }}) 
                                         {% endif %}
                                         with contributions from: 
+                                        <ul>
                                         {% for user in authors %}
                                             <li class="user {% if user.verified_professor %} verified{% endif %}">
                                                 <a href="{% url 'dashboard' user.id %}">{{ user.display_name }}</a>
                                             </li>
                                         {% endfor %}
+                                        </ul>
                                     </li>
                                 </ul>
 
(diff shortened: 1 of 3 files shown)


web/main/templates/includes/credits.html
```diff
@@ -45,23 +46,30 @@ <h4>Authors:</h4>
...
+                                            <a href="{{ clone.get_absolute_url }}">{{ clone.title }}</a> 
+                                        {% else %}
+                                            {{ clone.title }}
+                                        {% endif %}
+                                        {% if clone.first_published.entry_date %}
+                                            (first published {{ clone.first_published.entry_date | date:"M Y" }}) 
+                                        {% endif %}
+                                        with contributions from: 
+                                        {% for user in authors %}
+                                            <li class="user {% if user.verified_professor %} verified{% endif %}">
```
matteocargnelutti, line 66: I'd maybe wrap that `<li>` in a `<ul>` so you don't have 

```html
<ul>
  <li>
    <li></li>
  </li>
</ul>
```

My understanding is that in that situation, some HTML parsers will close both `<li>`s on the first `</li>` encountered, but results may vary.

**Example:**
Screenshot 2023-05-02 at 11 42 54 AM

matteocargnelutti (approved):
Made a minor HTML suggestion, but regardless: 🚀 !

web/main/templates/includes/credits.html
lizadaly: Oh yeah good catch!

PR #2022, opened 2023-05-16
Add new timestamps for recent casebooks in reporting

lizadaly:
@cath9 wanted to see which authors had recent activity on casebooks in reporting. 

This modifies all professor-level reports to include two new columns:

* `most_recently_modified_casebook_title`: the casebook that was most recently modified
* `most_recently_modified_casebook_modification_date`: the date on which it was modified

This is distinct from the most recently _created_ casebook, which may be an obsolete or unused clone and while technically be more recent, actually not have more recent activity.

"most recently modified" here means any modification to any of the `ContentNode` objects (so, the content of the chapter or resource) OR any modification to the `Casebook` object itself. It doesn't include any annotations added to it, including elisions, though the method could be modified to do that. (Ideally H2O would track last-mod date at the casebook level holistically through signals or some other method.)

(Also changed these methods to return strings rather than datetime objects to force Excel to default to understanding them as dates.)

### DIFF
--- a/web/reporting/admin/views.py
+++ b/web/reporting/admin/views.py
@@ -103,8 +103,10 @@ class ProfessorExportMixin(CsvResponseMixin):
             "attribution",
             "email_address",
             "affiliation",
-            "most_recent_casebook_title",
-            "most_recent_casebook_modified",
+            "most_recently_created_casebook_title",
+            "most_recently_created_casebook_creation_date",
+            "most_recently_modified_casebook_title",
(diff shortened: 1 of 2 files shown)


bensteinberg (approved):

bensteinberg:
Sometimes "approved" is not really approval, I guess.

bensteinberg (changes requested):
This needs a test change, I think.

bensteinberg (approved):

PR #2028, opened 2023-09-05
Update Django

bensteinberg:
https://docs.djangoproject.com/en/dev/releases/3.2.21/


matteocargnelutti (approved):
Tests pending

bensteinberg:
Nodejs has changed how it issues packages; stay tuned for an additional commit. This is likely to pose a problem in other repos.

bensteinberg:
See https://github.com/nodesource/distributions#new-update-%EF%B8%8F and https://github.com/nodesource/distributions/discussions/1639

PR #2030, opened 2023-10-04
Dependency updates

bensteinberg:
https://urllib3.readthedocs.io/en/latest/changelog.html#id9 (this link is unlikely to be accurate once there's a newer version)

https://pillow.readthedocs.io/en/stable/releasenotes/10.0.1.html and see also https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html -- this is a major version upgrade.

https://docs.djangoproject.com/en/dev/releases/3.2.22/


rebeccacremona (approved):
Thanks for making these PRs! Sorry I keep missing reviewing them.

PR #2042, opened 2024-03-19
Remove Celery

bensteinberg:
This draft PR removes Celery, which we had intended to use but did not. The original addition was in #1882. Subsequent related PRs include (at least) #1892, #1898, #1899, #1901, #1904, #1925, #1944, and #1960.


rebeccacremona (approved):
LGTM!

We are chatting in person about the implications of removing the celery backend for the database and migrations. ~We don't _think_ there is anything that should necessarily be added to this PR to remove those tables from the DB.... but we're not entirely sure yet.~ We think it is clearest to leave things as they are.

PR #2043, opened 2024-03-20
Suppress error for non-UUID in image route

bensteinberg:
This suppresses errors that occur when someone tries to get routes like `/image/png`, which should 404 instead. The slightly non-idiomatic use of the assertion is in order to have the same effect in the dev environment as in deployments; a bare try-except does not catch the error in dev, where I think the storage abstraction behaves differently -- the error just looks like
```
Traceback:

File "" in
```


rebeccacremona (approved):
Thanks for this!!!

PR #2044, opened 2024-03-20
Suppress error for non-existent page

bensteinberg:
This changes an error to a 404 when the paginator is asked for a page that doesn't exist. Currently, the error looks like
```
Internal Server Error: /casebooks/11306-abortion-guns-and-climate-change-reading-group-spring-2024/as-printable-html/17/

EmptyPage at /casebooks/11306-abortion-guns-and-climate-change-reading-group-spring-2024/as-printable-html/17/
That page contains no results
```

### DIFF
--- a/web/main/views.py
+++ b/web/main/views.py
@@ -13,7 +13,7 @@ from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.views import PasswordResetView, redirect_to_login
 from django.core.exceptions import PermissionDenied
-from django.core.paginator import Paginator
+from django.core.paginator import Paginator, EmptyPage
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):
Ah! This is so great!

Small suggestion: maybe add an info or debug level logging message, indicating which page of which casebook is empty? I don't see us having time to debug _why_ the page is empty, any time soon, but I think if it is empty, that means some bug happened somewhere, and it might be good to leave breadcrumbs for a potential future dev. But no strong feelings.

bensteinberg:
Sure, I'll add logging -- however, my instinct is that this is not from us exposing a bad link, but rather an ill-conceived behavior on the part of a bot; all the hits on the example above are from bingbot.

PR #2045, opened 2024-03-21
Restore playwright; upgrade black

bensteinberg:
This PR restores playwright to `requirements.in`, as we need it for functional tests; it also unpins and upgrades `black` and lets it make some changes.

https://github.com/psf/black/blob/main/CHANGES.md#2430


rebeccacremona (approved):
Hmmm, does this mean the functional tests aren't currently run by Github Actions? Interesting, I didn't realize that.

bensteinberg:
> Hmmm, does this mean the functional tests aren't currently run by Github Actions? Interesting, I didn't realize that.

I think they are running!

rebeccacremona:
I must be confused. If playwright is necessary for functional tests, and functional tests are run in Github Actions, and we removed playwright in an earlier PR... I would have expected to see the tests fail. Since we didn't... I'm confused.

bensteinberg:
Could playwright be built in to the runner?

PR #2050, opened 2024-05-16
add custom ali copyright language

teovin:
- Add a new `ali_licensed` field to the ContentCode model to indicate whether the section/resource has a special ALI copyright text
- Update the copyright template to check and display the new language
- Retroactively update previous entries via a task to populate the new column
- `licensed_materials` list will be updated when we receive it from ALI, right now I only used it as a placeholder. Thus this PR won't be merged until we finalize it.


web/main/models.py
```diff
@@ -1653,6 +1656,41 @@ def export_content(self, request):
...
 
+    def get_ali_license_text(self):
+        licensed_materials = [
+            {
+                "title": "Restatements of the Law",
+                "copyright_year": 2014,
+                "match_words": [
+                    "restatement",
+                    "restatements",
+                    "contracts",
```
bensteinberg: Matching "contracts" seems like it'll be too broad. @jcushman do you have a sense of this?

web/main/models.py
```diff
@@ -1653,6 +1656,41 @@ def export_content(self, request):
...
+                "copyright_year": 2016,
+                "match_words": ["model penal code", "mpc"],
+            },
+        ]
+
+        license_txt = ""
+        title = self.title.lower()
+
+        for item in licensed_materials:
+            if any(word.lower() in title.split() for word in item["match_words"]):
```
bensteinberg: I think this needs to compare phrases, not words, as `match_words` contains both; that is, I don't think we can use `title.split()`.

web/tasks.py
```diff
@@ -465,3 +465,17 @@ def export_node(
     print(
         f"Generated export file ({filesizeformat(len(file_contents))}) in {round(after-before,2)} seconds."
     )
+
+
+@task
+@setup_django
+def populate_ali_licensed(ctx):
```
bensteinberg, line 472: It might be good to for this task to be able to operate in batches or on subsets.

web/main/models.py
teovin: I removed the split() since the words in match words are both single words and the phrases anyway and `if any(word.lower() in title...` will return True for both. Also, the actual list data hasn't been finalized yet as I noted in the PR description.

jcushman (changes requested):
Thanks for making a start on this! As Ben noted, the current algorithm of looking for any word in "match_words" won't work, because nodes can contain words like "contracts" without being from the Restatement of the Law of Contracts. What we want is something that will detect each section in this PDF: https://www.ali.org/media/filer_public/f8/e9/f8e919d5-790e-408c-af0a-65db1687dbc5/2023-checklist.pdf , which you can also see listed on this page: https://www.ali.org/publications/

So for example:

- If the title contains "restatement" AND "contracts", it should have title="Restatement of the Law, Contracts", "years": "1971-2023"
- If the title contains "restatement" AND "agency", it should have title="Restatement of the Law, Agency", "years": "2026-2023"
- if it just says "Contracts: An introduction", it should have ali_licensed=false.

A good way to do this would be to add a test function to the test suite for your new `get_ali_license_text` function that gives it a bunch of challenging examples, run some google searches like "site:opencasebook.org restatement", "site:opencasebook.org contracts", "site:opencasebook.org agency", "site:opencasebook.org torts", and make sure the test suite includes expected results for a range of examples that look challenging to you from those search results.

teovin:
Thanks for the detailed explanation, I will work on it!

web/tasks.py
teovin: I tried running in batches, I tried using Django `Paginator`, and also tried first creating a list of `to_be_updated_ids` and then running a `.filter(id__in=<to_be_updated_ids>).update(ali_licensed=True)` on that list, but they all performed similarly. I will reach out to folks to see if there might be other ideas to make it run faster. In my local, it takes about 6 minutes to iterate over 1.8 million records.

teovin:
@jcushman Some of the titles in the pdf contain 2nd, 3rd, and some additional wording, for example `Conflict of Laws 2d (vols. 1-2) & Appendix (vols. 3-8)`. In my list of materials to check, I didn't include the additional wording. I am also not including them in the text displayed. Some materials have several different title variations due to this, such as `Property`. Due to this, for all the different variations of Property-related titles, for example, I will be displaying the title as below:

image

Another note is that since there are several publications under each title, and I will be grouping similar titles under one through above, I made the copyright years a range of the earliest and latest dates. For example: 2006-2023 for `Agency`. 

Let me know if you are ok with this approach, or would prefer it to be more granular

web/tasks.py
jcushman: For this it seems reasonable to do `for node in tqdm(ContentNode.objects.iterator()):`. `iterator` will iterate through the results instead of loading them all into ram, so the job will have O(1) ram usage instead of O(n). `tqdm` will print progress during the run.

It can definitely be nice to run updates in batches, and in a way that is resumable, but since it only takes 6 minutes and doesn't have to be run in maintenance mode I think this is fine.

I'd also recommend saving if it changes, rather than if it's True, so you can use it in the future when the get_ali_license_text() logic changes.

jcushman:
> Let me know if you are ok with this approach, or would prefer it to be more granular

This is also what I had in mind -- sounds great!

teovin:
One more thing to note is that since we didn't have the Model Penal Code publications in the pdf, I grabbed the two related materials' information from the ALI website (Sentencing, Sexual Assault and Related Offenses). Let me know if anything needs to be corrected for those.

web/main/views.py
```diff
@@ -2132,6 +2132,7 @@ def get(self, request, casebook, section):
             {
                 "casebook": casebook,
                 "section": section,
+                "ali_license": section.get_ali_license_text(),
```
jcushman: These want to use section.ali_license instead of the method, to avoid recalculating, right?

web/main/models.py
```diff
@@ -1653,6 +1657,22 @@ def export_content(self, request):
             return rich_text_export(self.resource.content, request=request, id_prefix=str(self.id))
         return self.resource.content
 
+    def get_ali_license_text(self):
+        with open(f"{settings.STATIC_ROOT}/data/ali_materials.json", "r") as file:
```
jcushman: Loading this file every time will be slow -- I think you could load this once in settings_base.py and store the dict as settings.ALI_MATERIALS or somesuch.

web/main/models.py
```diff
@@ -1653,6 +1657,22 @@ def export_content(self, request):
...
+
+        title = self.title.lower()
+        license_txt = ""
+
+        for item in licensed_materials:
+            if all(word in title for word in item["match_words"]):
+                license_txt = (
+                    f"{item['title']}, copyright @ {item['years']} by the American Law Institute. "
+                    f"Reproduced with permission, not as part of a Creative Commons license."
+                )
```
jcushman, line 1688: this wants a `break` to leave on success.

jcushman (changes requested):
This looks promising! The data structure looks right and the logic looks reasonable.

In addition to the notes I left for efficiency, can you add tests to make sure this performs as you intend? You can see in models.py we have various doctests like

        r"""
        Return all text, including spaces, from the html, using the LXML library.
        >>> html = ' \n <p> \r\n <em> ...'
        >>> assert ContentAnnotation.text_from_html(html) == ' \n  \r\n  foo  ...'
        """

This code will actually be run from pytest, and is able to import fixtures defined in web/conftest.py. So you would do something in the docstring like:

    >>> node = getfixture('contentnode')
    >>> node.title = 'Restatement of Contracts 2d, blah blah blah'
    >>> assert node.get_ali_license_text() == '...'

And include a few different examples to cover different scenarios. Alternatively you can add a test function to a reasonable place in main/test/ if the docstring approach starts to feel clunky.

web/static/data/ali_materials.json
```diff
@@ -0,0 +1,210 @@
...
+    {
+        "title": "Restatement of the Law, The U.S. Law of International Commercial and Investor-State Arbitration",
+        "years": "2023",
+        "match_words": [
+            "restatement",
+            "the u.s. law of international commercial and investor-state arbitrationn"
+        ]
+    },
+    {
+        "title": "Principle of the Law, Aggregate Litigation",
```
jcushman: These should be "Principles" of the Law.

jcushman (changes requested):
Just a typo in the "Principles" titles, then this is ready to go!

web/static/data/ali_materials.json
teovin: Just fixed, had followed a similar pattern as in the restatements without knowing much of their contents. Thank you!

jcushman (approved):

PR #2053, opened 2024-07-18
css update for printable view and href base url update

teovin:
- Add base URL to CL relative URLs in the CL case HTML so we don't return a 404, and instead take the user to the CL website. PS: If there is a strong preference for us to use or not to use `target="_blank"`, let me know and I can make an update accordingly. 
Sample casebook doc that would be fixed by this change: https://h2o-stage.lil.tools/casebooks/8195-test-casebook/resources/1-anderson-v-welding-testing-laboratory-inc/

- Add a CSS attribute to `all.css` so the text inside `<pre>` tags do not overflow in the printable-html view. I used the same attribute here that we use for the regular view. This was impacting CAP-sourced legal docs as well. 

**CL case before:**

image

**CL case after:**

image

### DIFF
--- a/web/main/legal_document_sources.py
+++ b/web/main/legal_document_sources.py
@@ -599,7 +599,8 @@ class CourtListener:
             ):
                 case_text = "".join(sub_opinion[content_type] for sub_opinion in sub_opinion_jsons)
                 if case_text:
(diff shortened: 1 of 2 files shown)


cmsetzer (approved):
Looks good to me!

PR #2055, opened 2024-07-30
Fix export of newly-uploaded images

rebeccacremona:
I _think_ this PR arranges so that images newly uploaded by verified professors to casebooks will appear in DOCX exports of those casebooks.

It turns off all the [URL munging](https://www.tiny.cloud/docs/tinymce/latest/url-handling/) that TinyMCE, the rich text editor, does on save. It is unclear to me why it munges URLs by default. The docs say:
> This option enables you to control whether TinyMCE is to be smart and restore URLs to their original values. URLs are automatically converted (messed up) by default because the browser’s built-in logic works this way. There is no way to get the real URL unless you store it away.

But I do not know what that means.

The route that handles uploads [returns the absolute URL](https://github.com/harvard-lil/h2o/blob/develop/web/main/views.py#L3278) of the upload: so, without any munging, we should get what we want.

Testing locally manually, it works. Adding a functional test to actually try it out as part of the test suite is _way_ hard, so I'm not going to, at this point. 

If for some reason this does not work in prod the same way it does locally, we can try again, this time configuring TinyMCE to use [custom logic](https://www.tiny.cloud/docs/tinymce/latest/url-handling/#urlconverter_callback) to force all image URLs to be absolute links to H2O, while leaving non-image links untouched. Fingers crossed it doesn't come to that 🤞. 

I added a basic test so that I could feel less worried that we are exposed to exfiltration, or similar, if relative paths are manually added to the TextBlock source by accident or by a mischievous user.

As a followup, I plan to spend a little time finding all the image tags in H2O casebooks text, and a) clean up any that I can, so that their export works retroactively, and b) possibly, deal with any other bad ones that turn up.

I am resisting the urge to do any further refactoring... which is extremely tempting.

### DIFF
--- a/web/frontend/libs/tinymce_extensions.js
+++ b/web/frontend/libs/tinymce_extensions.js
@@ -436,6 +436,10 @@ export function getInitConfig(selector, enhanced, code) {
     paste_remove_styles: true,
     paste_remove_styles_if_webkit: true,
     paste_strip_class_attributes: "all",
+    // image URLs
+    relative_urls: false,
+    convert_urls: false,
+    remove_script_host : false,
     media_dimensions: false,
     extended_valid_elements: extend_valid_elements,
     setup: (editor) => {
--- a/web/main/test/test_export.py
+++ b/web/main/test/test_export.py
@@ -1,4 +1,5 @@
 from io import BytesIO
+import itertools
 from pathlib import Path
 from zipfile import ZipFile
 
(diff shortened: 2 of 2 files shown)


rebeccacremona:
Note that images in headnotes will still not export: they are (evidently) intentionally stripped: https://github.com/harvard-lil/h2o/blob/develop/web/main/templates/export/casebook.html#L10

teovin:
Looks good to me, and works as expected in my local.

teovin (approved):

PR #2056, opened 2024-07-31
Try lambda build again

bensteinberg:
See https://stackoverflow.com/a/75149347 and https://github.com/docker/buildx/issues/1509#issuecomment-1378538197


rebeccacremona (approved):
Oh great!! Thanks for looking into this and finding this stopgap

PR #2057, opened 2024-08-01
Dependency updates

bensteinberg:
This is mainly here as a test of #2056 -- I think the only update that might be worth a look is python-docx, and it looks fine: 

https://github.com/python-openxml/python-docx/blob/master/HISTORY.rst


bensteinberg:
(I should have included this in #2056....)

bensteinberg:
The pandoc-lambda README needs some updating, but I'm going to do that in a separate PR, once we've confirmed that the image version is getting updated automatically and the image is getting built and pushed automatically.

rebeccacremona (approved):

PR #2062, opened 2024-09-30
switch to using cl api v4

teovin:
For more details, [visit](https://linear.app/harvard-lil/issue/LIL-2656/switch-h2o-to-use-courtlistener-api-v4).


cmsetzer (approved):
:tada:

PR #2065, opened 2025-01-02
Minimal upgrade to Postgres 16

bensteinberg:
Tests pass locally.

See the PG release notes for versions [13](https://www.postgresql.org/docs/13/release-13.html), [14](https://www.postgresql.org/docs/14/release.html), [15](https://www.postgresql.org/docs/15/release.html), and [16](https://www.postgresql.org/docs/16/release.html).

The RDS upgrade path from 12.19, which is what we are running now, does not extend directly to 16.6, so I took the liberty of upgrading our stage instance from 12.19 to 12.22, and then to 16.6; we can kick the tires on stage. The upgrades took some minutes; we'll go into maintenance mode when it's time to upgrade the production instances.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,7 +74,6 @@ services:
     # command: bash
 
 volumes:
-  db_data_12:
   db_data_16:
   minio_data:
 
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):
LGTM! (Same question, about when to drop the definition of the db_data_12 volume.)

PR #2071, opened 2025-07-03
A small database migration, for testing purposes

rebeccacremona:
We may find it useful to run a database migration, as we are testing new deployment workflows.

This PR makes a very small one.

I noticed that the user pronoun field currently has a max length of 63. That is a relatively unusual limit, and, it turns out, visually breaks the user profile layout rather significantly. 

I checked, and the longest string currently in that field is only 17 characters.

So, this PR changes the limit to 20 characters... which, incidentally, does not break the layout.

We do not have to merge or use this unless we want to.

We also can reverse the change, should we choose to run it, but decide we actually want a longer limit.


rebeccacremona:
LOL if we do use this, I should probably double check no one has added a longer string, in the meantime 😂

rebeccacremona:
We decided not to use this.

PR #2121, opened 2026-09-01
Publish artifacts before deploying anything

jcushman:
Follow-up to #2120, prompted by its first run on `main` failing.

## What happened

The suite passed — 608 tests, JS build, artifact extraction, all green. Then:

```
Build and publish pandoc-lambda image    success
Point the staging Lambda at the new image  FAILURE
Commit Assets                            skipped
Push docker images                       skipped
Build and publish prod-target image      skipped
```

An `AccessDeniedException` on `lambda:UpdateFunctionCode` (fixed separately in [lil-terraform#145](https://github.com/harvard-lil/lil-terraform/pull/145)) meant **nothing was published at all** — not the GHCR images, not the prod image, not the built assets. A tested, good artifact was withheld because an unrelated deployment could not authenticate.

## The change

Publishing and deploying are different kinds of work:

- **Publishing** is gated on the suite passing and nothing else.
- **Deploying** can fail for reasons that say nothing about whether the artifact is good — an IAM gap, a throttle, a service being mid-update.

So publishing now runs first:

```
Commit Assets
Push docker images              -> GHCR
Configure AWS credentials
Build and publish prod-target   -> ECR
Build and publish pandoc-lambda -> ECR
Point the staging Lambda at it  <- the only deploy, now last
```

A deployment failure leaves the published artifacts intact, so re-running deploys rather than rebuilding.

This is a pure reorder — no step's content changes. The added lines in the diff are the moved blocks plus one comment explaining the ordering, so the next person does not "tidy" it back.

## Not the unblock

To be clear about sequencing: this does not fix the failing run. [lil-terraform#145](https://github.com/harvard-lil/lil-terraform/pull/145) does, by granting the OIDC role the Lambda permission its predecessor IAM user had. This just limits the blast radius when a deploy step fails in future.

actionlint clean.

### DIFF
--- a/.github/workflows/lint_test.yml
+++ b/.github/workflows/lint_test.yml
@@ -119,46 +119,6 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
-      # One OIDC session for everything below that touches AWS. Replaces the
-      # long-lived AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY this workflow used
-      # to carry for the pandoc-lambda publish.
-      - name: Configure AWS credentials
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-        uses: aws-actions/configure-aws-credentials@v6.1.1
-        with:
-          role-to-assume: ${{ secrets.STAGING_AWS_ROLE_TO_ASSUME }}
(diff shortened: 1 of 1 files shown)


jcushman:
Superseded by #2122, which incorporates this reorder along with the review findings. Closing.

